PR #85, opened 2021-06-29
common class moved a folder

### LINKED ISSUE #64: Pull out classes to a common folder
We have some repeated classes (Transaction) in node.  Classes can be pulled into their own common folder and instantiated from there.

LijuJoseJJ:
this PR move common class files such as `proof`,  `public-inputs.mjs` and `transaction.mjs` from ` nightfall-client` and `nightfall-optimist` to folder `./common` in the root directory.
fixes #64 

PR updates `.gitignore` of `./nightfall-client` and `./nightfall-optimist` for the above common files as volume share create non-exist file in containers.

new code which share common classes to particular services  containers.
```
 volumes:
      - ./nightfall-client/src:/app/src
      - ./common/classes/proof.mjs:/app/src/classes/proof.mjs
      - ./common/classes/public-inputs.mjs:/app/src/classes/public-inputs.mjs
      - ./common/classes/transaction.mjs:/app/src/classes/transaction.mjs
```
 
since we have code `- ./nightfall-client/src:/app/src` for both `nightfall-client` and `nightfall-optimist` above mentioned create file in container does get written to the particular directory as well, that why code modification in `.gitignore` is needed.

There is another approach I could have taken, i.e to pull classes folder out of src folder and thus avoid problem of file creation at host when its get created in container because of code `- ./nightfall-client/src:/app/src`.
But, this change the code structure in a way that files `./src` folders are importing files from its sibling folders.


Westlad:
Should we make them bind mounts rather than volumes? It seems unusual to have persistent volumes that just hold static code.  The other volumes contain data that's computed at or after deployment of the container.

Westlad:
Can we avoid the empty files/gitignore thing by using a consistent path for the files both in the containers and in the repo?

LijuJoseJJ:
> Should we make them bind mounts rather than volumes? It seems unusual to have persistent volumes that just hold static code. The other volumes contain data that's computed at or after deployment of the container.
> 
> Can we use this as an opportunity to use a consistent style in docker-compose.yml to declare volumes/bind-mounts. Currently it's a bit of a mix? I prefer the 'full' style, e.g.:
> 
> ```yaml
>   volumes:
>       - type: volume
>         source: build
>         target: /app/build/
>       - type: bind
>         source: ./timber/src
>         target: /app/src/
> ```

done

LijuJoseJJ:
> Can we avoid the empty files/gitignore thing by using a consistent path to the class files both in the containers and in the repo?

done. please read updated PR description

Westlad (approved):

PR #93, opened 2021-07-05
Westlad/client rollback

### LINKED ISSUE #88: Client database needs to react to rollback situations
If there is a rollback, both Timber and Optimist will reset their state appropriately.  However this may leave the Client commitment database inconsistent with Timber (and thence the blockchain record).  This needs addressing.

Westlad:
Currently, nightfall-client does not respond to `Rollback` events.  This causes the problem that a rollback may remove a nullifier from the blockchain record.  This would mean that the corresponding commitment was no longer spent and thus could be spent 'again'.  With no rollback capability, nightfall-client will not, however, know this.  This will have the effect of locking up funds.

This PR fixes #88 i.e. the problem above. It makes nightfall-client respond to both `Rollback` and `BlockProposed` events.  Why does it need `BlockProposed`?  To complete a rollback, nightfall-client needs to know which block a nullifier is in, otherwise it can't tell which nullifiers it has to roll back.  `BlockProposed` gives it this information.

nightfall-client now recognises two states for a commitment: `isNullified` is set as soon as a commitment is spent.  This immediately prevents nightfall-client trying to spend it again. However, the commitment isn't 'properly' nullified until the nullifier appears on-chain, at this point nightfall-client also sets `isNullifiedOnChain` from `-1` to the L2 block number of the block that this particular nullifier is in, so that it can be rolled-back in the future if needed.  Of course, we can still get a situation where `isNullified = true` locks a commitment from being spent again but the nullifier never appears on-chain. This isn't really nightfall-client's faulty though - it's up to the user application to make sure they sign and send blockchain transactions that it's given.

I had considered extracting the information needed for nightfall-client from nightfall-optimist and timber, to avoid another connection to the blockchain client.  However, this doesn't work because only nightfall-client has access to the zkp private key and, without this, it's not possible to associate a commitment with its nullifier.  This means that neither nightfall-optimist or timber can tell nightfall-client which commitments it should release for spending.

To make this all work, some changes were needed to nightfall-optimist.  That's because nightfall-optimist does a duplicate nullifier check but, quite reasonably, did not include the concept of a nullifier being removed in a rollback. To fix this rollbacks now delete rolled-back nullifiers in nightfall-optimist (this is a stronger removal than the isMined flag the went before).  There are some other, minor changes to the duplicate nullifier logic (nullifiers are stamped with a blockHash when a block is saved, rather than after checking the block, to avoid an intermediate state where a block is considered 'mined' but its nullifiers aren't). Rollbacks also delete any Transfer or Withdraw transactions that are rolled back.  This is needed because otherwise they may be re-proposed before the corresponding transactions that created the input commitments, which is a challengeable causality violation.  We don't want to give an innocent proposer this headache.  It will be nice, as a future enhancement, to give the transactor their fee back but we don't do that yet to avoid overcomplicating the PR (hopefully this is a small, rare loss).

This PR does not include a capability for nightfall-client to recover if it falls out of sync.  That's because this PR is already quite big.  We'll do the state-sync as a followup PR.  This isn't too mission-critical because falling out of sync will just mean that (in the event of a rollback) some commitments will remain un-spendable unnecessarily.

### DIFF
--- a/nightfall-client/src/classes/commitment.mjs
+++ b/nightfall-client/src/classes/commitment.mjs
@@ -14,6 +14,8 @@ class Commitment {
 
   isNullified = false;
 
+  isNullifiedOnChain = -1;
+
   #computedIndex; // this is the index of this commitment in the Merkle tree. unlike all other numbers in this class, it's a normal Number, rather than a GN for compaitibility with Timber.
 
   constructor({ zkpPublicKey, ercAddress, tokenId, value, salt }) {
--- a/nightfall-client/src/event-handlers/block-proposed.mjs
+++ b/nightfall-client/src/event-handlers/block-proposed.mjs
@@ -1,0 +1,24 @@
+import logger from '../utils/logger.mjs';
+import { markNullifiedOnChain } from '../services/commitment-storage.mjs';
+import getProposeBlockCalldata from '../services/process-calldata.mjs';
+
+/**
+This handler runs whenever a BlockProposed event is emitted by the blockchain
+*/
+async function blockProposedEventHandler(data) {
+  const { nullifiers, blockNumberL2 } = await getProposeBlockCalldata(data);
+  if (nullifiers.length)
+    logger.debug(
+      `Nullifiers appeared on chain at block number ${blockNumberL2}, ${JSON.stringify(
+        nullifiers,
+        null,
+        2,
+      )}`,
+    );
+  // these nullifiers have now appeared on-chain. Thus their nullification
+  // has been confirmed (barring a rollback) and we need to update the
+  // commitment database to that effect
+  markNullifiedOnChain(nullifiers, blockNumberL2);
+}
+
+export default blockProposedEventHandler;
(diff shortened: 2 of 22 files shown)


IlyasRidhuan:
While I think it's a good idea to have client listen to the blockchain, I'm curious as to why this doesn't work.

> I had considered extracting the information needed for nightfall-client from nightfall-optimist and timber, to avoid another connection to the blockchain client. However, this doesn't work because only nightfall-client has access to the zkp private key...

It requires a small modification of the transactions collection to include `blockNumberL2` (which incidentally may be needed for #94 and/or #95). Can't `client` request all transactions that `optimist` has (regardless of `mempool` status) and :

- Filter transactions by the commitments + nullifiers it has stored locally(these are transactions it cares about)
- if there are commitments that are not in any transactions then `isNullified: false` and `isNullifiedOnChain: -1`
- if a commitment is in a transaction marked as `mempool:false` then `isNullifed: true`
- if a nullifier is in a transaction that has  a `blockNumberL2: x` then `isNullifiedOnChain: x`.

Updating can either be through regular long-polling by client to optimist or through a pub/sub websocket where `optimist` can push changes to it's transactions collections.

Westlad:
Yes, that's quite true.  I was a bit imprecise.  What I meant was that Client can't simply do a query to Optimist for the information; it can't ask 'give me the spend status of this commitment', it still has to process blockchain information itself.  In the method you describe it's effectively using Optimist as a proxy server for connecting to the blockchain client. There seems no advantage in routing like that, other than to move load from the blockchain client to the Optimist server, given that the blockchain client already has a suitable subscription interface.

IlyasRidhuan (approved):

PR #102, opened 2021-07-14
Clean up transactions that were proposed in blocks but did not make it on chain

### LINKED ISSUE #95: Transactions in proposedBlocks that are not mined in L1 are not returned to mempool
During their proposal windows, block proposers will create as many blocks as fast as they can (provided there is sufficient unprocessed transactions) and broadcast them to the Ethereum network. Inevitably, their proposal window will come to an end and `changeProposer` will be called to rotate the current proposers.

As L2 block proposers are reliant on Ethereum miners picking up these blockProposed transactions and including them in an ethereum block, there is no guarantee that all of a proposers blocks are mined before a `changeProposer` is. This will cause any remaining blocks created by the proposer to be discarded (these blockProposed transactions are reverted).

Currently when blocks are assembled by proposers, the transactions in these blocks are marked as `mempool: false` so that they are not included in future blocks. If the blocks do not end up being mined, there currently does not exist a mechanism to reset these transactions to `mempool: true`. 

A solution may use one (or a combination) of the following:

1. Add `blockNumberL2` to transactions collection. Therefore, receiving a new blockProposed event that has this `blockNumberL2`  but is not the block we created would indicate that we had "lost" some of our blocks and thus should reset any transactions associated with these "lost" blocks.
2. Detect and handle the occurrence of the revert and reset transactions based on this revert.
3. Enhance the `newCurrentProposer` handler, to recognise if the `currentProposer` has been rotated from us. This should trigger a check of the current on-chain `blockHashes[blockHashes.length -1]` and compare with the local `blocks` collection and enable us to reset.

image

IlyasRidhuan:
This PR closes #95.

When a new `changeProposer` event is received, check if we were the last proposer. 

If we are the last proposer, use the on-chain L2 information to reset any transactions that we have included in proposed blocks that did not make it on-chain before the `currentProposer` was rotated away from us.

### DIFF
--- a/nightfall-optimist/src/event-handlers/new-current-proposer.mjs
+++ b/nightfall-optimist/src/event-handlers/new-current-proposer.mjs
@@ -1,5 +1,12 @@
+import config from 'config';
 import logger from '../utils/logger.mjs';
-import { isRegisteredProposerAddressMine } from '../services/database.mjs';
+import {
+  isRegisteredProposerAddressMine,
+  addTransactionsToMemPoolFromBlockNumberL2,
+} from '../services/database.mjs';
+import { getContractInstance } from '../utils/contract.mjs';
+
+const { STATE_CONTRACT_NAME } = config;
 /**
 This handler runs whenever a BlockProposed event is emitted by the blockchain
 */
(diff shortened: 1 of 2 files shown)


Westlad:
Runs fine on its own but fails on merging with master.  Maybe a rebase and see what happens?

Westlad (approved):

PR #106, opened 2021-07-21
Improve deregister proposer code

### LINKED ISSUE #98: Prevent proposers from unregistering if they have pending blocks that could still be challenged
This is already flagged in the code as a TODO (see pic) 
image

One simple solution to this is to make proposers wait a week before they can 'deregister' (or at least, before they can withdraw their original bond). This is what optimism does (https://research.paradigm.xyz/optimism#incentives--bonds).

IlyasRidhuan:
This closes #98 as well as the comments in that issue.

This PR improves the handling of deregistering proposers:

1. In `deRegisterProposer`, we rotate proposer if the `currentProposer` is being deregistered and now correctly un-splice the deregistering proposer from the linked list.

2. A new `TimeLockedBond` struct is used to hold the bond of deregistering proposers that are waiting out the challenge period.

3. Bonds are now withdrawn using `withdrawBond` that checks that enough time has passed before a withdraw can be processed. A new api endpoint in `optimist` is used to trigger this.

### DIFF
--- a/nightfall-deployer/contracts/Challenges.sol
+++ b/nightfall-deployer/contracts/Challenges.sol
@@ -260,8 +260,7 @@ contract Challenges is Stateful, Key_Registry, Config {
     // blocks
     removeBlockHashes(badBlockNumberL2);
     // remove the proposer and give the proposer's block stake to the challenger
-    state.removeProposer(badBlock.proposer);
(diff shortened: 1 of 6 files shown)


IlyasRidhuan:
This should be good to go after #109 is merged

Westlad (approved):

PR #107, opened 2021-07-21
tokenType key added to allow zero valued ERC20 deposit 

### LINKED ISSUE #97: Add explicit field for token type
Currently the token type (erc20/721/1155...) is inferred from the value/tokenId fields. The downside to this is erc20 commitments of value 0 aren't possible. Adding an explicit field for tokenType will make things simpler and allow 0-valued commitments. The extra gas-cost will be 0, since a uint8 will fit within the existing txData without spilling into a new uint256.

LijuJoseJJ:
fixes #97 

please be thorough with the reviewing of this PR


Westlad:
Looks good.  Can you update the `test-gas.mjs` script too please?

LijuJoseJJ:
> Looks good. Can you update the `test-gas.mjs` script too please?

done

Westlad (approved):

Issue #108, opened 2021-07-22
Differentiate between commitments that are on blockchain and those are pending to be on blockchain 


ChaitanyaKonda:
Today commitments are stored in the NC database by a sender of a transaction during the creation of transfer proof even before the transfer transaction ends up on the blockchain. NC could potentially pick this commitment to spend even before the commitment is added to the blockchain. In this event, a transaction that will spend this commitment will fail because a successful proof to spend this commitment cannot be produced as timber won't be able to provide sibling path or root for this commitment. It is preferable to ensure that NC does not pick a commitment that has not yet been added to the blockchain to avoid user experience errors that will show this error later. This could be done by adding a flag to the collection of commitments that shows the commitment is pending to be added to the blockchain.

PR #114, opened 2021-07-27
Mark on-chain commitments

MirandaWood:
PR for #108, changes are:

- `isNullified` -> now `isPendingNullification` when chosen then cleared (if transfer fails) or `isNullified` if succeeds
- Commitments marked `isOnChain: <blocknum>` when nullifiers are marked as on-chain
- Commitments created via a deposit are marked `isDeposited`
- During a rollback, deposited commitments from that block have their on-chain status revoked and commitments output from a transfer are deleted

Mutex isn't used in the main part of these changes (marking a commitment on-chain so it cannot be chosen) since the user must wait until the mark is added - the _reverse_ of: the user must spend a commitment before `isNullified` is added. In that case, we get a double spend where only one wins. In this case, the user can't choose a commitment and just has to wait and try again. So it's a 'less serious' race condition (it still may be something we want to add in future).


nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -93,11 +135,14 @@ async function findUsableCommitments(zkpPublicKey, ercAddress, tokenId, _value,
...
     .toArray();
   if (commitmentArray === []) return null;
   // turn the commitments into real commitment objects
-  const commitments = commitmentArray.map(ct => new Commitment(ct.preimage));
+  const commitments = commitmentArray
+    .filter(commitment => Number(commitment.isOnChain) > Number(-1)) // filters for on chain commitments
+    .map(ct => new Commitment(ct.preimage));
 
   // Now filter all commitments to also work with those that timber has already seen.
   const knownCommitments = (
```
IlyasRidhuan, line 148: I believe this line is now redundant because the `.filter(commitment => Number(commitment.isOnChain) > Number(-1))` has the same effect of checking that `commitment.index !== null` . The variable names will just need to be updated.

I think this had to be previously split up into two calls because promisfying `.filter` does work/ is not composable

nightfall-client/src/services/transfer.mjs
```diff
@@ -199,9 +204,10 @@ async function transfer(transferParams) {
       .filter(commitment => commitment.preimage.zkpPublicKey.hex(32) === senderZkpPublicKey.hex(32))
       .forEach(commitment => storeCommitment(commitment, senderZkpPrivateKey)); // TODO insertMany
     // mark the old commitments as nullified
-    // oldCommitments.forEach(commitment => markNullified(commitment));
+    oldCommitments.forEach(commitment => markNullified(commitment));
     return { rawTransaction, transaction: optimisticTransferTransaction };
   } catch (err) {
+    oldCommitments.forEach(commitment => clearNullified(commitment));
```
IlyasRidhuan, line 210: - This doesnt look like the right usage for `clearNullified` as that expects a blockNumberL2?
- I **think** this is redundant because `markNullified` is the last call in the `try` block anyways (so we won't have prematurely marked) and i don't see how it would throw an error (i believe it may only do so for upserts)

nightfall-client/src/services/withdraw.mjs
```diff
@@ -133,6 +137,7 @@ async function withdraw(transferParams) {
     // markNullified(oldCommitment);
     return { rawTransaction, transaction: optimisticWithdrawTransaction };
   } catch (err) {
+    clearNullified(oldCommitment);
```
IlyasRidhuan, line 140: Same as in `transfer`

nightfall-client/src/services/commitment-storage.mjs
MirandaWood: I missed that - makes sense! Will revert that change, thanks

nightfall-client/src/services/transfer.mjs
MirandaWood: This is based on a change I did for NF2 where the try block also included the Timber call, so it could fail there but (IIRC) the commitment could be marked as `pendingNullification`, so I did a clear to make sure that change was reverted. Obviously here the `clearNullified` would fail so I'll remove it!

IlyasRidhuan: Yeah good point. There will be some places where the state of the commitment DB will need to be cleaned up now before exiting. Where you have put it should be fine as that will clean it up (as long as it's an idempotent update).

IlyasRidhuan (approved):

PR #119, opened 2021-07-29
fix: mutex now awaits markNullified function

Westlad:
I recently added a mutex to prevent two concurrent calls to a nightfall-client endpoint both selecting the same commitment before it was marked nullified.  Unfortunately we do not await the `markNullified` promise before releasing the mutex.  This renders the mutex mostly ineffective because another process can run before the markNullified is complete. 

This PR fixes the bug by awaiting to the `markNullified`.

### DIFF
--- a/nightfall-client/src/services/commitment-storage.mjs
+++ b/nightfall-client/src/services/commitment-storage.mjs
@@ -115,7 +115,7 @@ async function findUsableCommitments(zkpPublicKey, ercAddress, tokenId, _value, 
   );
   if (singleCommitment) {
     logger.info('Found commitment suitable for single transfer or withdraw');
-    markNullified(singleCommitment);
+    await markNullified(singleCommitment);
     return [singleCommitment];
   }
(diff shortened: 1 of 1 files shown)


nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -187,6 +187,7 @@ async function findUsableCommitments(zkpPublicKey, ercAddress, tokenId, _value,
     );
   }
   commitmentsToUse.forEach(commitment => markNullified(commitment));
+  await Promise.all(commitmentsToUse);
```
IlyasRidhuan, line 190: I don't know if this works as intended, `commitmentsToUse` isn't an array of promises so AFAIK this awaiting does nothing. `forEach` constructs are also a pattern to ignore returns.

We should move this to a `map` and return an array of promises that can then be awaited?

Westlad: Good spot!  Annoyingly my test case misses this one.

IlyasRidhuan (approved):

PR #127, opened 2021-08-11
github action refactored to test tx-gas.mjs test suites

LijuJoseJJ:


Westlad (commented):
To run test-gas properly, we need to set the number of transactions in a block to be 32 (rather than the 2 we usually use for the other tests) but I can figure out if this is being done here.

LijuJoseJJ:
> To run test-gas properly, we need to set the number of transactions in a block to be 32 (rather than the 2 we usually use for the other tests) but I can't figure out if this is being done here.

correction done

Westlad:
I think this was caused by me removing the `Geth`-based tests.  Now that we run Geth in a separate docker project, these tests no longer work.

LijuJoseJJ:
> I think this was caused by me removing the `Geth`-based tests. Now that we run Geth in a separate docker project, these tests no longer work.

conflict resolved

Westlad (approved):

PR #140, opened 2021-08-26
Westlad/chain reorg pt 4

Westlad:
This PR adds the capability to recover from a layer 1 chain reorganisation, adjusting layer 2 data in response to maintain a consistent layer 2 record.  All events except for a Rollback are supported (Rollback has been held as the code is currently being updated elsewhere).

Details of the approach taken are contained in the comments in [this](https://github.com/EYBlockchain/nightfall_3/blob/westlad/chain-reorg/nightfall-optimist/src/event-handlers/chain-reorg.mjs) file's header.

Some related files have been renamed to better reflect what they do (`subscribeToEvents` is now `startEventQueue` for example).

To test, you need to have the Geth blockchain running (`./geth-standalone -s`).  Wait until the DAG generation is completed (check by running `./geth-standalone -l` in the same or another terminal) and then start nightfall_3 without ganache (`./start-nightfall -l`) in another terminal.  Finally run the test (`npm run test-chain-reorg`) in yet another terminal.


Westlad:
This latest commit adds the capability to cope with Nightfall transactions on both chain segments.  It doesn't have a specific test yet.  I intend to add that as a separate PR because the test setup is complex.

nightfall-optimist/src/services/database.mjs
```diff
@@ -27,13 +27,16 @@ export async function saveCommit(commitHash, txDataToSign) {
...
+Function to retrieve a commit, by commitHash, it also returns the 'retrieved'
+which will be true if the commitment hash has already been retrieved
 */
 export async function getCommit(commitHash) {
   const connection = await mongo.connection(MONGO_URL);
   const db = connection.db(OPTIMIST_DB);
   const query = { commitHash };
-  return db.collection(COMMIT_COLLECTION).findOne(query);
+  const commit = await db.collection(COMMIT_COLLECTION).findOne(query);
+  if (commit) db.collection(COMMIT_COLLECTION).updateOne(query, { $set: { retrieved: true } });
```
IlyasRidhuan, line 38: I believe this will need to be `awaited` or we will end up returning before the update is run and the commit marked as `retrieved`?

nightfall-optimist/src/services/database.mjs
```diff
@@ -215,9 +223,24 @@ Function to save a (unprocessed) Transaction
...
+  // 3) We're trying to save a transaction that we've seen before but it was re-mined due to a chain reorg. In
+  //    this case, it's fine, we just update the layer 1 blocknumber and transactionHash to the new values
   const connection = await mongo.connection(MONGO_URL);
   const db = connection.db(OPTIMIST_DB);
-  return db.collection(TRANSACTIONS_COLLECTION).insertOne(transaction);
+  const query = { transactionHash: transaction.transactionHash };
+  const update = { $set: transaction };
+  const existing = await db.collection(TRANSACTIONS_COLLECTION).findOne(query);
+  if (!existing || !existing.blockNumber) {
+    if (existing) transaction.mempool = false; // it's a re-mined transaction.  This will stop block-assembler making a new block
```
IlyasRidhuan, line 240: My understanding is that this line runs IFF the transaction is in the database AND if the `blockNumber === null` (i.e. this is a transaction that has been remined). It's not clear to me why this would guarantee that this means `mempool = false` for this tx. 

It is possible for me to have submitted a transaction onto a chain fork that did not find its way into a block. `mempool = false` should be a flag that is only set as part of `blockProposed` handling or am I missing something.

nightfall-optimist/src/services/database.mjs
Westlad: Good shout.

nightfall-optimist/src/services/database.mjs
Westlad: Yes, quite right.  I need to get the block assembler to ignore re-mined transactions.

Westlad: The problem is that we will then see re-mined L2 blocks which originated on the orphaned chain.  These are not necessarily invalid because if there are no L2 blocks on the new chain, then the re-mined blocks could well be re-mined in the correct blockNumberL2 order.  Thus they won't get reverted at layer 1. However, we don't know what order they'll actually turn up in.  Thus we don't know if our re-mined transaction will actually end up being eventually incorporated in a block or not.  So we don't know whether to allow it to be incorporated into a fresh block or not. 
The simplest thing is just to dis-allow re-mined transactions, which is what my code does by pulling them out of the mem-pool.  That may not be optimal though and eventually we may think up a more sophisticated approach

Westlad: Good point.  I've made the mempool logic more subtle by including a check the the transaction submitted handler, which will mark `mempool = false` iff the transaction is a re-mined one that is _also_ already placed in a L2 block.

nightfall-optimist/src/services/database.mjs
Westlad: fixed

nightfall-optimist/src/event-handlers/new-current-proposer.mjs
```diff
@@ -18,10 +18,11 @@ async function newCurrentProposerEventHandler(data, args) {
   try {
     logger.info(`Proposer Handler - Current proposer is ${currentProposer}`);
     // remember the current proposer.  We don't store it in the DB as it's an
-    // ephemeral thing. Instead, we have a nice object to remember it. The
+    // ephemeral thing. Instead, we have an object to remember it. The
     // object is instantiated at the 'main()' level, so will stay in scope even
     // when this handler exits.
     proposer.address = currentProposer;
+    proposer.proposers.unshift(currentProposer); // useful if we need to revert the proposer because of a chain reorg.
```
IlyasRidhuan, line 25: This global `proposers` array will grow unbounded with each proposer rotation and contain a large number of duplicates. Given the relative rarity of a chain re-org I would expect this proposers array to become very large which is both a potential memory constraint and inefficient (AFAIK unshift is O(n) ).

Noting that to `removeNewCurrentProposer` we need a way to access past currentProposers; alternate ways could be to: 
- Use the on-chain `proposers[currentProposer].previousAddress` to step back when a `NewCurrentProposer` is removed or,
- Change the `NewCurrentProposer` event to emit the `LinkedAddress` struct which contains the `previousAddress`, `currentAddress` and `nextAddress`.

Westlad: Yes, this isn't optimal.

Westlad: Now I read the currentProposer LinkedAddress variable.  That should be the actual value I want because it will have reverted to whatever currentProposer was before the chain reorg.

IlyasRidhuan (approved):

PR #141, opened 2021-08-28
github action workflow added for blockchain running on host

### LINKED ISSUE #133: Make new Geth tests in GitHub Actions
The previous Geth tests have been removed because they no longer work.  However it would be great to reinstate them based around the new approach of running the Geth blockchain as a separate docker project.  One issue with doing this is that the new Geth tests make use of the Mac-specific reference to the docker host (`host.docker.internal`), so that Nightfall_3 can access the Geth blockchain via the Mac localhost ports.  This may be hard to work around.

LijuJoseJJ:
this PR adds a new workflow to github action's PR check to test integration with ganache running on host machine.

Note: Changes in this require Docker version to 20.04 and above. for mac user please upgrade to Docker Desktop v3.6.0 which should do the job

fixes #133


Westlad:
Shouldn't this be running with a Geth blockchain?  I think it won't provide a sufficiently rigourous test with Ganache because of the need to test potential timing issues, Ganache being faster, generally.

LijuJoseJJ:
> Shouldn't this be running with a Geth blockchain? I think it won't provide a sufficiently rigourous test with Ganache because of the need to test potential timing issues, Ganache being faster, generally.

understood, working on it

LijuJoseJJ:
> > Shouldn't this be running with a Geth blockchain? I think it won't provide a sufficiently rigourous test with Ganache because of the need to test potential timing issues, Ganache being faster, generally.
> 
> understood, working on it

done

Westlad (approved):

PR #142, opened 2021-09-01
Fix duplicateTransactions challenge

IlyasRidhuan:
### Safer access pattern
Building a challenge for a `duplicateTransaction` uses an unsafe access pattern via `getBlockByTransactionHash` and `findOne`. `findOne` uses an unreliable "natural ordering" to return the blocks and is not suitable if a call might return two items.

The bad block containing the duplicate is saved before this challenge code runs, therefore there will be two blocks in the collection that could be returned from this call. This means there is a 50:50 chance that the wrong block is returned from this call.

In this case, the wrong block is defined as the bad block itself. This would lead to the challenge using the bad block twice as inputs to the on-chain challenge, rather than the bad block and an earlier valid block that contains the first time the duplicate was used.

### Fix issue with duplicate challenge
One of the reasons this hasn't caused an issue before (despite the 50:50 odds) is that the on-chain `challengeNoDuplicateTransaction` does not guard against using the same block as both inputs. This meant that any block was arbitrarily challengeable.

### DIFF
--- a/nightfall-deployer/contracts/Challenges.sol
+++ b/nightfall-deployer/contracts/Challenges.sol
@@ -88,6 +88,10 @@ contract Challenges is Stateful, Key_Registry, Config {
     // first, check we have real, in-train, contiguous blocks
     state.isBlockReal(block1, transactions1, block1NumberL2);
     state.isBlockReal(block2, transactions2, block2NumberL2);
+    // If the duplicate exists in the same block, the index cannot be the same
+    if(block1NumberL2 == block1NumberL2)
+      require(transactionIndex1 != transactionIndex2, 'Cannot be the same index');
+
     require(
       Utils.hashTransaction(transactions1[transactionIndex1]) ==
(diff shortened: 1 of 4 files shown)


nightfall-deployer/contracts/Challenges.sol
```diff
@@ -88,6 +88,10 @@ contract Challenges is Stateful, Key_Registry, Config {
     // first, check we have real, in-train, contiguous blocks
     state.isBlockReal(block1, transactions1, block1NumberL2);
     state.isBlockReal(block2, transactions2, block2NumberL2);
+    // If the duplicate exists in the same block, the index cannot be the same
+    if(block1NumberL2 == block1NumberL2)
```
Westlad, line 92: Think this should be `block1NumberL2 == block2NumberL2`

IlyasRidhuan: ahh yep, tab-completed too quickly

Westlad (approved):

PR #145, opened 2021-09-07
feat: priority queues

### LINKED ISSUE #134: Split Optimist event processing queue
Currently `optimist` uses a queue to process event-handlers (enforcing sequential processing). All events are enqueued into the same queue, which raises two potential issues.

1.  Since transaction events are pre-verified before adding to the local database, a large volume of transactions can lead to a DoS vector to block assembly. A `currentProposer` may spend most of their block production time verifying incoming transactions rather than building blocks.

2.  Given the event-loop nature of node, the current design of the `makeBlock` means node will attempt to interleave the process whenever it can. This may lead to unintended events as a new `currentProposer` may begin making blocks without having first ensured it has processed all outstanding `blockProposed` events.

All events are not equivalent in terms of their priority.

### Potential fixes
Certain events should be split into high and low priority queues. High priority are state-based events that are directly depended on by other on-chain state altering functions (e.g. `blockProposed`, `NewProposer`, `Rollback`, etc). Low priority events include `TransactionSubmitted`.

The block assemble function should also be converted to an enqueue-able or controllable event to ensure it is only triggered when the `optimist` instance database correctly reflects the on-chain state. E.g., check priority queue is empty before making blocks

Westlad:
fixes #134 

This PR creates two queues for incoming blockchain events.  Events in the lower priority queue will only be processed if the higher priority queue is empty, otherwise the handler will await the higher priority queue emptying.  The code is generic and more queues can easily be added by putting Queue objects in the `queues` array.

TransactionSubmitted events are placed in the lower priority queue.  All other events are placed in the higher priority queue. To facilitate this, eventHandlers have an associated `priority`, with `priority = 0` being the highest.

The block assembler now monitors the highest priority queue and will only make blocks when it is empty.

### DIFF
--- a/nightfall-optimist/src/services/event-queue.mjs
+++ b/nightfall-optimist/src/services/event-queue.mjs
@@ -31,7 +31,7 @@ This function will return a promise that resolves to true when the next highest
 priority queue is empty (priority goes in reverse order, prioity 0 is highest
 priority)
 */
-export function nextHigherPiorityQueueHasEmptied(priority) {
+export function nextHigherPriorityQueueHasEmptied(priority) {
   return new Promise(resolve => {
     const listener = () => resolve();
     if (priority === 0) resolve(); // resolve if we're the highest priority queue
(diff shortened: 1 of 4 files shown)


nightfall-optimist/src/services/event-queue.mjs
```diff
@@ -39,15 +60,19 @@ async function buffer(eventObject, eventHandlers, ...args) {
...
+    queues[priority].push(async () => {
+      await nextHigherPiorityQueueHasEmptied(priority); // prevent eventHandlers running until the higher priority queue has emptied
+      eventHandlers.removers[eventObject.event](eventObject, args);
+    });
     // otherwise queue the event for processing.
   } else {
     logger.info(`Queueing event ${eventObject.event}`);
-    eventQueue.push(() => eventHandlers[eventObject.event](eventObject, args));
+    queues[priority].push(async () => {
+      await nextHigherPiorityQueueHasEmptied(priority); // prevent eventHandlers running until the higher priority queue has emptied
```
IlyasRidhuan, line 71: typo for priority

IlyasRidhuan (approved):

Issue #96, opened 2021-07-06
Handling idle proposers


IlyasRidhuan:
Currently, a proposer that does not produce a block during their proposal window (either intentionally or not) is not punished. Idle periods cause a decrease in the network's throughput and can be performed without consequence.

In the most extreme case, a bad actor can register multiple addresses as proposers on the network. They can then choose to remain idle or simply create blocks that transact between their own accounts. Both cases reduce the effective throughput of the network. Furthermore, once they are done with the attack - the bond can be safely returned in full via `de-registration`.

Transacting between their own account does not seem easily solvable. This effectively reduces to detecting sybils which is AFAIK does not have (can't have) an on-chain challenge. If a bad actor chooses to 'wash trade' like this, there will still be liable for the on-chain ethereum fees to propose the blocks - therefore long term attacks will cost money.

Detection of honest or dishonest idle behaviour may rely on the use of the proposers list and the `blockHashes` array. Gaps can be identified if the proposers of `blockHashes[x]` and `blockHashes[x+1]` are `proposerA` and `proposerC` respectively whilst the proposers linked list is ordered such that `proposersA` -> `proposersB` -> `proposersC`.

Care has to be taken as the following scenario may cause an unfair challenge.

```
Block1             Block2            Block3             Block4
proposerA    ->    proposerB   ->    proposerC    ->    proposerD
Good Block         Bad Block         Good block         Good block
```

If at this point Block2 is challenged and Rolled-back,  all of proposerCs' blocks will be deleted. When proposerB is removed, the order of propers will be proposerA -> proposerC -> proposerD. This will enable proposerD to challenge that proposerC was idle, when in reality the Rollback made it appear so.

<img width="1938" alt="Screen Shot 2021-07-06 at 14 00 40" src="https://user-images.githubusercontent.com/9293647/124550006-9ac9a680-de62-11eb-9e13-25bba755a70a.png">

PR #162, opened 2021-09-17
Hari/idle proposers

dwebchapey:
partial fix on idle proposers issue #96 

Consider following  following scenario 
```
Block1             Block2            Block3            Current proposer
proposerA    ->    proposerB   ->    proposerC    ->    proposerD
Good Block         Bad Block         Good block         Yet to publish
```
If at this point Block2 is challenged and Rolled-back, all of proposerCs' blocks will be deleted. When proposerB is removed, the order of propers will be proposerA -> proposerC -> proposerD. 

This will enable proposerD to challenge that proposerC was idle, when in reality the Rollback made it appear so.

Inorder to get control of this scenario , the proposer state can also be rolled up corresponding to the last genuine block , in this case , Block 1, proposer A . The current proposer will be set to block after Block1 proposer , which is proposer C now (proposer B is removed after roll back ) .  

in `Challenges.sol>Challenges>challengeAccepted()` function , we perform 3 operations

1. emitRollback
2. removeBlockHashes
3. rewardChallenger

In `rewardChallenger` we call `removeProposer` where we recreate the proposers list after delinking bad proposer and also doing a check that if removed proposer is 'currentProposer'  , we set the proposer to next address.  In this fix , we set the `currentProposer` as next address irrespective of current state of proposer list.

```

  function removeProposer(address proposer) public onlyRegistered {
    address previousAddress = proposers[proposer].previousAddress;
    address nextAddress = proposers[proposer].nextAddress;
    delete proposers[proposer];
    proposers[previousAddress].nextAddress = proposers[nextAddress].thisAddress;
    proposers[nextAddress].previousAddress = proposers[previousAddress].thisAddress;
   //change currentProposer to next aaddress irrespective of whether bad proposer is currentProposer
   // if(proposer == currentProposer.thisAddress) {
      // Cannot just call changeCurrentProposer directly due to the require time check
      proposerStartBlock = block.number;
      currentProposer = proposers[nextAddress];
      emit NewCurrentProposer(currentProposer.thisAddress);
   // }
  }

```

This would reset the state of both L2 blocks and proposer list to pre rollup state with current proposer set as next addr of latest genuine L2 proposer.

Points to ponder

- In above example , if a rollback at B happen when D is current proposer , the new proposer will be C . This can be argued since D is made to wait further to be proposer due to no fault of its own . But the hypothesis is that since C's blocks were also 'unfairly' slashed from chain , it would be fair to reset the counter to C. 

- Scenario when rollback will be beneficial for a newly joined proposer since they could be picked sooner
Eg :  If a new proposer E join right while C is current proposer it will added to back of the queue , the new proposer list will be 

A <-> _**B <-> E <-> C**_ <-> D  

After B is rolled back , proposer E gets to be current proposer instead of proposer C which gives unfair advantage to E over C.

### DIFF
--- a/nightfall-deployer/contracts/Challenges.sol
+++ b/nightfall-deployer/contracts/Challenges.sol
@@ -275,6 +275,8 @@ contract Challenges is Stateful, Key_Registry, Config {
     for (uint256 i = lastBlock; i >= blockNumberL2; i--) {
       state.popBlockData();
     }
+    // set current proposer the one with latest valid block
+    //state.setCurrentProposer(state.getBlockData(blockNumberL2-1).blockProposer);
     require(
       state.getNumberOfL2Blocks() == blockNumberL2,
       'After removing blocks, the number remaining is not as expected.'
--- a/nightfall-deployer/contracts/Challenges.sol
+++ b/nightfall-deployer/contracts/Challenges.sol
@@ -274,7 +274,7 @@ contract Challenges is Stateful, Key_Registry, Config {
     uint256 lastBlock = state.getNumberOfL2Blocks() - 1;
     for (uint256 i = lastBlock; i >= blockNumberL2; i--) {
       state.popBlockData();
-    }  
+    }
     require(
       state.getNumberOfL2Blocks() == blockNumberL2,
       'After removing blocks, the number remaining is not as expected.'
--- a/nightfall-deployer/contracts/Challenges.sol
+++ b/nightfall-deployer/contracts/Challenges.sol
@@ -263,10 +263,6 @@ contract Challenges is Stateful, Key_Registry, Config {
     // challenged from the linked list of blocks and all of the subsequent
     // blocks
     removeBlockHashes(badBlockNumberL2);
-
-    //set current proposer the one with latest valid block
-    //state.setCurrentProposer(state.getBlockData(badBlockNumberL2-1).blockProposer);
-    
     // remove the proposer and give the proposer's block stake to the challenger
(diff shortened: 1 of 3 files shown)


nightfall-deployer/contracts/Challenges.sol
```diff
@@ -274,7 +274,7 @@ contract Challenges is Stateful, Key_Registry, Config {
     uint256 lastBlock = state.getNumberOfL2Blocks() - 1;
     for (uint256 i = lastBlock; i >= blockNumberL2; i--) {
       state.popBlockData();
-    }
+    }
```
IlyasRidhuan, line 277: nit: we don't have a linter running on `.sol` files but this white space should be removed

nightfall-deployer/contracts/State.sol
```diff
@@ -154,12 +154,13 @@ contract State is Structures, Config {
     delete proposers[proposer];
     proposers[previousAddress].nextAddress = proposers[nextAddress].thisAddress;
     proposers[nextAddress].previousAddress = proposers[previousAddress].thisAddress;
-    if(proposer == currentProposer.thisAddress) {
+   // if(proposer == currentProposer.thisAddress) {
```
IlyasRidhuan, line 157: Delete the code instead of just commenting it out. As much as we can comments should be reserved for actual comments.

IlyasRidhuan (changes requested):

Westlad (approved):

PR #179, opened 2021-09-29
More robust wait for commitment

IlyasRidhuan:
As also identified by @LijuJoseJJ , sometimes the tests fail because `commitment/salt` has not yet been stored in the local database before we try to query it.

Here we just wait to ensure the `commitment` variable a value before progressing.


Westlad (commented):
Could you remove the blank lines please?

Westlad (approved):

PR #189, opened 2021-10-06
fix: incorporate block assembler to event queue

### LINKED ISSUE #173: Incorporate block assembler into event queue
Currently, the block assembler runs in the background as part of an `async`. This means that while it is non-blocking, it will be inserted in the node event loop whenever there is a free slot.

While this ensures that block assemblers will make blocks as fast as possible, it makes it difficult to control the block assembler when larger state event are happening (e.g. rollbacks). These state events have synchronicity controlled via a priority `Queue` object.

While the block assembly code waits for a certain priority queue to be empty, it is still difficult to verify how effective this wait is in controlling the block assembler.

E.g. A block assembler will wait for the `Rollback` event to be popped off the queue (i.e. the queue is now empty) before trying to assemble a block. However, the rollback event make take some time to perform - meanwhile the block assembler operations may be interleaved within the node event loop, this could cause issues.

A better way would be to incorporate the block assembly as enqueue-able events into the priority queues. This should provide guarantees that an event is wholly processed before assembling begins.

dwebchapey:
This PR fixes #173 

1.  Enqueue `conditionalMakeBlock()` to `fastQueue` via `queueBlockAssembler()` . This prevents `makeBlock()` from executing while on chain events are getting processed.
2.  `queueBlockAssembler()` is triggered on two scenarios 

-  from `newCurrentProposerEventHandler` when proposer is me
- after completion of `conditionalMakeBlock()` given proposer is still me , priority zero queue is zero and enough unprocessed transactions are available in local db

Before calling `makeBlocks()` , we do 2 checks

1. proposer.IsMe  ?  move to stage 2  : exit without re-queuing. This stops putting assemblers into the queue when we lose the proposer role
2. fastQueue === 0 && unprocessedTx >= txPerBlock ? proceed to makeBlocks , re-enqueue event  and exit: re-queue event and exit.

Before  re-enqueue , wait for `fastQueue` to be empty and also db to procure enough unprocessed transactions . This will prevent unnecessary en-queue and de-queue operations. 
The idea is that we fail fast if any of these checks fail, and we optimistically re-enqueue the `conditionalMakeBlock` also reducing queue operations.


nightfall-optimist/src/services/event-queue.mjs
```diff
@@ -17,21 +17,23 @@ Web3js handles event removal (due to a chain reorg) by re-emitting the (now remo
 event with its `removed` property set to true. In the code below, we look out for
 and catch these removals, processing them appropriately.
 */
+/* eslint-disable import/no-cycle */
```
IlyasRidhuan, line 20: Cyclic imports are an anti-pattern and so we should not just disable the check if one exists.

The specific block-assemble logic should be kept out of this file. There is a strong chance this event-queue code will need to re-used in `client` and thus elevated to a `common-files` level.

Instead a generic enqueue function would be preferred - this should also prevent the cyclic dependency

nightfall-optimist/src/services/event-queue.mjs
```diff
@@ -43,6 +45,19 @@ export function nextHigherPriorityQueueHasEmptied(priority) {
...
 
+/**
+This function will push conditionalmakeblock into fastQueue and assures
+no onchain event is executed in parallel with makeblocks . Additonal checks 
+has to be done for makeblocks since following on chain event in order is important
+which is performed and explained in block-assembler service
+*/
+export async function queueBlockAssembler(proposer) {
+  queues[0].push(async () => {
+    await nextHigherPriorityQueueHasEmptied(0); // prevent conditionalmakeblock from running until fastQueue is emptied
```
IlyasRidhuan, line 56: `await` called on queue 0 has no impact , so i'm unclear why it is being called here. We should not need to do a wait on empty check here

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -41,18 +42,23 @@ async function makeBlock(proposer, number = TRANSACTIONS_PER_BLOCK) {
...
+  // if we are the current proposer, and there are enough transactions waiting
+  // to be processed, we can assemble a block and create a proposal
+  // transaction. If not, we must wait until either we have enough (hooray)
+  // or we're no-longer the proposer (boo).
+  if (proposer.isMe && makeBlocks) {
+    // makeBlocks flag is set to false to make sure only 1 instance of makeBlock will run at a time
+    // it is important for queues length to be empty since a rollback or any onchain event following
+    // in queue after blockassemble should not be done So if other events are present in queue
+    // we wait for queue to empty and re enqueue blockassembler again
+    makeBlocks = false;
```
IlyasRidhuan, line 57: I don't see the effect that `makeBlocks` has on controlling the number of `conditionalMakeBlock` instances are in the queue.

IlyasRidhuan (changes requested):

dwebchapey:
- Removed cyclic dependency by creating a generic enqueue function `eventQueueManager`
- Parametrised `eventQueueManager` with callback , priority and args 
- Removed dependency on `makeBlocks` for single instance check

nightfall-optimist/src/services/event-queue.mjs
dwebchapey: `priority` is passed as an argument to make function generic to all queues

nightfall-optimist/src/services/event-queue.mjs
```diff
@@ -43,6 +43,13 @@ export function nextHigherPriorityQueueHasEmptied(priority) {
   });
 }
 
+export async function eventQueueManager(callback, priority, args) {
+  queues[priority].push(async () => {
+    await nextHigherPriorityQueueHasEmptied(priority); // prevent conditionalmakeblock from running until fastQueue is emptied
```
IlyasRidhuan, line 48: I'm unclear with what this line does

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -41,18 +41,18 @@ async function makeBlock(proposer, number = TRANSACTIONS_PER_BLOCK) {
...
-    // if we are the current proposer, and there are enough transactions waiting
-    // to be processed, we can assemble a block and create a proposal
-    // transaction. If not, we must wait until either we have enough (hooray)
-    // or we're no-longer the proposer (boo).
-    if ((await numberOfUnprocessedTransactions()) >= TRANSACTIONS_PER_BLOCK && proposer.isMe) {
+  // if we are the current proposer, and there are enough transactions waiting
+  // to be processed, we can assemble a block and create a proposal
+  // transaction. If not, we must wait until either we have enough (hooray)
+  // or we're no-longer the proposer (boo).
+  if (proposer.isMe) {
```
IlyasRidhuan, line 51: Is there a reason this is split across two `if` statements?

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -76,10 +76,11 @@ export async function conditionalMakeBlock(proposer) {
...
       // blocks with them
-      await removeTransactionsFromMemPool(block); // TODO is await needed?
+      await removeTransactionsFromMemPool(block);
+      await eventQueueManager(conditionalMakeBlock, 0, proposer);
+      // TODO is await needed?
+    } else {
+      await eventQueueManager(conditionalMakeBlock, 0, proposer);
     }
-    // slow the loop a bit so we don't hammer the database
-    await new Promise(resolve => setTimeout(resolve, 1000));
```
IlyasRidhuan, line 82: Do we still need some slowing down of this function call?

nightfall-optimist/src/services/block-assembler.mjs
dwebchapey: The outer `If` does not re-enqueue `conditonalMakeBlock` since proposer is no longer the host node.

nightfall-optimist/src/services/block-assembler.mjs
IlyasRidhuan: @dwebchapey

PR #194, opened 2021-10-07
generic naming fixes of ENV and config variables for infura work

### LINKED ISSUE #193: Change variable names to be generic
Renaming variables such that they are generic despite the blockchain deployed on 
For example, `BLOCKCHAIN_TESTNET_URL` to `BLOCKCHAIN_HOST, BLOCKCHAIN_PORT or BLOCKCHAIN_URL`.
We are doing this because the code will later be used for main net deployment too. In line with this, we should be making our naming generic.

Renaming `ENABLE_TESTNET_DEPLOY` to `USE_INFURA` might be suitable considering the context. We seem to use this flag while instantiating a web3 instance to use an Infura secret in the options or not. 

Finally, a quick change from `WEB3_RECONNET` to `WEB3_RECONNECT`

LijuJoseJJ:
this PR fixes #193

### DIFF
--- a/common-files/utils/web3.mjs
+++ b/common-files/utils/web3.mjs
@@ -21,19 +21,19 @@ export default {
     logger.info('Blockchain Connecting ...');
 
     let provider;
-    if (config.ENABLE_TESTNET_DEPLOY) {
+    if (config.USE_INFURA) {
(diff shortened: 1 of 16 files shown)


ChaitanyaKonda (approved):
I will create a separate issue for removing `ETH_PRIVATE_KEY` and `FROM_ADDRESS` as docker environment variables

PR #198, opened 2021-10-11
Refactor modular division and make point compression consistent

### LINKED ISSUE #158: Refactor helpful point compression for consistency
`edwardsCompress` in `client` and `compressG1` in `optimist` (and duplicated in `common-files`) throw away the x and y coordinate respectively when performing point compression. While these functions are used on different curves, their implementation is curve-agnostic so it would be good to pick a consistent methodology and implementation.

A refactor could also look to elevate this and other shared crypto code to the `./common-files/crypto` folder. As an example, point de-compression could be put in `common-files` and be made more generic by accepting a curve parameter / equation.

IlyasRidhuan:
This simple PR deduplicates `modular-division` from `client` and `optimist`, it is now located in `common-files`.

It also sets point compression to drop the y-coordinate which closes #158


IlyasRidhuan:
Closing this for now - there are good reasons why the point compression is done differently

Issue #177, opened 2021-09-27
Add direct transfers and instant withdrawals to CLI/SDK


Westlad:
Currently I don't belivee this functionality is available in the SDK and, if not, it would be beneficial to add it.

Westlad:
A bit more detail:  The [CLI](https://github.com/EYBlockchain/nightfall_3#command-line-interface) and underlying SDK implements on-chain Deposits, Transfers and Withdrawals.  However, it's possible to directly send a Transfer or Withdraw (but not a Deposit) to a Proposer, to save the cost of posting it on-chain.  This functionality is shown in the [tests](https://github.com/EYBlockchain/nightfall_3/blob/master/test/http.mjs).  At present the CLI and SDK do not implement this functionality and it needs adding.
Likewise, it's possible to instantly withdraw funds without waiting for the 1 week finalisation period, if one pays a fee to a Challenger who is providing this service. This functionality is also in another [test](https://github.com/EYBlockchain/nightfall_3/blob/master/test/instant-withdrawal.mjs) but not yet implemented in the CLI/SDK.

PR #212, opened 2021-10-13
Westlad/instant withdrawals  and bug fixes Pt 1

Westlad:
This is a partial resolution of #177.  It adds instant withdrawals to the SDK (`Nf3 class`). It also fixes a bug in `Nf_3` whereby calls to `Nf3.close` did not shut down open Websockets, and a bug in the off-chain Keccak hash calculation for a Transaction where the proof was not compressed first and, finally, a bug in the on-chain Transaction hash calculation where `compressedSecrets` was not included.  The two methods of hash calculation now agree.

Direct transfers will be added in Pt 2.

To test you have to use the CLI.  There is a new listener which will (uncritically) service all Instant-Withdraw requests.  Run this up with `./liquidity-provider` in a spare window, after you have started nightfall_3 and the `./proposer` listener.  Then you can run the CLI itself (`./nf`). 

To test the Instant-Withdraw command first make sure you have a Withdraw transaction in a Layer 2 block (the fastest way to do this is make a block with two Deposits, then a block with a Deposit and a Withdraw. When that is done, then you can request that it is converted to an Instant Withdraw (`Instant-Withdraw` command). The operation of the command should be self-explanatory and the defaults will work.

### DIFF
--- a/cli/lib/nf3.mjs
+++ b/cli/lib/nf3.mjs
@@ -25,6 +25,8 @@ class Nf3 {
 
   web3;
 
+  websockets = [];
+
   shieldContractAddress;
 
   proposersContractAddress;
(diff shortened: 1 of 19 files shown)


daveroga (approved):
Bug in Nf3.close about remaining websockets opened seems to be resolved with your solution.
Also the solution that adds the instant withdrawal functionality seems ok. I will add the corresponding sdk tests once merged to the master in order to test the integration.

Issue #196, opened 2021-10-07
Split out wallet balances by pkd


Westlad:
Currently, wallet balances are available in the SDK, listed by ERC Address of the token type that the balance is held for.  However, if more than one user is using a particular instance of nightfall-client then all the balances of all the users are summed together.  We should split out the balance by ERC Address _and_ pkd key (or equivalently compressed pkd key) because that key is unique to the user.

PR #214, opened 2021-10-14
David/wallet

druiz0992:
Wallet first version. It requires PR #201 and #213 to work.

to start wallet:
1) Generate browser version of sdk:
```
cd cli
npm install
npm run build
```
2) deploy nightfall (only ganache for now)
3) start proposer from root folder
```
./proposer
```
4) Launch wallet
```
cd wallet
npm install
npm start
```

fixes issue #161


wallet/.eslintignore
```diff
@@ -0,0 +1,2 @@
+node_modules/*
+src/**
```
ChaitanyaKonda, line 2: Is there a reason why we lint ignored the `src`?

wallet/src/constants.js
```diff
@@ -0,0 +1,43 @@
+/* ignore unused exports */
+const DEFAULT_PRIVATE_KEY = '0x4775af73d6dc84a0ae76f8726bda4b9ecf187c377229cb39e1afa7a18236a69e';
```
ChaitanyaKonda, line 2: As you pointed over the sprint call, this would be replaced by metamask wallet passing this key I am assuming

wallet/src/constants.js
```diff
@@ -0,0 +1,43 @@
...
+const DEFAULT_ENVIRONMENT = 'Localhost';
+const NF3_GITHUB_ISSUES_URL = 'https://github.com/EYBlockchain/nightfall_3/issues';
+const DEFAULT_TOKEN_TYPE = 'ERC20';
+const TOKEN_TYPE = {
+  ERC20: 'ERC20',
+  ERC721: 'ERC721',
+  ERC1155: 'ERC1155',
+};
+const DEFAULT_TOKEN_ADDRESS = {
```
ChaitanyaKonda, line 11: We will be supporting any ERC20, ERC721 and ERC1155 contract

wallet/README.md
```diff
@@ -0,0 +1,15 @@
+# Nightfall Wallet
+
+## Dependencies
+### SDK
+Nifghtfall SDK is located in `../cli/lib/nf3.mjs`. To build it:
+```
+cd ..
```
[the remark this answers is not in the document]
druiz0992: Problem is that this is covered in the PR i did to make SDK browser compatible. I will merge this PR with master and then it should work

wallet/src/constants.js
druiz0992: Yes, this is a place holder for now.

wallet/src/constants.js
ChaitanyaKonda: This is a place holder too I presume

wallet/README.md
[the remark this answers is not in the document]
ChaitanyaKonda: I am having the same problem still. 
Doing `cd ..` as suggested in `nightfall_3/wallet/README.md` is taking me to nightfall_3.

Is this perhaps the following 
``
npm run build
``

Westlad:
Although there are no merge conflicts, I think possibly this now doesn't work with the `per user` reporting of wallet balances implemented by the merge of #196.  Suggest a rebase to latest master to check.  Merging current master with this PR breaks it:
image

PR #226, opened 2021-10-18
westlad/instant-withdrawal Pt 2:  added direct transfer and withdraw

### LINKED ISSUE #177: Add direct transfers and instant withdrawals to CLI/SDK

Westlad:
fixes #177 
This PR adds direct transfers and withdrawals.  

The SDK now has functionality (`.addPeer`) to add URLs for Proposers who are prepared to accept transfer and withdrawal transactions directly, rather than requiring them to be posted (notarised) to the blockchain.  Doing this saves the user ~60kGas per transaction.

It's easy to use:  The Proposer's URL is added in the background by the CLI calling `.addPeer` and then one can choose either a direct or on-chain transaction when setting up a transfer or withdrawal in the CLI.  The `.transfer` and `.withdraw` methods of `Nf3` now have a boolean called `offchain` which indicates whether to do the transaction on-chain or directly.

This service is not available for deposits; these must be done on-chain so that the Shield contract can be paid the value of the deposit commitment.

### DIFF
--- a/nightfall-client/src/services/withdraw.mjs
+++ b/nightfall-client/src/services/withdraw.mjs
@@ -111,7 +111,9 @@ async function withdraw(withdrawParams) {
     if (offchain) {
       const peerList = await discoverPeers('Local');
       Object.keys(peerList).forEach(async address => {
-        logger.debug(`offchain transaction - calling ${peerList[address]}/proposer/offchain-transaction`);
+        logger.debug(
+          `offchain transaction - calling ${peerList[address]}/proposer/offchain-transaction`,
+        );
         await axios
           .post(
             `${peerList[address]}/proposer/offchain-transaction`,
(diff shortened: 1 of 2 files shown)


nightfall-client/src/services/withdraw.mjs
```diff
@@ -111,6 +111,7 @@ async function withdraw(withdrawParams) {
     if (offchain) {
       const peerList = await discoverPeers('Local');
       Object.keys(peerList).forEach(async address => {
+        logger.debug(`offchain transaction - calling ${peerList[address]}/proposer/transfer`);
```
daveroga, line 114: The logger message should be '.../proposer/withdraw' instead of '.../proposer/transfer'

Westlad: Good point!

PR #229, opened 2021-10-19
Westlad/balances by pkd

### LINKED ISSUE #196: Split out wallet balances by pkd

Westlad:
fixes #196 

This PR splits out wallet balances by user and ERC contract address.  It can be tested by selecting 'View Wallet Balance' from the CLI.  Previously, all balances were summed for each ERC contract address.  

Note that the test deployment of NF_3 uses a stub ERC contract, rather than real ones and thus the contract address will be the same, irrespective of the token type chosen.  This would not be the case in a real deployment using real ERC contracts.

A 'User' is defined by their unique pkd key (see https://github.com/EYBlockchain/nightfall_3/blob/master/doc/in-band-secret-distribution.md for more details). As the pkd key is actually a curve point, we use Edwards compression to reduce the point to a single number (essentially the y coordinate with a parity bit).

### DIFF
--- a/cli/src/index.mjs
+++ b/cli/src/index.mjs
@@ -127,13 +127,18 @@ async function askQuestions(nf3) {
 Simple function to print out the balances object
 */
 function printBalances(balances) {
+  console.log('BALANCES', balances);
   if (Object.keys(balances).length === 0) {
     console.log('You have no balances yet - try depositing some tokens into Layer 2 from Layer 1');
     return;
   }
-  const table = new Table({ head: ['ERC Contract Address', 'Layer 2 Balance'] });
-  table.push(balances);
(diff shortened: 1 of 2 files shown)


ChaitanyaKonda (approved):

daveroga (approved):
Seems ok. I'll review it deeply in sdk testing once merged.

druiz0992 (approved):

PR #252, opened 2021-10-26
fix: previous proposer check before enqueue

### LINKED ISSUE #227: fix broken test in master
The final 'good block' test, that runs after the negative tests, is broken currently and needs fixing.  It currently has a block number out of order error. Additionally, this test does not appear to fail under these conditions and so the error was not picked up by automated testing. That needs fixing too.

dwebchapey:
This fixes #227 

Two instances of `blockassembler`  was triggered when the proposer before and after rollback is the same. The fix is to not enqueue `blockassembler` from `newCurrentProposerEventHandler` in this scenario.


dwebchapey:
Duplicate to #261

Issue #32, opened 2021-05-27
Remove Nightfall Client dependency on Nightfall Optimist 


ChaitanyaKonda:
Today historic root block hash call is retrieved from NO by NC

ChaitanyaKonda:
Reintroduced

PR #261, opened 2021-10-29
Enqueue Block Assembler

### LINKED ISSUE #227: fix broken test in master

IlyasRidhuan:
This enqueues the block assembler using the `end` emitter from `queue`. Essentially, we try to enqueue the assembler when the queue is empty.

A few things to note:

- We check the `proposer.isMe` condition in the emitter to fail fast. Otherwise, a non-proposer will constantly try to re-enqueue and eventual consume their memory. We should investigate if long periods of re-enqueueing (i.e. is the proposer just does not have transactions to submit) causes memory issues.

- Note that the `proposer.isMe` check in `block-assembler` is now redundant, however removing that args input will require a bit more re-architecting so I've left it for now.

- I've been able to run the tests without any errors, however i can see the "Out of order block" error happen on multiple consecutive `npm t` runs (without restarting).  This is harder to debug as `client` is potentially very out of sync at this point. Another reason why this could be happening is that we happen to have a `Block-assemble` event queued in front of a `Rollback` event, therefore we try to assemble the block before processing the `Rollback` (even though on-chain the `Rollback` has completed). We could fix this by having logic that purges the assemble events if a Rollback is "received" (not processed).

- Finally, there is a new `after` hook in the `neg-http` tests, the websocket was quitting too early. This hook will capture the final 'good' block that is proposed (it was being dropped before). This was we do not immediately get an "Out of order block" error if we try and run the tests again without restarting.

Overall this greatly simplifies the enqueuing code and makes it more consistent. Closes #227

### DIFF
--- a/nightfall-optimist/src/classes/block.mjs
+++ b/nightfall-optimist/src/classes/block.mjs
@@ -3,12 +3,11 @@ An optimistic layer 2 Block class
 */
 import config from 'config';
 import Timber from 'common-files/classes/timber.mjs';
-import { getContractInstance } from 'common-files/utils/contract.mjs';
 import Web3 from 'common-files/utils/web3.mjs';
 import { compressProof } from '../utils/curve-maths/curves.mjs';
-import { getLatestTree } from '../services/database.mjs';
+import { getLatestTree, getLatestBlockInfo } from '../services/database.mjs';
 
-const { ZERO, PROPOSE_BLOCK_TYPES, STATE_CONTRACT_NAME } = config;
+const { ZERO, PROPOSE_BLOCK_TYPES } = config;
 
 /**
 This Block class does not have the Block components that are computed on-chain.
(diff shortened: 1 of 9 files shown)


IlyasRidhuan:
Hmm - i just did a quick rebuild and it ran without an error on mine. I'm assuming the error would be from when `client` is getting the blockNumberL2 from `optimist` via the call `getBlockAndTransactionsByRoot`. 

Two possible situations is that the block **never got saved** by `optimist` (the logs should reveal if it was invalid) or that the block **hasnt been saved** (should be able to check if the `blocks_collection` has the block). 

If it is the latter then it's likely a race between the triumvirate (optimist--client--timber)  then #32 should help eventually.

Westlad (approved):

ChaitanyaKonda (approved):

PR #268, opened 2021-11-04
Add queue to client

### LINKED ISSUE #230: Nightfall Client needs a queue for events
Similarly to `optimist`, `client` needs a queue to ensure synchronous processing of events. We should elevate the `queue` code from `optimist` and move that into `common-files`.

1. Blockchain events (e.g. `blockProposed`, `Rollback`) should be enqueued
2. Some user-driven events such as `/transfer` should **maybe** also be enqueued (or respect the queue)

(2) requires a bit more understanding of the implications if a `/transfer` or `/withdraw` is called while a blockchain event is being processed.

IlyasRidhuan:
This elevates the queue implementation out of `optimist` into `common-files` so it also be used in `client` without duplication.

Note that to do this quickly, we exploit the fact that `common-files` is copied into the `client` and `optimist` during the build. This allows us to keep the `queues` array inside the `event-queue` file (as there is a fresh instantiation in each copy of the `event-queue` file in both `client` and `optimist`).

Long-term we can certainly do better than this and should look at a way to instantiate the `queue` in the `main` files for `optimist` and `client` rather than instantiating in `common-files` and copying it across. This provides some benefits in terms of flexibility on setup (e.g. `client` does not need two queues) and minimises any potential issues with global mutation and the copy process.

P.S there is a slight bump in some packages as a result of fixing some high-risk vulnerabilities with `npm audit fix`

This closes #230

### DIFF
--- a/common-files/utils/event-queue.mjs
+++ b/common-files/utils/event-queue.mjs
@@ -1,3 +1,5 @@
+/* eslint import/no-extraneous-dependencies: "off" */
+
 /**
 If we're changing the layer 2 state, we want to make sure that we can complete
 that change before the state is further altered by incoming events.  A good
@@ -49,7 +51,7 @@ processed.  It's useful if you want to ensure that Nightfall has had an opportun
 to update its database with something that you know has happened on the blockchain
 but that Nightfall may not have processed yet, because it's still in the event queue.
 */
-export function flushQueue(priority) {
+function flushQueue(priority) {
   const p = new Promise(resolve => {
     queues[priority].push(cb => {
       cb(null, resolve());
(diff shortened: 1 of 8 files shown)


ChaitanyaKonda (approved):

ChaitanyaKonda:
This looks good
I will make an issue for the following
> Long-term we can certainly do better than this and should look at a way to instantiate the queue in the main files for optimist and client rather than instantiating in common-files and copying it across. This provides some benefits in terms of flexibility on setup (e.g. client does not need two queues) and minimises any potential issues with global mutation and the copy process.

Is there any reason why we didn't implement the queue directly in main for client ?

Westlad (approved):

PR #276, opened 2021-11-09
Daveroga/api status widthdraws

daveroga:
This PR provides method in the SDK to get the pending withdraws for a wallet and indicates wether they are valid for finalising the withdraw (older than 7 days).

When we take a commitment from the nightfall-client database and use it to form a withdraw transaction, we make the commitment as nullified (markNullified(oldCommitment)).  The difficulty is that if you later on query the db, you won't easily know if the nullified commitment was the input to a withdraw transaction or a transfer transaction - or indeed, what the transactionHash was.

To solve this I've followed this steps:

- Modify `markNullified` to also store the transaction object in a property `transactionNullified` of the commitment document.
- In the `withdraw` and `transfer` methods that call this `markNullified` pass this transaction as parameter (Note that for this case we only need withdraws but we have also manage transfer for future cases).
- Create view function `isValidWithdrawal`  in `Shield.sol` contract similar to what `finaliseWithdrawal` does but only to check if a transaction is valid for finalising the withdrawal.
- Create method `getPendingWithdraws` in the SDK nf3 to get pending withdraws for a pkd. This method will get all pending withdraws in the db based on commitments and filtering by `transactionType=3` that is the type corresponding to withdraws. After that we will check all this pending withdraws for being valid for finalising with the `isValidWithdrawal` method of the `Shield.sol` contract and will return the collection of this pending withdraws with a `valid` additional property that will be `true` in case of being valid to finalise the withdraw and `false` otherwise.
- Added corresponding tests to `test/e2e/nightfall-sdk.test.mjs` to test this new feature that can be tested with `npm run test-e2e`.

Example
```json
{
    "compressedPkd": "0xa47e1471eb85eaf82e72ef8aad52dff32a73e33dccfd2dfde338f0dcfa2585c4",
    "ercAddress": "e1b7b854f19a2cebf96b433ba30050d8890618ab",
    "balance": 10,
    "transactionNullified": {
      "fee": "0x0000000000000000000000000000000000000000000000000000000000000001",
      "historicRootBlockNumberL2": "0x000000000000000000000000000000000000000000000000000000000000000a",
      "transactionType": "0x0000000000000000000000000000000000000000000000000000000000000003",
      "tokenType": "0x0000000000000000000000000000000000000000000000000000000000000000",
      "publicInputHash": "0x008ac05bf45942bcd8394a61476e542ea481003399ba6a5c31b16fc92ad98004",
      "tokenId": "0x0000000000000000000000000000000000000000000000000000000000000000",
      "value": "0x000000000000000000000000000000000000000000000000000000000000000a",
      "ercAddress": "0x000000000000000000000000e1b7B854F19A2CEBF96B433ba30050D8890618ab",
      "recipientAddress": "0x0000000000000000000000009C8B2276D490141Ae1440Da660E470E7C0349C63",
      "commitments": [Array],
      "nullifiers": [Array],
      "compressedSecrets": [Array],
      "proof": [Array],
      "transactionHash": "0x4556e6c35cb00d39e876197e40bffefd943f4eee617c06a599ec388aeb82d95b"
    }
    "valid": false
}
```

### DIFF
--- a/cli/lib/nf3.mjs
+++ b/cli/lib/nf3.mjs
@@ -575,7 +575,7 @@ class Nf3 {
   @method
   @async
   @return {Promise} A promise that resolves to an axios response.
-  */  
+  */
   async registerChallenger() {
     return axios.post(`${this.optimistBaseUrl}/challenger/add`, { address: this.ethereumAddress });
   }
--- a/cli/lib/nf3.mjs
+++ b/cli/lib/nf3.mjs
@@ -641,6 +641,39 @@ class Nf3 {
   }
 
   /**
+  Get if it's a valid withdraw transaction for finalising the
+  withdrawal of funds to L1 (only relevant for ERC20).
+  @method
+  @async
+  @param {string} withdrawTransactionHash - the hash of the Layer 2 transaction in question
+  */
+  async isValidWithdrawal(withdrawTransactionHash) {
+    let res;
+    let valid = false;
+
+    try {
+      res = await axios.get(
+        `${this.optimistBaseUrl}/block/transaction-hash/${withdrawTransactionHash}`,
+      );
+    } catch (e) {
+      // transaction is not in block yet
+      valid = false;
+    }
+
+    if (res) {
+      const { block, transactions, index } = res.data;
+      res = await axios.post(`${this.clientBaseUrl}/valid-withdrawal`, {
+        block,
+        transactions,
+        index,
+      });
+      valid = res.data.valid;
+    }
+
+    return valid;
+  }
+
+  /**
   Returns the balance of tokens held in layer 2
   @method
   @async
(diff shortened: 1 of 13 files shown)


cli/lib/nf3.mjs
```diff
@@ -666,6 +699,38 @@ class Nf3 {
...
+  /**
+  Returns the pending withdraws commitments
+  @method
+  @async
+  @returns {Promise} This promise resolves into an object whose properties are the
+  addresses of the ERC contracts of the tokens held by this account in Layer 2. The
+  value of each propery is an array of commitments originating from that contract.
+  */
+  async getPendingWithdraws() {
+    const res = await axios.get(`${this.clientBaseUrl}/commitment/withdraws`);
```
druiz0992, line 711: Is it possible to move the functionality of getPendingWithdraws to the endpoint rather than to the SDK? The SDK should only filter per pkd. What do you think?

daveroga: Yes, I've refactored this doing all the work in the endpoint.

Westlad (approved):

druiz0992 (approved):

Issue #241, opened 2021-10-22
Integrate local timber into client


IlyasRidhuan:
Similar to #185, , the new local timber instance (`timber-mist`) should be integrated into `client`. 

This has multiple benefits such as simpler internal syncing during rollbacks  - as well as forming a key part to a robust solution to  #231.

PR #278, opened 2021-11-11
Add second historic root

IlyasRidhuan:
This PR changes the `double_transfer` circuit behaviour such that the `siblingPaths` of each input can be verified against different `roots`. This enables the pre-computation of static sibling paths for commitments. This in-turn simplifies the spending of commitments and eliminates the need for `client` to store the entire merkle tree of all commitments.

An effect of this change is the addition of an extra `blockNumberL2` to the on-chain transaction structure and the need to modify checks involving the `PublicInput`.

Finally, this PR fixes a minor type-casting issue involving off-chain transfers in `nightfall-optimist/src/routes/proposer.mjs`

This is the last PR necessary before #241 can be completed.


Westlad (approved):

nightfall-deployer/contracts/Challenges.sol
```diff
@@ -138,20 +154,70 @@ contract Challenges is Stateful, Key_Registry, Config {
...
+      blockL2ContainingHistoricRoot,
+      transactionsOfblockL2ContainingHistoricRoot,
+      blockNumberL2ContainingHistoricRoot
+    );
+    state.isBlockReal(
+      blockL2ContainingHistoricRoot2,
+      transactionsOfblockL2ContainingHistoricRoot2,
+      blockNumberL2ContainingHistoricRoot2
+    );
+    // check the historic root is in the block provided.
```
ChaitanyaKonda, line 201: It might be missing this?
```
    require(
      transactions[transactionIndex].historicRootBlockNumberL2[1] ==
      blockNumberL2ContainingHistoricRoot2,
      'Incorrect historic root block'
    );
```

ChaitanyaKonda (changes requested):

nightfall-deployer/contracts/Challenges.sol
IlyasRidhuan: yep that's a good catch!

ChaitanyaKonda (approved):

PR #279, opened 2021-11-11
client chain-reorg functionality

### LINKED ISSUE #205: Create basic chain reorg code for nightfall-client
Whilst Optimist and Timber are able to reset their Layer 2 data in response to a Layer 1 chain reorganisation, Client does not yet have this capability and it should be added.  It should follow the documented approach added by PR #200.

Westlad:
fixes #205 

This PR provides the capability to nightfall-client to cope with a chain reorganisation and to update its L2 database to reflect the new truth of the cannonical chain.  It can currently handle removal of `blockProposed` events by a chain-reorg but not removal of an L2 `rollback` event.  That will be a later PR. 

A new eventHandler has been added that runs when removal of a `blockProposed` event occurs and performs the L2 database updates.  These are actually just resetting information for commitments and nullifications that have been marked as on-chain.  The logic for this approach is discussed [here](https://github.com/EYBlockchain/nightfall_3/blob/westlad/client-reorg/doc/chain-reorgs.md#layer-2-client).

To test, you need to run up the local Geth blockchain, which has the ability to force chain-reorgs by pausing some of the docker containers.  Do:
```sh
./geth-standalone -s #starts the chain
./geth-standalone -l #starts logging to the terminal
```
Wait until the DAG generation is complete then run up nightfall_3  in another terminal with options to connect it to the Geth chain (`./start-nightfall -l -s`). Then finally run the test in a third terminal: `npm run test-chain-reorg`.

### DIFF
--- a/nightfall-client/src/services/commitment-storage.mjs
+++ b/nightfall-client/src/services/commitment-storage.mjs
@@ -37,9 +37,7 @@ export async function storeCommitment(commitment, nsk) {
   };
   // a chain reorg may cause an attempted overwrite. We should allow this, hence
   // the use of replaceOne.
-  return db
-    .collection(COMMITMENTS_COLLECTION)
-    .replaceOne({ _id: commitment.hash.hex(32) }, data, { upsert: true });
+  return db.collection(COMMITMENTS_COLLECTION).insertOne(data);
 }
 // function to update an existing commitment
 export async function updateCommitment(commitment, updates) {
--- a/nightfall-client/src/services/commitment-storage.mjs
+++ b/nightfall-client/src/services/commitment-storage.mjs
@@ -37,6 +37,14 @@ export async function storeCommitment(commitment, nsk) {
   };
   return db.collection(COMMITMENTS_COLLECTION).insertOne(data);
 }
+// function to update an existing commitment
+export async function updateCommitment(commitment, updates) {
+  const connection = await mongo.connection(MONGO_URL);
+  const db = connection.db(COMMITMENTS_DB);
+  const query = { _id: commitment._id };
+  const update = { $set: updates };
+  return db.collection(COMMITMENTS_COLLECTION).updateOne(query, update);
+}
 
 // function to get count of commitments. Can also be used to check if it exists
 export async function countCommitments(commitments) {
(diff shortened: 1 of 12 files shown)


daveroga (approved):

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -28,11 +28,23 @@ export async function storeCommitment(commitment, nsk) {
...
     isNullified: commitment.isNullified,
-    isNullifiedOnChain: Number(commitment.isNullifiedOnChain),
+    isNullifiedOnChain: Number(commitment.isNullifiedOnChain) || -1,
     nullifier: nullifierHash,
     blockNumber: -1,
   };
-  return db.collection(COMMITMENTS_COLLECTION).insertOne(data);
+  // a chain reorg may cause an attempted overwrite. We should allow this, hence
+  // the use of replaceOne.
+  return db
```
IlyasRidhuan, line 37: The `storeCommitment` call that we are trying to guard for here is the one in `blockProposed` handler. However, I believe it will only get called in the `else` condition of `countCommitments(transaction.commitments)) > 0`. That is, we only call `storeCommitment` if we do not have the commitment already in our local DB.

Is there a situation i've missed where we actually replace an existing document with this call?

nightfall-optimist/src/services/database.mjs
```diff
@@ -466,7 +474,8 @@ export async function saveTree(blockNumber, blockNumberL2, timber) {
   const connection = await mongo.connection(MONGO_URL);
   const db = connection.db(OPTIMIST_DB);
   return db.collection(TIMBER_COLLECTION).insertOne({
-    _id: blockNumber,
+    _id: timber.root,
```
IlyasRidhuan, line 477: we'll need to re-key this with something under than `root`. A block full of withdrawals will have the same root and throw a duplicate error. Perhaps `blockNumberL2` will be better?

nightfall-client/src/services/commitment-storage.mjs
Westlad: Yes, you're quite right.  This was an unnecessary protection.  I've removed it

nightfall-optimist/src/services/database.mjs
Westlad: Oh blast.  I forgot about the dastardly 'block of withdrawals'.

IlyasRidhuan (approved):

PR #284, opened 2021-11-13
David/wallet keys

### LINKED ISSUE #248: Wallet - implement pkd flow
Flow:
1. User connects to wallet using Metamask
2. Metamask signs a constant meesage. The result is hashed. This hash is sent to the endpoint as a seed to generate the keys.
3. The endpoint returns N set of keys to the user.
4. The user (via the wallet) can select which key to use to operate.

This way, the user doesn't need to enter any password to retrieve the keys, and the wallet doesn't need to store any key in local storage.

Issue considered done when:
- Wallet needs to implement this flow.
-  Additionally, under `accountInfo` or in main panel, the user may select/view default pkd.
- Deposit transaction modal will show pkdId and not pkd.
- `accountInfo` will show which accounts are active (either there is l2Balance or existing commitment)

druiz0992:
Added key flow to the wallet.

Closes #248


daveroga (approved):

cli/lib/nf3.mjs
```diff
@@ -95,7 +86,10 @@ class Nf3 {
     if (typeof this.ethereumSigningKey === 'string') {
       this.ethereumAddress = await this.getAccounts();
     }
-    return this.subscribeToIncomingViewingKeys();
+    // set zkp keys from mnemonic if provided
+    if (typeof mnemonic !== 'undefined') {
+      this.setzkpKeysFromMnemonic(mnemonic, 0);
```
Westlad, line 91: I think you need to await `this.setzkpKeysFromMnemonic` or return it as a promise, otherwise `init` will return before the `setzkpKeysFromMnemonic` has finished.  Thus code that is awaiting `init` may not run correctly.

cli/lib/nf3.mjs
```diff
@@ -119,6 +113,28 @@ class Nf3 {
...
+      this.mnemonic.phrase = mnemonic;
+    }
+    this.mnemonic.addressIndex = addressIndex.toString();
+    this.zkpKeys = (
+      await axios.post(`${this.clientBaseUrl}/generate-keys`, {
+        mnemonic: this.mnemonic.phrase,
+        path: `m/44'/60'/0'/${this.mnemonic.addressIndex}`,
+      })
+    ).data;
+    this.subscribeToIncomingViewingKeys();
```
Westlad, line 137: the `this.subscribeToIncomingViewingKeys()` Promise needs to be returned, otherwise the code which is awaiting `setzkpKeysFromMnemonic` will run before `subscribeToIncomingViewingKeys()` has finished.

Westlad (changes requested):

cli/lib/nf3.mjs
druiz0992: You're right.

cli/lib/nf3.mjs
druiz0992: Ok, Should i do the same change here? ```  setzkpKeys(keys) {
    this.zkpKeys = keys;
    this.subscribeToIncomingViewingKeys();
  }
```

Westlad: Yes, good spot.

Westlad (approved):

PR #286, opened 2021-11-15
Integrate timber into client

### LINKED ISSUE #231: Race condition when syncing client and timber during rollback
When a rollback occurs, `client` needs to wait for `timber` to rollback first before it can begin its own rollback. This is because it need the post-rollback commitment tree information to know which commitments may no longer be valid.

The syncing is done by periodically comparing the `leafCount` the rollback is meant to be at and the `timber.leafCount`.

However, it may be the situation that other events (e.g. `blockProposed`) happen soon after the `Rollback`. In this scenario the `timber.leafCount` may have been rolled back and then moved again due to the new event - all before the next check by `client`.

This leaves `client` in an infinite loop as it will believe that `timber` has still not performed the `Rollback`. Furthermore, given the asynchronous code processing and the lack of a queue (currently), this infinite loop does not block the processing and `client` will continue to perform other tasks such as transfers, without correcting it's database.

The long term fix for this will occur with the inclusion of a queue (#230) and the addition of the local timber library to `client`

IlyasRidhuan:
This PR completes the integration of `timber` into `client`.

The main changes are twofold: the use of a local `TIMBER_COLLECTION` to store the frontier and roots after each block (similar to the change in `optimist`)  and the calculation of a static `siblingPath` during the `blockProposed` handler for each user-owned commitment .

Some smaller changes including: small default value changes to the `timber` class, and a change in the `rollback-resync` test.

This closes #231  and #241.

Note: Some syncing code remains in `optimist`, while this can be removed, it results in a lot of pollution in the logs as `timber` continues to listen but be out of sync. The plan would be to remove all of this when `timber` is completely removed as a service.


Westlad (approved):

IlyasRidhuan:
Good catch! That is indeed the issue (i.e. a change from the default value from `0` -> `'0'`.

I will push a quick fix

MirandaWood (approved):

PR #291, opened 2021-11-15
Docker images and release

### LINKED ISSUE #273: Docker images for Client, Optimist, Deployer, Zokrates Worker
These images will be later used inside a github action to create and provide updated versions of each service

MirandaWood:
This PR adds Dockerfiles copying local dependencies into build images and github actions to push that image to the new github repo, `ghcr.io`. I removed binds where possible, but shared volumes (e.g. `build`, `proving-files`) are still required.
I've already pushed the images manually and created `docker-compose.images.example.yml`, which uses them. If you replace `docker-compose.yml` with this file everything should still work.

Note that the actions file is not tested, however a test run failed on the actions tab of the repo on a previous push. Hopefully it will run on this PR creation...

Closes #273.

### DIFF
--- a/.github/workflows/publish-docker-images.yml
+++ b/.github/workflows/publish-docker-images.yml
@@ -18,7 +18,7 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
     - name: Login to docker registry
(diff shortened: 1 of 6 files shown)


Westlad (approved):
I assume we haven't created a Timber image because we're removing Timber  as a separate service very soon.

druiz0992 (approved):

Issue #297, opened 2021-11-16
Wallet - Reject login if user fails to sign initial metamask message


druiz0992:
When rejecting metamask signature after selecting backup mnemonic, user can log in but backup is not done. It would be better in this case to reject login

Issue #294, opened 2021-11-16
Wallet - Review flow when Metamask plugin is not available


druiz0992:
If Metamask plugin is not available, current wallet allows to log in, even though no account exists. If Metamask plugin doesn't exist, login should not be possible.

Issue #293, opened 2021-11-16
Wallet - Mnemonic backup doesn't work all the times


druiz0992:
When requesting to do a backup of the mnemonic, 50% of the times the backup is not attempted.

PR #301, opened 2021-11-17
Fix multiple wallet minor issues

### LINKED ISSUE #292: Return this.subscribeToIncomingViewingKeys Promise
Must return `this.subscribeToIncomingViewingKeys();` Promis, otherwise the code which is awaiting setzkpKeys will run before subscribeToIncomingViewingKeys() has finished.
```
setzkpKeys(keys) {
    this.zkpKeys = keys;
    this.subscribeToIncomingViewingKeys();
  }
```

druiz0992:
PR fixes some minor issues found during testing.

Close #292, #293, #294, #297


daveroga (approved):

ChaitanyaKonda:
Does this need to be tested?

druiz0992:
@ChaitanyaKonda What does need to be tested?

ChaitanyaKonda:
For example, everytime a PR is pushed, the backend code gets tested against http.mjs and neg-http.mjs using github action. This makes it easier to review too and understand what features were added from the test files. 

Considering this is front end, one way we could do the same is to test the front end from the wallet interface for each PR.
It makes easy to understand the functionality added and provide any relevant feedback accordingly.

I could try to test from wallet interface from the instructions for bringing up the wallet as provided in the previous PR but it does not seem to work for me.

LijuJoseJJ (approved):

PR #312, opened 2021-11-19
Remove Optimist dependency in client

### LINKED ISSUE #32: Remove Nightfall Client dependency on Nightfall Optimist

### LINKED ISSUE #232: Instant withdrawal failing when error status code 500
image

This could be due to the fact that a liquidity provider has not yet seen the commitment that needs to be withdrawn. A deeper investigation to under the source might be required.

IlyasRidhuan:
This is quite a large PR but unfortunately it is hard to split into multiple PRs due to the changes being tightly coupled to success in the tests.  I've broken this PR into commits that (hopefully) can be reviewed individually.

This closes #32, #287, and through the test fixes,  also closes #232.

Hopefully this description of each commit will help with review:

1. `refactor: move classes and related libs to common and small fixes`
To add the `transaction` and `block` collection, we need access to some hashing and `compressProof` functionality, therefore we need to move some files (i.e. `Fq`, `curve-maths/utils`) into common-files. This also contains a fix described in #306 .

2. `feat: add standard database operations`
This adds a lot of the database operations to interact with the `transaction` and `block` collections. This is mostly taken from the `optimist` instance.
Note: The one difference is that in the `block` collection we do not store the `blockHash`. There is no good reason for this other than it would involve also elevating the `Block` class from `optimist` to the `common-files` so we would have access to `Block.calcHash`. We do not need the `blockHash` in `client` so it is left out for now.

3. `feat: remove optimist dependency` --> `feat: update cli to remove optimist dependency`
This is the bulk of the work, where we remove the `optimist` dependency from `client`. There are a lot of minor changes which pertain to saving the appropriate transactions and blocks during the `blockProposedHandler` and retrieving the relevant withdraw transactions in `commitment-storage`.
A repeated minor change is the padding in `ercAddress`. An ethereum address is 20bytes, however, because of the ease with interacting with `geth` this is padded to 32 bytes and stored as such.  However, when we use `ercAddress` in `getWalletCommitments` and `getWithdrawCommitments` we use `BigInt(ercAddress).toString(16)` which contains no padding. This creates annoying compatibility issues when we use the result of `getContractAddress` as a lookup. This is now changed to be "address padded".

4. `test: fix or update tests`
The tests are updated and/or fixed to reflect the removal of `optimist` during `/finalise-withdrawal` and `instantWithdrawal`.


MirandaWood (approved):
All tests pass on my end and commitment storage changes look good, so approving. Not too familiar with withdraw and block proposed code so hopefully the 2nd reviewer can confirm I haven't missed anything!

Westlad (approved):

PR #316, opened 2021-11-22
Chait/adversary

ChaitanyaKonda:
This PR introduces a service called nightfall-adversary which will only be used for writing bad block tests. This removed the need for creating bad blocks at an application level and makes it easier to write positive and negative tests for nightfall with its native optimist services (suggestion to merge adversary into optimist below). Adversary is similar to optimist in most ways except at any time can be given a block config by calling the endpoint `/proposer/config` which provides a list of block types to be created by the adversary

The various types are 
- ValidBlock’’
- IncorrectRoot'
- 'DuplicateTransaction'
- 'InvalidDepositTransaction'
- 'IncorrectHistoricRoot'
- 'IncorrectProof'
- 'DuplicateNullifier'
- 'IncorrectLeafCount'

An example of how the block config could look
```
const blockConfig = [
  'ValidBlock',
  'ValidBlock',
  'IncorrectRoot',
  'DuplicateTransaction',
  'InvalidDepositTransaction',
  'IncorrectHistoricRoot',
  'IncorrectProof',
  'DuplicateNullifier',
  'IncorrectLeafCount',
  'ValidBlock',
];
```

Adversary also has the capacity to ignore `IncorrectHistoricRoot` and `DuplicateNullifier` if there are no transfer or withdraw transactions in mempool. It will retain this config to apply on a later block.
If no block config is submitted or if the block config is empty, adversary will continue to make valid blocks.

We will retain the current neg-http.mjs file to be used in GitHub action tests such that any new PR will still need to work against the current successful negative tests. These negative tests will be removed and replaced eventually by tests that check only the client’s balance state (as defined for ping pong tests)

To note: Adversary needs the same updates that will also be made to optimist.
Another easier approach could be to merge adversary into optimist, this might not be desirable in a public release, but adversary does not require the changes created in optimist to be duplicated within it. These files can be removed after testing.
If this is not desirable, then once optimist image is created, adversary’s image can be created from optimist’s by copying the relevant adversary files to optimist. This will need some refactoring of adversary such that any changes made to the same files in both will now need to be made in a different file in adversary such that adversary does not overwrite optimist’s files.
Tacking with one of these approaches will be created as a new issue.

 
What this PR does
- Introduces nightfall adversary
- A new test file that uses SDK to write tests from client perspective
    - This file also sends a block config file to nightfall adversary
    - This uses different addresses for different parties
    - The tests send repeated transfer transactions from client whilst the adversary is making good blocks or bad blocks 
    - It provides new utils
        - for test that registers proposer and waits for this proposer to be current proposer again after they are removed due to a successful challenge. Helpful for testing purposes. `waitForProposer`
        - for waiting for sufficient balance from pending transactions or creating a deposit with sufficient balance, that could be used before a transfer transaction.  `waitForSufficientBalance`
- Adds IS_CHALLENGER to the intialClientSync to fix optimist instances that are not instantiated as challengers to not create a challenge. Before this fix, they try to create a challenge and later fail to send to a connected challenger app and log an error for this
- Also adds new functions to the SDK  `getCurrentProposer`, `getZKPKeys`
- Changed `init` in SDK to take addressIndex parameter too. This is required to generate different addresses from same mnemonic. Previously, it was set to 0 which means when the same mnemonic is passed to `.init()` for different apps, it will generate them same addresses
- Fixes a bug in SDK that only submits a transaction that commits a challenge but does not reveal the challenge
- This PR also fixes the check transaction in transaction checker such that 
    - For ERC20, the tokenId must be zero
    - For ERC721, the value must be zero
    - For ERC1155, no restrictions on tokenId or value because even zero for both is acceptable  Earlier the wrong checks were enforced for tokenId and value independent of type of token 
    - the tokenType must be amongst ERC20 or ERC721 or ERC1155 for all types of transactions. Earlier this check was only imposed for deposit and withdraw wrongly
    - In transaction checker, for validity of various types of transactions, the check for historicRootBlockNumberL2 value being empty or otherwise is missed in single transfer and withdraw for the 2nd element in historicRootBlockNumberL2 array
    - Removed public input hash checking in transaction-checker because the tests for manipulating duplicate nullifiers, incorrectRoots or incorrectHistoricRoots will involve editing the public input hash of transaction, if not then public input hash challenge is raised for all of these. Considering that public input hash is being removed, the adversary was written with this in perspective and hence we removed public input hash checker 
- Remnants of public input hash check will be removed soon in the relevant PR being worked on by Miranda

Testing instructions
This will automatically be tested by GitHub action using the `npm run adv` script

Nightfall Adversary is a replica of Nightfall optimist. 
For easier review of nightfall adversary code, only these files within  nightfall adversary  were changed. The rest need not be reviewed.

- Nightfall-adversary/src/routes/proposer.mjs ( get `/current-proposer`, post `/config`)
- Nightfall-adversary/src/block-assembler.mjs (`concatBlockConfig()`, edited `makeBlock()`)
- Nightfall-adversary/src/database.mjs (`getTransactionsOnChain()`, replaced `OPTIMIST_DB` with `ADVERSARY_DB`)
- Nightfall-adversary/src/state-sync.mjs (replaced `await startMakingChallenges();` with `if (process.env.IS_CHALLENGER === 'true') await startMakingChallenges();`)
- Nightfall-adversary/src/transaction-checker.mjs (edited `checkTransactionType()`, commented out `checkPublicInputHash()`, added `checkHistoricRoot()` to `checkTransaction()`)
- Nightfall-adversary/src/classes/invalid-block.mjs (entirely new file)
- package.json (added new script `adv`)


Westlad:
Could we import the nightfall-optimist code that is in the nightfall adversary directory? That would keep the code DRY.  This is difficult to do currently because the docker files are not at the top of the repo but, if they were moved, it would be possible to copy in the correct code to adversary and only retain modules where there is a difference.

ChaitanyaKonda:
> Could we import the nightfall-optimist code that is in the nightfall adversary directory? That would keep the code DRY. This is difficult to do currently because the docker files are not at the top of the repo but, if they were moved, it would be possible to copy in the correct code to adversary and only retain modules where there is a difference.

When we originally had a discussion on stand up we decided to keep them separate but to be merged into master instead of having adversary in it's own branch. Later after writing the code, it seemed like the best thing to do was to create adversary from the same image as optimist and copy over the diff set of files for adversary during this image creation. So I created a separate [issue](https://github.com/EYBlockchain/nightfall_3/issues/319) for this already as this issue was taking too long and I had to continuously catch up to the changes everyone was making. I can pick this issue to up next sprint. 

https://github.com/EYBlockchain/nightfall_3/issues/319

MirandaWood (approved):

ChaitanyaKonda:
Closing this branch. A new PR with part of adversary code will be created as part of #319.
Other non adversary fixes, will come in their own branch

Issue #285, opened 2021-11-14
Add approve functionality to SDK


druiz0992:
When doing a deposit of an ERC20, ERC721 and ERC1155, token owner needs to first approve Shield contract.

PR #322, opened 2021-11-22
David/approve

### LINKED ISSUE #249: Wallet - add info to local storage
Following information needs to be added to localStorage:
- default pkd
- existing tokens
-  contact pkd

info is stored per Ethereum address.

druiz0992:
Call ERC `approve` function before deposit. Allowance is set to infinite after first deposit to avoid keep asking.

There is a special change that I like reviewed in the Shield.sol contract. In order to be able to receive ERC721 and ERC1155 tokens, Shield.sol contract needs to implement `onERC721Received` and `onERC1155Received` functions.

 Fixes #249 #285 

---

Simplest way to test this PR is by using `cli`. It automatically uses real ERC tokens. For ERC721, use ids 0-3. For ERC1155, use Ids 0-4.


cli/lib/nf3.mjs
```diff
@@ -222,6 +225,30 @@ class Nf3 {
...
   }
 
+  /**
+   Sends an approve transaction to an ERC20/ERC721/ERC1155 contract for a certain amount of tokens
+   * @param {string} ercAddress - The Ethereum address of the transaction sender
+   * @param {string} tokenType  - Type of token 
+   * @param {string} value  - Amount of tokens to be approved (in Wei)
+   * @returns {Promise} transaction
+   */
+  async approve(ercAddress, tokenType, value) {
```
Westlad, line 235: What's the distinction between this function in `nf3.mjs` and the one in `tokens.mjs`? Do we need both?

druiz0992: My mistake. Approve function in nf3.mjs should not have been committed.

Westlad (approved):

daveroga (approved):

Issue #266, opened 2021-11-03
Prevent block assembly when a bad block is detected


IlyasRidhuan:
When a bad block is detected there should be a corresponding challenge (or challenges) sent to the main chain containing the fraud proof that trigger a rollback.

If the `currentProposer` knows that a bad block is in the L2 chain, they should not make any blocks until after the challenge and rollback occurs - otherwise they run the risk of their block being pruned as part of the eventual Rollback.

**Note: A change like this will heavily constrain the extent of challenge + Rollback testing we can do because we will no longer be able to create a L2 chain where a bad block is not on the tip.** This would thus require an adversarial proposer for testing (possibly related #255 )

PR #330, opened 2021-11-25
Stop Proposing on top of bad blocks and clean up on proposer rotation

### LINKED ISSUE #139: Cleanup check for previous proposer
The fix in #102  and the subsequent PR for the race condition #111 highlights the following issue:

1. The way block-assembler works (i.e. it runs async-ly in the background rather than in a queue) results in a race condition between the events processed in queue.

2. The race condition here is that there are situations where we are simultaneously the proposer AND the previous proposer , so this code ends up deleting blocks made by the block assembler!! 

3. This current fix basically prevents the code in the if condition from being run (which is bad). 

4. A longer term solution would be fix the inherent design that causes race conflicts between the block-assembler and events in the queue.

5. `!proposer.isMe` can be removed if the above point is solved.

IlyasRidhuan:
This PR closes #139 and #266. These are combined as a single PR because the former will not past tests without the latter. This is not due to the PR itself but the way that `neg-http` is written - leaving the database in a non-valid state causes unintended errors.

1) To prevent the block-assembler from running, we introduce a new queue (`stopQueue`). This is a manual queue so does not automatically run queued jobs.  Before running the `block-assembler`, not only do we check that the `stateQueue` (i.e. queue[0]) is empty - but also that there is nothing queued in the `stopQueue`.

Elements are enqueued into `stopQueue` when a bad block is detected in the `blockProposeHandler`. Finally, they are dequeued at the completion of `rollbackHandler`.

2) To clean up the proposer, we modify `removeTransactionsFromMempool` such that it accepts the `transactionHashes` and the `blockNumberL2`.  When we assemble blocks locally we mark transactions as `mempool': false` and `blockNumberL2:-1`. We will update the `blockNumberL2` when we call the same function in `blockProposeHandler`.

Thus, we have a way to tell identify transactions that are included in blocks but have not been mined on the Ethereum mainnet. When the proposer is rotated and we were the last proposer, we reset these transactions as we know they will not find their way into the L2 Block chain.

### Tests
We skip the deep rollbacks in the `rollback-resync` tests as we can no longer create bad blocks that deeper than the chain tip at this point in time.

Finally, we also strip back neg-tests as the speed of ganache, coupled with our testing methods means that we need to minimise excessive transactions to ensure we have the correct, specific transaction types for a given test (e.g. Test 7 requires a `single_transfer`).

### DIFF
--- a/common-files/utils/event-queue.mjs
+++ b/common-files/utils/event-queue.mjs
@@ -26,7 +26,8 @@ import logger from 'common-files/utils/logger.mjs';
 const { MAX_QUEUE } = config;
 const fastQueue = new Queue({ autostart: true, concurrency: 1 });
 const slowQueue = new Queue({ autostart: true, concurrency: 1 });
-export const queues = [fastQueue, slowQueue];
+const stopQueue = new Queue({ autostart: false, concurrency: 1 });
+export const queues = [fastQueue, slowQueue, stopQueue];
 
 /**
 This function will return a promise that resolves to true when the next highest
@@ -68,6 +69,10 @@ async function enqueueEvent(callback, priority, args) {
   });
 }
 
+async function dequeueEvent(priority) {
+  queues[priority].shift();
+}
+
 async function queueManager(eventObject, eventArgs) {
   // First element of eventArgs must be the eventHandlers object
   const [eventHandlers, ...args] = eventArgs;
@@ -105,4 +110,4 @@ async function queueManager(eventObject, eventArgs) {
 }
 
 /* ignore unused exports */
-export { flushQueue, enqueueEvent, queueManager };
+export { flushQueue, enqueueEvent, dequeueEvent, queueManager };
--- a/nightfall-optimist/src/event-handlers/block-proposed.mjs
+++ b/nightfall-optimist/src/event-handlers/block-proposed.mjs
@@ -39,7 +39,7 @@ async function blockProposedEventHandler(data) {
     );
     // mark transactions so that they are out of the mempool,
     // so we don't try to use them in a block which we're proposing.
-    await removeTransactionsFromMemPool(block); // TODO is await needed?
+    await removeTransactionsFromMemPool(block.transactionHashes, block.blockNumberL2); // TODO is await needed?
 
     const latestTree = await getLatestTree();
     const blockCommitments = transactions.map(t => t.commitments.filter(c => c !== ZERO)).flat();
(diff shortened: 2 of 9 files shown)


daveroga (approved):

ChaitanyaKonda:
> We skip the deep rollbacks in the rollback-resync tests as we can no longer create bad blocks that deeper than the chain tip at this point in time.
 
A bad proposer's optimist can continue to create bad blocks if they make the changes, in which case we might need these tests. It makes good sense to add this directly to Adversary code in order to test deep rollbacks in rollback-resync and not Optimist code because good proposer should create no blocks on top of a bad block.

So, this test might not work after this PR but we should be able to have this working after adding adversary code? @IlyasRidhuan 

> Finally, we also strip back neg-tests as the speed of ganache, coupled with our testing methods means that we need to minimise excessive transactions to ensure we have the correct, specific transaction types for a given test (e.g. Test 7 requires a single_transfer).

I'm afraid I did not understand this.
Also, does changing code in proposer to stop proposing bad blocks or cleaning up proposer rotation break this neg-test ?

IlyasRidhuan:
> A bad proposer's optimist can continue to create bad blocks if they make the changes, in which case we might need these tests. It makes good sense to add this directly to Adversary code in order to test deep rollbacks in rollback-resync and not Optimist code because good proposer should create no blocks on top of a bad block.

`rollback-resync` kinda did both jobs for historical reasons - i believe that back in the day when we use to delete everything during a rollback, the only way to recover the `optimist` database was to run a re-sync (hence the name).

I think this test has to go into the wider test suite eventually because even though a good proposer doesn't create blocks on top of a bad one, it must still process blocks that may be created on top of a bad one by a bad proposer. So we will need to ensure that is can handle deeper rollbacks.

> So, this test might not work after this PR but we should be able to have this working after adding adversary code? @IlyasRidhuan

Yep, we will need to modify the adversary to ignore some of the changes in this PR. I've just skipped the tests for now.

> Finally, we also strip back neg-tests as the speed of ganache, coupled with our testing methods means that we need to minimise excessive transactions to ensure we have the correct, specific transaction types for a given test (e.g. Test 7 requires a single_transfer).
> I'm afraid I did not understand this. Also, does changing code in proposer to stop proposing bad blocks or cleaning up proposer rotation break this neg-test ?

Your last point is one of the reasons why i have had to remove some additional transactions. Basically we were doing more transactions than we needed to run `neg-http`. We could get away with it before because the `optimist` would eventually sort itself out - sometimes it would make multiple blocks within 1 challenge test but it would be resolved by a single rollback.

The new change means we only make 1 block per challenge test (i.e. only 1 block is made before a rollback is called whereas in the past 2 or more may be created). This seems to add an issue where it is not consistent as to what actual transactions we end up using to make a block. E.g. in Test 7 we need to ensure we have at least 1 transfer so we can replace a nullifier. The test seems to end up trying to make a block of only deposits which causes this check to fail. This is more an issue of **how we test** rather than with the PR itself.

ChaitanyaKonda (approved):

PR #326, opened 2021-11-23
Hari/groth16

dwebchapey:
fixes #289 

- Change Zokrates worker and env variables to `g16 , bellman`
- Modify vk structure to g16 format
- `Verifier.sol` and `Pairing.sol` files  from zokrates `export-verifier`

PR #332, opened 2021-11-26
Remove public input hash

### LINKED ISSUE #89: Remove public inputs hash
We currently hash public elements of a zksnark.  This is gas-efficient because the verification key length is minimised and the verification algorithm thus uses a minimum of gas.  However, the hash needs to be computed as part of the zksnark circuit and this is a major driver of constraint count and thus the time needed to compute a proof.  The problem is compounded by the use of a SHA256 hash to keep gas costs low (because in a non-optimistic solution the hash has to be recomputed on-chain for each transaction).

For NF_3, we only do an on-chain verification in the case of a challenge so we're less sensitive to Gas costs. Thus, we can remove the hashing of the public inputs and use the public input directly in the circuit. In this way we benefit from faster proof computation.

MirandaWood:
Closes #89 
Depends on #326 - I merged with this branch to build on G16, so makes sense for #326 to be merged to master first!

## Remove input hash

The main goal of this PR is to remove the costly public input hash from the circuits so we can generate proofs inside the browser. This requires a pretty big refactor everywhere since we now must deal with public inputs which differ between circuits. Main changes:

- Remove public input hash from circuits, replace with field element public inputs which are converted inside the circuit (not too costly constraint-wise)
- Remove public input hash from classes/client calls, replace with gathering public inputs for each verify call (on and off-chain)
- Since we must expand the proof challenge cases to cover 4 cases (one per circuit, instead of one for all circuits), I replaced the public input hash challenge with a larger proof challenge. Having both is possible but very heavy on the contract sizes. Since the contract gathers the public inputs from the tx, I believe the challenge covers the input hash failure cases.

### Testing

Use the usual tests: `npm test`. I've changed the stubs to work with the new parameters and they all run on my side. The main thing to check will be the **negative tests**, since these contain the challenges. I'm not as familiar with the code as anyone reading this so looking out for anything odd during a neg-http run would be appreciated!

## Outstanding changes

- Test on Ropsten
- Review use of truffle optimiser with runs = 1 - this could make some txs very costly gas wise, so if we can reduce the size of Challenges.sol in other ways that would be preferable
- Possibly remove invalid tx challenge and ensure single transfer/withdraw txns only contain one commitment/nullifier
- _Remove all references to public input hash in node (they exist but are never called)_ Done
- Possibly remove .publicInputs property of tx class - originally I passed these to the contract instead of the hash, but it became too costly (and it's basically sending the same info twice, just in a different form)
- Fully test proof challenge for transfer and withdraw (I think the neg test only covers a bad deposit proof)


ChaitanyaKonda:
> Remove all references to public input hash in node (they exist but are never called)
> Possibly remove .publicInputs property of tx class - originally I passed these to the contract instead of the hash, but it became too costly (and it's basically sending the same info twice, just in a different form)

The definition of done for this issue could be considered to remove all artifacts about public input hash including the above. 
During resolution of merge conflicts, can we have this resolved please?

After this PR is reviewed and merged, I will create issue for
> Review use of truffle optimiser with runs = 1 - this could make some txs very costly gas wise, so if we can reduce the size of Challenges.sol in other ways that would be preferable

MirandaWood:
> The definition of done for this issue could be considered to remove all artifacts about public input hash including the above. During resolution of merge conflicts, can we have this resolved please?

Done, ready to review once tests have passed

MirandaWood:
Could the final reviewer please squash and merge?
Apologies for the messy commits! I fixed conflicts before doing a git pull locally, so had to re-do them, causing history issues.

ChaitanyaKonda (approved):

IlyasRidhuan (approved):

PR #333, opened 2021-11-26
Westlad/ping pong

### LINKED ISSUE #302: Deploy to servers for ping pong test
In Ireland, deploy
- proposer app which uses proposer SDK
- challenger app which uses challenger SDK
- client1 app which uses client SDK
- optimist instance (which includes zokrates binary. Issue needs #299)
- proposer and challenger apps will talk to this optimist instance
- client instance
- client1 app will talk to this client instance 
- zokrates worker instance (because the test is in server and not browser)

In Ohio, deploy
- bad proposer app which proposer SDK
- optimist instance (which includes zokrates binary Issue needs #299)
- bad proposer will talk to this optimist instance
- client2 app which uses client SDK
- client instance
- client2 app will talk to this client instance 

Run each of these using `screen` terminal multiplexer so that we can see them run even if we log out of the server and log in later.
Running as background daemon could be an option. But displaying output is easier for the former

Ensure this deployment talks to testnet deployment. Use the same circuit keys and contract abis.

Reach out to @ChaitanyaKonda for these files. Pending the push to S3. @druiz0992 will create a shared volume for ireland and ohio with these files

### LINKED ISSUE #303: Create ping pong test
Depends on #302 

Create a simple test with 1 Block Proposer, 1 Challenger, 2 Clients using the instances deployed in Ireland and Ohio.
The test should run for a few days with valid transactions submitted and valid blocks created.
Should include deposit, transfer, withdraw transactions.

The test will include a node script that will periodically confirm the balance that the script thinks the user should have against the balance stored in the client's database

Westlad:
This PR does two distinct things:
1) Adds a set of scripts to run the Ping-Pong test on two servers. 
2) Adds a github action to build images for the tests.  This is so that the tests can be run on a server with docker images, rather than being dependent on the local file system. Thus it's a more 'production-like' test.

Note that this PR fixes #302, fixes #303.

The code can be tested on a Mac by following the instructions in the [readme](https://github.com/EYBlockchain/nightfall_3/blob/westlad/ping-pong/test/ping-pong/README.md).
It should also run on a linux machine but this is untested.


Westlad:
Can someone with a working wallet check that this doesn't break it please? The wallet compiles ok but I can't fully test it.

ChaitanyaKonda (approved):

druiz0992 (approved):
As Linux user, procedure is as you point out in the instructions. Add external IP to host.docker.internal and and run tests as described.
One nuisance I found is that I cant CRTL+C ganache or pong-app, and I need to kill the terminal

IlyasRidhuan (approved):

PR #342, opened 2021-12-01
David/wallet error msg

### LINKED ISSUE #221: Wallet - Display nightfall network error messages
Currently, there is a console.log for the errors that may happen after a transaction.

druiz0992:
Show error message in wallet when wallet expects to connect to localhost and metamask is connected to a different network, It also shows error message when a transaction fails.

This PR also loads erc contract addresses from contract name and not from an expected address. 

NOTE: You must clear local storage

To test it:
- connect metamask to Ropsten network, start the wallet, and try to log in. An error message should appear
- Send a deposit without having reset the metamask nonce => An error message should appear
- Send a transfer/withdraw without any valid commitment => An error message should appear

Fixes #221


ChaitanyaKonda:
@druiz0992 What would be the best way to test this please?

druiz0992:
Added summary on how to test it in PR description

daveroga:
It seems now the address of the ERC20 token is different and throws an error just entering into the wallet with Metamask.

ChaitanyaKonda:
It says `incorrect network` even when I'm connected to `http://localhost:8546`

ChaitanyaKonda:
It says `incorrect network` even when I'm connected to `http://localhost:8546`

Screenshot 2021-12-14 at 15 01 06

druiz0992:
@ChaitanyaKonda , it’s not supposed to recognize the fact that you switched metamask from an unexpected network to ganache. It just alerts you that you attempted to login into nightfall using an unexpected network. So, the message is just a warning from a previous attempt to connect, not the fact that you switched to ganache

druiz0992:
@daveroga , thanks for pointing this out. I fixed your issue. You will have to clear local storage as well.

druiz0992:
@ChaitanyaKonda, I am not able to reproduce this error. If you attempt to login using Ropsten, it will show the error message. Then, you switch to ganache and reattempt to login, and it will succeed. Can you please retry?

daveroga (approved):

ChaitanyaKonda:
I started the wallet using `npm run start:ropsten` and I tried to connect it to `http://localhost:8546`. So it seems to be correct in pointing `wrong network`. The last comment was was my mistake. But at the same time, ropsten should be the correct network in this case. But it seems to show wrong network if I pick ropsten network too.

Wallet instructions in `README.md` in the root dir suggests running wallet with ropsten using `npm run start-ropsten` but the correct instruction seems to be `npm run start:ropsten`. Could we update this please? 

We are specifying the error messages as follows
- Incorrect network
- Transaction Failed
- Withdraw Failed
Could we provide the reason for the errors too?

druiz0992:
@ChaitanyaKonda, i fixed README. I give a bit more of info when network is incorrect. By i can't read all error types (I can't read metamask errors). I will look at this, but I believe for now it should be ok as we are alerting of errors in transactions (whereas before there wasn't any alert when a transaction failed).  We need the fixes from this PR. I will raise an issue to be more specific with errors.

Westlad:
The error message shows correctly if I try to use Ropsten.  However when I log in normally to the correct network by signing the Metamask transaction, the wallet screen opens correctly for a few seconds and then an error throws:
image

druiz0992:
@Westlad, yes, we saw that.  Have you cleared local storage? Right now it is attempting to load a token at an address that does not exist. This PR fixes that, but you need to clear local storage as well (please, see PR description).

ChaitanyaKonda:
Tested and works well.

ChaitanyaKonda (approved):

PR #349, opened 2021-12-06
Add Zokrates binary to Optimist

### LINKED ISSUE #299: Add zokrates binary into nightfall optimist
- Use the zokrates binary directly within nightfall optimist to verify proof during `checkTransaction`
- Remove the dependency for communicating with Zokrates Worker service from Nightfall Optimist

MirandaWood:
Closes #299 

This small PR removes the dependency on zokrates-worker in optimist. It replaces the off-chain verify call in `transaction-checker` with a 'local' version of verify using the binary. This appears to be the only place zokrates was directly called.

The usual `npm test`s can be run, with a focus on the transaction checker logging. On my side, it correctly passes a verifying proof and fails an incorrect proof for the challenge.


nightfall-optimist/src/services/verify.mjs
```diff
@@ -0,0 +1,85 @@
...
+ * @param {Object} [options] - Options for output
+ * @param {Boolean} options.createFile - Whether or not to output a json file
+ * @param {String} [options.directory=./] - Directory to output files in
+ * @param {String} [options.fileName=proof.json] - Name of JSON proof file ()
+ */
+export default async function verify({
+  vk,
+  proof,
+  inputs,
+  provingScheme = 'gm17',
```
dwebchapey, line 28: Switch to G16 and bellman to align with changes in #326 ?

ChaitanyaKonda (approved):

nightfall-optimist/src/services/verify.mjs
MirandaWood: Done, thanks for the reminder!

ChaitanyaKonda (approved):

Westlad (approved):

PR #356, opened 2021-12-07
Hari/duplicate commitment

### LINKED ISSUE #336: During double transfer, the same commitment is being picked twice
During `double transfer`, the same commitment is being picked twice by `findUseableCommitments`. 
To replicate this error, create a block with two ERC20 deposits of value 1000 each and then create a block with two ERC20 transfers of value 5 each. You will notice `findUseableCommitments` will pick the same commitment to spend twice.

dwebchapey:
This closes #336 .
The issue was because all available commitments are greater than `target` value and hence `commitsLessThanTargetValue.length` was zero . 
In this scenario , `sortedCommits[commitsLessThanTargetValue.length]` and `sortedCommits [0]` return same commitment , the one with index 0.

This is fixed by returning first two sortedCommits , ie; `sortedCommits [0]` and `sortedCommits [1]` values if `commitsLessThanTargetValue.length` is zero.

> To recreate the original issue , run `npm run duplicate-test`  by commenting out line #423 in `nightfall-client/src/services/commitment-storage.mjs` 
> `if (commitsLessThanTargetValue.length === 0) return [sortedCommits[0], sortedCommits[1]];` and note that the same commitment is picked both times for double transfer. Attaching a screenshot for reference 
> 
> Screenshot 2021-12-14 at 4 25 58 PM

> To review the PR with fix , uncomment the line #423 and run `npm run duplicate-test`. The commitments chosen will be different as you can see in attached screenshot reference.
> 
> Screenshot 2021-12-14 at 4 32 19 PM

An extra check  `commitments.length < 2 ` is done for double transfer to exit if 2 commitments are not available.

Additionally , this PR also fixes duplicateNullifier check in optimist in `transaction-checker.mjs` . Only `double transfer (case 2 )` need the check since other 3 cases fail if `nullifier[1]`  is non zero.


ChaitanyaKonda:
Has your fix been tested by replicating this situation as described in the issue description?

nightfall-optimist/src/event-handlers/transaction-submitted.mjs
```diff
@@ -63,8 +63,9 @@ async function transactionSubmittedEventHandler(eventParams) {
     const transactionNullifiers = transaction.nullifiers.filter(
       hash => hash !== '0x0000000000000000000000000000000000000000000000000000000000000000',
     ); // Deposit transactions still have nullifier fields but they are 0
+    const uniqueNullifiers = Array.from(new Set(transactionNullifiers));
```
ChaitanyaKonda, line 66: Duplicate nullifiers check has to happen inside `checkTransactionType` of `checkTransaction` call of the `transaction-submitted` event handler. 

When writing the PR description, please specify the various fixes done.

nightfall-optimist/src/services/check-block.mjs
```diff
@@ -80,10 +80,17 @@ async function checkBlock(block, transactions) {
   // to check a block against itself (hence the second filter).
   const storedMinedNullifiers = await retrieveMinedNullifiers(); // List of Nullifiers stored by blockProposer
   const blockNullifiers = transactions.map(tNull => tNull.nullifiers).flat(Infinity); // List of Nullifiers in block
+  const transactionNullifiers = blockNullifiers.filter(
```
ChaitanyaKonda, line 83: Because we re checking `assert(nullifier[0].hash != nullifier[1].hash)` inside `double_transfer.zok`, if same commitment is spent twice in a double transfer, the proof verification will. This will be caught by `checkBlock` under `verifyProof` of `checkTransaction`. So this is redundant fix here

ChaitanyaKonda (changes requested):

dwebchapey:
Successfully tested by recreating the scenario mentioned in issue #336

ChaitanyaKonda:
Could you provide instructions for a reviewer to test it please?

dwebchapey:
To test , please do `npm run duplicate-test`
This test create four ERC20 deposits of value 1000 each and then create a block with two ERC20 transfers of value 5 each.
The commitments picked are different.

The issue #336 , `To replicate this error, create a block with two ERC20 deposits of value 1000 each and then create a block with two ERC20 transfers of value 5 each.` .  If only 2 deposits of value 1000 are committed , client can do only one transfer in same block because after the first double transfer `commitments`  that are on-chain are zero.

Screenshot 2021-12-14 at 2 52 02 PM

ChaitanyaKonda (approved):

ChaitanyaKonda:
Considering we can't do a gist for test file as suggested above, let us test it with the duplicate test as you have suggested already, then please submit a new PR after this merge such that this new PR removes this test file.

MirandaWood (approved):
Tested with duplicate test and all working!

dwebchapey:
Pushed new PR after removing issue specific test file.

Issue #358, opened 2021-12-10
Liquidity provider approve withdraw


druiz0992:
Currently, liquidity provider does this in `cli/src/liquidity-prodiver.mjs`:
```
await approve(
    erc20Address,
    nf3.ethereumAddress,
    nf3.shieldContractAddress,
    'ERC20',
    '1000000000000',
    nf3.web3,
  );
```
where contract address, erc type and amount are constants. This information should be retrieved from the withdrawal transaction hash and not be a constant.

daveroga:
Pushed a branch `daveroga/instant-withdraw-approve-sdk` that solves it in sdk.

PR #369, opened 2021-12-10
David/liquidity provider

### LINKED ISSUE #341: Substitute xy-pkd for compressed pkd in wallet deposit and transfer
Currently, to do a deposit or a transfer from the wallet, user needs to enter x and y components of the pkd as the destination address. It would be much simpler for the user to enter a compressed pkd have the wallet do the conversion to x-y format.

druiz0992:
Includes several improvements:
- Add approve function to liquidity provider 
- Substitute xy pkd by compressed pkd when doing transactions
- Set ERC721 supply to 1M
- Added pending deposit and pending spent balances info to wallet and cli

Closes #341 #358

### DIFF
--- a/nightfall-client/src/routes/finalise-withdrawal.mjs
+++ b/nightfall-client/src/routes/finalise-withdrawal.mjs
@@ -12,11 +12,11 @@ const router = express.Router();
 router.post('/', async (req, res, next) => {
   logger.debug(`finalise-withdrawal endpoint received POST ${JSON.stringify(req.body, null, 2)}`);
   try {
-    const { rawTransaction: txDataToSign } = await finaliseWithdrawal(req.body);
+    const { rawTransaction: txDataToSign, transaction } = await finaliseWithdrawal(req.body);
     logger.debug('returning raw transaction');
     logger.silly(`raw transaction is ${JSON.stringify(txDataToSign, null, 2)}`);
     // convert commitment from GN to hex form for transmission
-    res.json({ txDataToSign });
+    res.json({ txDataToSign, transaction });
   } catch (err) {
     logger.error(err);
     next(err);
--- a/cli/lib/nf3.mjs
+++ b/cli/lib/nf3.mjs
@@ -289,8 +289,7 @@ class Nf3 {
   @param {string} tokenId - The ID of an ERC721 or ERC1155 token.  In the case of
   an 'ERC20' coin, this should be set to '0x00'.
   @param {object} keys - The ZKP private key set of the sender.
-  @param {array} pkd - The transmission key of the recipient (this is a curve point
-  represented as an array of two hex strings).
+  @param {string} compressedPkd - The compressed transmission key of the recipient
   @returns {Promise} Resolves into the Ethereum transaction receipt.
   */
   async transfer(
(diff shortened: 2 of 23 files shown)


daveroga (approved):

ChaitanyaKonda:
Could you provide instructions for a reviewer to test it please?

ChaitanyaKonda:
`requestInstantWithdrawal` will be called by a user who would like to withdraw instantly. They will call this function to withdraw and set a fee they are willing to pay.

LP will be listening to instant withdrawal events using `getInstantWithdrawalRequestedEmitter` provided by SDK. On receipt of an instant withdrawal request, they will provide liquidity by calling `advanceInstantWithdrawal` in SDK.

Providing shield contract approval to transfer tokens on behalf of LP, should be done separely. Any LP, when they register as LP, they will do a bulk approval at that instant. This approval limit can be topped up frequently as required.

Considering the above design, we should remove `approve` inside `advanceInstantWithdrawal` and add this to `cli/src/liquidity-provider.mjs` before calling `getInstantWithdrawalRequestedEmitter`.

Looking at this file, it seems to be done there already. So we can remove `approve` inside `advanceInstantWithdrawal`.

Also, what is the second transaction a user is signing in wallet when a request for instant withdrawal is made?

nightfall-client/src/routes/finalise-withdrawal.mjs
```diff
@@ -12,11 +12,11 @@ const router = express.Router();
 router.post('/', async (req, res, next) => {
   logger.debug(`finalise-withdrawal endpoint received POST ${JSON.stringify(req.body, null, 2)}`);
   try {
-    const { rawTransaction: txDataToSign } = await finaliseWithdrawal(req.body);
+    const { rawTransaction: txDataToSign, transaction } = await finaliseWithdrawal(req.body);
```
ChaitanyaKonda, line 15: We don't seem to be using the `transaction` that was added here. 
Could we remove this please?

druiz0992: The second transaction that its signed when requesting an instant withdrawal is the request of an instant withdraw. 
```  /**
  Enables someone with a valid withdraw transaction in flight to request instant
  withdrawal of funds (only relevant for ERC20).
  @method
  @async
  @param {string} withdrawTransactionHash - the hash of the Layer 2 transaction in question
  @param {number} fee - the amount being paid for the instant withdrawal service
  */
  async requestInstantWithdrawal(withdrawTransactionHash, fee) {
    try {
      // set the instant withdrawal fee
      const res = await axios.post(`${this.clientBaseUrl}/set-instant-withdrawal`, {
        transactionHash: withdrawTransactionHash,
      });
      return this.submitTransaction(res.data.txDataToSign, this.shieldContractAddress, fee);
    } catch {
      return null;
    }
  }```

ChaitanyaKonda: I thought this was the first and only transaction the user signs

druiz0992: The first transaction the user signs is the withdraw request. The second is the instant withdraw request

ChaitanyaKonda: That makes sense. Thank you

daveroga (approved):

ChaitanyaKonda (approved):

PR #380, opened 2021-12-16
Use Transaction Hash when deciding if client should save blocks

### LINKED ISSUE #360: Instant withdraw doesn't work when a transfer to a different account and a withdraw are part of the same L2 block
Instant withdraw doesn't work when a transfer to a different account (not one's account) and a withdraw are part of the same L2 block. nighfall client cannot find withdraw hash. However, withdraw is executed (funds are not in ones account anymore).

IlyasRidhuan:
This PR closes #360.

Previously, we used commitments to decide if we should store block and transaction information of transactions we care about in `client`. However, when a withdrawal transaction is made - it does not have commitments. Therefore, we do not store it when we should. 

This PR changes this behaviour so that we use `transactionHash` to decide if we should store a block and its transactions.

This can be tested in the following way (you may need to rebuild containers and purge volumes if you are from master): 

1. Fetch `daveroga/test-bug-instant-withdraw` and `./start-nightfall` - stubs circuits can be used.
2. Run `npm run test-e2e`, which should fail at `should allow instant withdraw of existing withdraw`.
3. Cherry-pick this commit onto the branch using `git cherry-pick <this commit sha>`
4. `npm run test-e2e` should now pass.

### DIFF
--- a/nightfall-client/src/event-handlers/block-proposed.mjs
+++ b/nightfall-client/src/event-handlers/block-proposed.mjs
@@ -7,6 +7,7 @@ import {
   storeCommitment,
   countCommitments,
   setSiblingInfo,
+  countTransactionHashes,
 } from '../services/commitment-storage.mjs';
 import getProposeBlockCalldata from '../services/process-calldata.mjs';
 import Secrets from '../classes/secrets.mjs';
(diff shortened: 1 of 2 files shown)


ChaitanyaKonda (approved):

ChaitanyaKonda:
Test passed

daveroga (approved):

PR #382, opened 2021-12-17
Hari/block zero challenges

### LINKED ISSUE #340: Block challenges for the first block
This applies to Invalid Root and Invalid Leaf Count challenges.

When creating these challenge against block n, we need to retrieve the transactions from block n-1 or the frontier from block n-2.

If n == 0 (the first block), the challenge throws an error as the transactions for `blockNumberL2 === -1` is understandably not found.

The challenge should be modified such that we can still challenge this block.

dwebchapey:
This Fixes #340 .

A block object` IDENTITY_BLOCK` is created with all zero fields in `challenges.mjs`. This is assigned to `priorBlock` and/or `priorPriorBlock` when the blockNumber of block to be challenged is 0 ( also for 1 in case of priorPriorBlock ).

In `challengeNewRootCorrect` and `challengeLeafCountCorrect` of `Challenges.sol`, `isBlockReal` of previous block is not performed if priorBlock is an `IDENTITY_BLOCK`.  It is also mandatory for the condition `blockL2.blockNumberL2 == priorBlockL2.blockNumberL2+1`  to be true on all situations expect when `blockL2.blockNumberL2 === 0` where the condition `priorBlockL2.blockNumberL2 ===0 `should also hold true .

Break the iterative state pop in `removeHashes` `blockNumberL2 == 0`

The `libChallengeNewRootCorrect` inside the contract `challengesUtil.sol` also perform `MerkleTree_Stateless.checkPath` validity check only if `blockNumberL2 != 0`.

**Test file description**
A test file `test/block-zero-challenge.mjs` is created to perform incorrect root and incorrect leaf count check at block zero. A bad block is generated with two deposit transactions. The incorrect root check and leaf count are executed in a similar way to `neg-http.mjs` . 

**Guidelines to review**

- To test , run `npm run block-zero-test` . 
- To review the fix, please check the logs and notice challenges and subsequent rollbacks occurring at block zero.  
- Inorder to check whether challenge for block zero is getting accepted with a faulty created priorBlock(not `IDENTITY_BLOCK`) , modify the values in `IDENTITY_BLOCK` in `challenges.mjs`. 
For eg:  Change the value of `leafCount` and/or `blockNumberL2` to a non zero value


MirandaWood (approved):

package.json
```diff
@@ -13,7 +13,8 @@
     "test-ercmocks": "env-cmd -e development mocha --timeout 0 --bail test/e2e/ercmocks.test.mjs",
     "lint": "eslint . --ext js,mjs,jsx && find-unused-exports",
     "prepare": "husky install",
-    "doc:build:sdk": "jsdoc -c jsdoc.json cli/lib/nf3.mjs"
+    "doc:build:sdk": "jsdoc -c jsdoc.json cli/lib/nf3.mjs",
+    "block-zero-test": "env-cmd -e development mocha --timeout 0 --bail test/block-zero-challenge.mjs"
```
IlyasRidhuan, line 17: It's great there is a test case for this - but I'm not sure if we want to be adding these specific kinds of test to the repo. Providing instructions to apply to existing test (if they're simple enough) should be sufficient.

E.g. for this one, would it not be sufficient to comment out parts of `neg-http` to recreate `block-zero-challenge`

nightfall-deployer/contracts/Challenges.sol
```diff
@@ -35,7 +35,9 @@ contract Challenges is Stateful, Key_Registry, Config {
   ) external {
     checkCommit(msg.data);
     // check if the block hash is correct and the block hash exists for the block and prior block
+    if (blockL2.blockNumberL2 != 0) {
```
IlyasRidhuan, line 38: Do we need to check if a block that is passed in as `blockNumberL2 ==0` is indeed an identity block?

package.json
dwebchapey: Yes . That should be sufficient . But @ChaitanyaKonda suggested to add a separate test file to make it easier for reviewers who are reviewing without context. Once the issue receive enough approvals , I will push a new PR without the issue specific test file.

nightfall-deployer/contracts/Challenges.sol
dwebchapey: Do you mean check whether it is indeed genesis block in L2? A block which passes `blockNumberL2 ==0` will be genesis block.  The identity block value is assigned to `priorBlock` and/or `priorPriorBlock` when  `blockNumberL2 ==0` .

IlyasRidhuan: > A block which passes blockNumberL2 ==0 will be genesis block.

This is not necessarily true, a bad actor can submit a block as part of the input to this function where the `blockNumberL2 == 0` but the `leafCount > 0` which will incorrectly bypass this check right?

ChaitanyaKonda: @dwebchapey

dwebchapey: Added checks to confirm if priorBlock is `IDENTITY_BLOCK`

ChaitanyaKonda: @IlyasRidhuan Has this been resolved?

package.json
ChaitanyaKonda: @IlyasRidhuan is right. My earlier suggestion was for a bug fix that we weren't going to replicate with an easy change to the test. If we can do a quick change to test file by adding something as simple as `.skip` to some tests, then we should do it. This will eliminate the additional step of a temporary test file and having to remove it. 

Another suggestion was to have a test file as `gist` that reviewers could copy + paste and test with, but you can't push to gist.  Could you check that again please?

nightfall-deployer/contracts/Challenges.sol
```diff
@@ -34,8 +34,11 @@ contract Challenges is Stateful, Key_Registry, Config {
     bytes32 salt
   ) external {
     checkCommit(msg.data);
+    require(blockL2.blockNumberL2 == priorBlockL2.blockNumberL2+1 || (blockL2.blockNumberL2 == 0 && priorBlockL2.blockNumberL2 == 0),'previous block number is incorrect');
```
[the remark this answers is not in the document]
dwebchapey: Yes . It is redundant but my thought was to exit if the first check itself fail. I will do a check for the block zero case separately. Would it be better to write the checks inside `ChallengesUtil.sol `and `Utils.sol` instead of `Challenges.sol` ?

dwebchapey:
Bootstrapping Nightfall by making sure first few blocks are good blocks which means that block zero challenges will not be made. Closing this PR since earlier proposed fix is not required.

Issue #385, opened 2021-12-17
Review current restrictions for ERC721 and ERC1155 about tokenId and value


daveroga:
Some restrictions in the Shield contract for ERC721 and ERC1155 contracts should be changed and could cause a breaking change in some parts of the code.
- tokenId should be different from 0 in the contract now for ERC721 and ERC1155, but it could be 0 in both ERC721 and ERC1155 according to EIPs https://eips.ethereum.org/EIPS/eip-721.
- value always should be 0 in ERC721 in the contract now. The restriction could be always to be 1 and then it will be easier to get balances from commitments as we don't have to check for the tokenType.
- value in ERC1155 should be different from 0 in the contract now. But in ERC20 there is no restriction to be different from 0 in the contract. It should be the same restrictions for ERC20 and ERC1155 in terms of value as an ERC1155 can contain different ERC20.

Review these restrictions in order to cover all the ERC721 and ERC1155 cases.

Issue #384, opened 2021-12-17
Incorrect balances in ERC1155 in SDK


daveroga:
When using function `getLayer2BalancesDetails` in the SDK, ERC20 and ERC721 balances are ok but ERC1155 are wrong.

Issue #381, opened 2021-12-16
Tx nonce error in combining ERC20 and ERC721 deposits in SDK


daveroga:
When testing different ERC20 and ERC721 deposits the error `Error: Returned error: the tx doesn't have the correct nonce. account has nonce of: 235 tx has nonce of: 234` is thrown.

You can reproduce the issue by:
- start ganache with `./start-nightfall -g -s` in one terminal.
- start a proposer with `./proposer` in another terminal.
- start a liquidity provider with `./liquidity-provider` in another terminal.
- start the nightfall cli with `.\nf`
- Make 1 ERC20 deposit
- Make 1 ERC721 deposit (tokenId: 1)

daveroga:
It seems the issue is in `deposit` function in the nf3 SDK.
When doing the deposit the first time for an ERC contract is calling an `approve` function that is sending a transaction directly without signing it. This is because returning the `encodeABI` of the transaction to be sign in the `deposit` function is conditioned to `process.env.USER_ETHEREUM_SIGNING_KEY` I don't have.

I've updated `approve` function to receive an parameter `encodeABI` in order to return the encodeABI or sending directly the transaction without signing it. In `deposit` function in the SDK I sent this parameter based on `this.ethereumSigningKey` of the nf3 SDK.

In tokens.mjs
```
async function approve(
  ercAddress,
  ownerAddress,
  spenderAddress,
  tokenType,
  value,
  provider,
  encodeABI,
) {
  const abi = getAbi(tokenType);
  const ercContract = new provider.eth.Contract(abi, ercAddress);

  switch (tokenType) {
    case TOKEN_TYPE.ERC20: {
      const allowance = await ercContract.methods.allowance(ownerAddress, spenderAddress).call();
      const allowanceBN = new Web3.utils.BN(allowance);
      const valueBN = new Web3.utils.BN(value);
      if (allowanceBN.lt(valueBN)) {
        if (process.env.USER_ETHEREUM_SIGNING_KEY || encodeABI)
          return ercContract.methods.approve(spenderAddress, APPROVE_AMOUNT).encodeABI();
        await ercContract.methods
          .approve(spenderAddress, APPROVE_AMOUNT)
          .send({ from: ownerAddress });
      }
      return Promise.resolve(false);
    }

    case TOKEN_TYPE.ERC721:
    case TOKEN_TYPE.ERC1155: {
[...]

Issue #379, opened 2021-12-15
Add ERC721 and ERC1155 tokens SDK tests to e2e test


daveroga:
Now the e2e tests only covers tests for ERC20 tokens. We need to add SDK tests for ERC721 and ERC1155 tokens.

PR #386, opened 2021-12-20
David/wallet erc1155

### LINKED ISSUE #351: Working ERC1155 functionality in wallet
Currently ERC1155 is shown in the wallet as a regular token. However, this doesnt make much sense because ERC1155 contains several tokens inside and showing an aggregated balance is meaningless. Adapt the wallet to show balance per token within ERC1155.
cli should also be adapted to manage erc1155

druiz0992:
Includes functionality to work with ERC1155 tokens in the wallet, as well as adding tests for ERC721 and ERC1155 to sdk testing.
- In order to have ERC1155 working correctly, we have removed restrictions to Shield contract about `tokenId` for ERC1155 and ERC721. Now, `tokenId` 0 is allowed. 
- Calculation of balances in `commitments-storage.mjs` from commitment value and `tokenId` now for an ercAddress. If it's an ERC20 tokenId will be 0. (Ex.: [ { balance: 60, tokenId: 0 } ])
- Fix nonce error when changing deposit from ERC20 to ERC721. Approve transaction with missing encodeABI was the cause.

From wallet now you can
- Visualize ERC1155 tokens
- If you select an ERC1155 token, details will be displayed
- You can add new ERC1155 tokens (new ERC1155 address), or add a new token ID to an existing ERC1155 token

To test SDK, `npm run test-e2e`
To test functionality, launch wallet tests by `./start-nightfall -wt -s` and search for `Selenium tests PASSED` in `wallet-test` container

Closes #351, #385 , #384 , #381 , #379, #365


daveroga (approved):

cli/lib/nf3.mjs
```diff
@@ -786,12 +810,13 @@ class Nf3 {
...
   Set a Web3 Provider URL
   */
-  setWeb3Provider() {
+  async setWeb3Provider() {
     this.web3 = new Web3(this.web3WsUrl, { transactionBlockTimeout: 200 }); // set a longer timeout
     if (typeof window !== 'undefined') {
       if (window.ethereum && this.ethereumSigningKey === '') {
         this.web3 = new Web3(window.ethereum);
-        window.ethereum.request({ method: 'eth_accounts' });
+        await window.ethereum.enable();
```
ChaitanyaKonda, line 818: Deprecated function. Needs to be replaced with `ethereum.request({ method: 'eth_requestAccounts' })`

cli/lib/tokens.mjs
```diff
@@ -13,7 +13,15 @@ Sends an approve transaction to an ERC20/ERC721/ERC1155 contract for a certain a
...
 */
-async function approve(ercAddress, ownerAddress, spenderAddress, tokenType, value, provider) {
+async function approve(
+  ercAddress,
+  ownerAddress,
+  spenderAddress,
+  tokenType,
+  value,
+  provider,
+  encodeABI,
```
ChaitanyaKonda, line 23: This must be `ethereumSigningKey` not `encodeABI`

doc/lib/nf3.mjs.html
```diff
@@ -727,7 +727,7 @@ <h1 class="page-title">Source: nf3.mjs</h1>
   value of each propery is the number of tokens originating from that contract.
   */
   async getLayer2Balances() {
-    const res = await axios.get(`${this.clientBaseUrl}/commitment/balance`);
+    const res = await axios.post(`${this.clientBaseUrl}/commitment/balance`);
```
ChaitanyaKonda, line 730: We are requesting data through this API call. It should be `post`

cli/lib/nf3.mjs
```diff
@@ -734,26 +744,40 @@ class Nf3 {
...
   from that contract.
   */
-  async getLayer2PendingSpentBalances() {
-    const res = await axios.get(`${this.clientBaseUrl}/commitment/pending-spent`);
+  async getLayer2PendingSpentBalances(ercList, filterByCompressedPkd) {
+    const params = {
+      compressedPkd: filterByCompressedPkd === true ? this.zkpKeys.compressedPkd : null,
+      ercList,
+    };
+    const res = await axios.post(`${this.clientBaseUrl}/commitment/pending-spent`, params);
```
ChaitanyaKonda, line 780: This must be `GET` not `POST`

cli/lib/nf3.mjs
```diff
@@ -734,26 +744,40 @@ class Nf3 {
...
   value of each propery is the number of tokens pending deposit from that contract.
   */
-  async getLayer2PendingDepositBalances() {
-    const res = await axios.get(`${this.clientBaseUrl}/commitment/pending-deposit`);
+  async getLayer2PendingDepositBalances(ercList, filterByCompressedPkd) {
+    const params = {
+      compressedPkd: filterByCompressedPkd === true ? this.zkpKeys.compressedPkd : null,
+      ercList,
+    };
+    const res = await axios.post(`${this.clientBaseUrl}/commitment/pending-deposit`, params);
```
ChaitanyaKonda, line 759: This must be `GET` not `POST`

cli/lib/nf3.mjs
```diff
@@ -704,12 +707,19 @@ class Nf3 {
...
   value of each propery is the number of tokens originating from that contract.
   */
-  async getLayer2Balances() {
-    const res = await axios.get(`${this.clientBaseUrl}/commitment/balance`);
+  async getLayer2Balances(ercList, filterByCompressedPkd) {
+    const params = {
+      compressedPkd: filterByCompressedPkd === true ? this.zkpKeys.compressedPkd : null,
+      ercList,
+    };
+    const res = await axios.post(`${this.clientBaseUrl}/commitment/balance`, params);
```
ChaitanyaKonda, line 722: This must be `GET` not `POST`

nightfall-client/src/routes/commitment.mjs
```diff
@@ -30,10 +30,12 @@ router.get('/salt', async (req, res, next) => {
   }
 });
 
-router.get('/balance', async (req, res, next) => {
-  logger.debug('commitment/balance endpoint received GET');
+router.post('/balance', async (req, res, next) => {
```
ChaitanyaKonda, line 33: Please revert all these `POST` changes. `GET` is correct. I'm failing to see how this is a POST when we are requesting from server on the API level and on database level querying the database too

nightfall-deployer/contracts/ERC1155Mock.sol
```diff
@@ -11,10 +11,10 @@ contract ERC1155Mock is ERC1155 {
...
 
     constructor() ERC1155("https://game.example/api/item/{id}.json") {
-        _mint(msg.sender, GOLD, 10**18, "");
-        _mint(msg.sender, SILVER, 10**27, "");
-        _mint(msg.sender, THORS_HAMMER, 1, "");
-        _mint(msg.sender, SWORD, 10**9, "");
-        _mint(msg.sender, SHIELD, 10**9, "");
+        _mint(msg.sender, GOLD, 1100000, "");
+        _mint(msg.sender, SILVER, 1200000, "");
+        _mint(msg.sender, THORS_HAMMER, 10, "");
```
ChaitanyaKonda, line 16: Hahaha! That's a good one

nightfall-deployer/contracts/Shield.sol
```diff
@@ -185,7 +185,7 @@ contract Shield is Stateful, Structures, Config, Key_Registry {
           uint256(t.value)
         );
     } else if(t.tokenType == TokenType.ERC721) {
-      if (t.value != 0 || t.tokenId == ZERO) // value should always be equal to 0 and tokenId cannot be equal to ZERO
```
ChaitanyaKonda, line 188: Without this change, do the tests fail somewhere?

test/http.mjs
```diff
@@ -764,7 +764,7 @@ describe('Testing the http API', () => {
 
   describe('Check commitment wallet balance', () => {
     it('Should return the wallet balance', async function () {
-      const res = await chai.request(senderUrl).get('/commitment/balance');
+      const res = await chai.request(senderUrl).post('/commitment/balance');
```
ChaitanyaKonda, line 767: GET

wallet/README.md
```diff
@@ -1,14 +1,7 @@
...
 ## Testing ERC Contracts
-For testing purposes, account `0x9C8B2276D490141Ae1440Da660E470E7C0349C63` with privateKey `0x4775af73d6dc84a0ae76f8726bda4b9ecf187c377229cb39e1afa7a18236a69e` has been setup in testnet wit some 
-Ether and some ERC tokens.
-The following is the list of test ERC contracts deployed:
-- 0xb5acbe9a0f1f8b98f3fc04471f7fe5d2c222cb44 => ERC20 contract
-- 0x103ac4b398bca487df8b27fd484549e33c234b0d => ERC721 contract. Existing token Ids are 1,2, and 3
-- 0x9635c600697587dd8e603120ed0e76cc3a9efe4c => ERC1155 contract. Existing token Ids are 1 through 4.
-
-When starting the wallet, you can add the address of these tokens if they do not appear by default.
+For testing purposes, account `0x9C8B2276D490141Ae1440Da660E470E7C0349C63` with privateKey `0x4775af73d6dc84a0ae76f8726bda4b9ecf187c377229cb39e1afa7a18236a69e` has been setup in testnet wit some Ether and some ERC tokens.
```
ChaitanyaKonda, line 4: We should never have private keys in our repo even if we were to use it in ropsten.
Bots can steal ropsten ether, creating trouble for us.
Can we remove this please?
As for code, we will pass through environment variables and the environment variable files can be local and added to `.gitignore`

test/e2e/nightfall-sdk.test.mjs
```diff
@@ -733,6 +1295,73 @@ describe('Testing the Nightfall SDK', () => {
     });
   });
 
+  describe('Restoring tokens', () => {
```
ChaitanyaKonda, line 1298: `Restoring tokens`, please explain the purpose of this test

ChaitanyaKonda (dismissed):

ChaitanyaKonda (dismissed):

ChaitanyaKonda:
Tested with E2E, wallet tests and UI.

- E2E passed well

- UI did not show 1155 functionality to test it
Screenshot 2022-01-17 at 18 39 05

- Wallet tests (./start-nightfall -wt -s) are failing with error 
Screenshot 2022-01-17 at 20 15 59

Additional changes are required but we can handle that it in a different issue to have this PR merged soon.
I created a new [issue](https://github.com/EYBlockchain/nightfall_3/issues/413) for this

ChaitanyaKonda (changes requested):
Comments made in the files changed section

cli/lib/tokens.mjs
druiz0992: No, Wallet doesn’t have access to signing key. `encodeABI` is a flag to let the approve function be sent and signed by metamask

nightfall-deployer/contracts/Shield.sol
druiz0992: Shield contract was incorrect. An ERC721 can have id 0

wallet/README.md
druiz0992: I agree, but this key is present in many places in the repo. Just do a search.

test/e2e/nightfall-sdk.test.mjs
druiz0992: It’s transferring the tokens back to the original account. For NFT it’s mandatory, otherwise we test would fail the second time it’s launched as user doesn’t have that specific token Id. For the rest of tokens, it’s good practice because in testnet launching this test many times would make a user run out of tokens.

druiz0992:
@ChaitanyaKonda , thanks for the thorough review. Regarding why ERC1155 tokens don’t show up, my guess is that you probably are working with a saved password, and the configuration is stored in local storage. I think if you clear the local storage, the the default configuration with mock erc1155 tokens will show.

Regarding second issue, I think there is a problem in your laptop. Try to  prune dockers and do an new setup. I know it works because that what GitHub actions are running.

cli/lib/nf3.mjs
druiz0992: You're right.

doc/lib/nf3.mjs.html
druiz0992: done

cli/lib/nf3.mjs
druiz0992: done

cli/lib/tokens.mjs
ChaitanyaKonda: Should this `ethereumSigningKeyExists` because we seem to be checking with condition `!!ethereumSigningKeyExists` in the function that calls it?

doc/lib/nf3.mjs.html
ChaitanyaKonda: It was my mistake. I meant `get`. Apologies

nightfall-deployer/contracts/Shield.sol
ChaitanyaKonda: That is true according to [EIP](https://eips.ethereum.org/EIPS/eip-721)
> While some ERC-721 smart contracts may find it convenient to start with ID 0 and simply increment by one for each new NFT, callers SHALL NOT assume that ID numbers have any specific pattern to them, and MUST treat the ID as a “black box”.

On other hand value does not exist in ERC721. It is essentially a null field. Because solidity does not have null as a concept, we use 0. There can't be any value including 1 for ERC721 because value does not represent count. 

Can we change this check to  `if (t.value != 0)`?

wallet/README.md
ChaitanyaKonda: That is true. 
I will create a new issue for having this removed instead of handling this here
The history in github will still hold this key even after removing this key, we might have to change this to new keys while tackling the issue and not push it. I will mention this in the issue too

nightfall-deployer/contracts/Shield.sol
```diff
@@ -195,7 +195,7 @@ contract Shield is Stateful, Structures, Config, Key_Registry {
           ''
         );
     } else if (t.tokenType == TokenType.ERC1155) {
-      if (t.tokenId == ZERO && t.value == 0) // disallow this corner case
+      if (t.value == 0) // disallow this corner case
```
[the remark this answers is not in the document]
druiz0992: Original code didn't allow 0 value for ERC1155. We didn't modify this. Do you want us to allow ERC1155 to have 0 value?

test/e2e/nightfall-sdk.test.mjs
ChaitanyaKonda: So this is done to run the same tests again with restarting nightfall ? Essentially making the test file idempotent?

Also, can we add some comments on top for situations such as this please, would help while reviewing or for whoever works in the same test file later?

cli/lib/tokens.mjs
druiz0992: For the wallet case `!!ethereumSigningKey` is false, so approve will be executed as
 ``` await ercContract.methods
          .approve(spenderAddress, APPROVE_AMOUNT)
          .send({ from: ownerAddress });
```

nightfall-deployer/contracts/Shield.sol
druiz0992: ok

test/e2e/nightfall-sdk.test.mjs
druiz0992: done

nightfall-deployer/contracts/Shield.sol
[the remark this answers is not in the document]
druiz0992: i allow erc1155 tokens to have 0 value

nightfall-deployer/contracts/Shield.sol
ChaitanyaKonda: If we need balance for ERC721 then we need to change how we calculate balance after querying the database in Client. This will be different from that of ERC20 . The shield contract should not be written to cater for this. 
The logic to get balance of ERC721 would be to count ERC721 tokens from database that are not `onChain` and not `isNullified`?

druiz0992: as agreed, i maintain check of erc720 value to 0 and disable erc720 tests. There will be a new issue to describe the problem and track the solution

nightfall-deployer/contracts/Shield.sol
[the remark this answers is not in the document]
ChaitanyaKonda: This line wasn't yet reverted in 198 and 234 of `Shield.sol`.
I might have missed on what was changed when you say done ?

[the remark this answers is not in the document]
druiz0992: I think i may have finally broken github. Check file here https://github.com/EYBlockchain/nightfall_3/blob/david/wallet-erc1155/nightfall-deployer/contracts/Shield.sol#L187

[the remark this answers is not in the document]
ChaitanyaKonda: https://github.com/EYBlockchain/nightfall_3/blob/david/wallet-erc1155/nightfall-
deployer/contracts/Shield.sol#L187 is the resolution in another issue I raised above. 
We resolved that in that issue and closed it.

This is a different issue for ERC1155.  
The lines that are affected here are 198 and 234 for ERC1155 and not 187 for ERC721.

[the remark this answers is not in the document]
druiz0992: We agreed on the following:
- fungible tokens can have value of 0.
- NFT must have value of 0.
- ERC1155 and ERC721 can have an id 0

Then, if the above is true, it is possible for a ERC1155 to have value of 0 and token id 0. Why should i keep this `if (t.tokenId == ZERO && t.value == 0)` then?

doc/lib/nf3.mjs.html
ChaitanyaKonda: This hasn't been changed. But accepting for now. Please change in a future PR to not have further delay for this merge

ChaitanyaKonda (approved):

PR #387, opened 2021-12-20
daveroga/Remove subdirectory-level Dockerfiles

### LINKED ISSUE #374: Remove subdirectory-level Dockerfiles
We now have two Dockerfiles for each component. These are the ones in the root directory, used to make docker images, and those in each component subdirectory, used by `docker-compose.yml`.  We should only have one because of the DRY principle, and also some code is broken because we're not updating both.

It seems best to keep the ones in the root because these don't have the problem that their context is limited a subdirectory. The latter makes it hard to incorporate code which is common across subdirectories.

daveroga:
Fixes #374 .

We now have two Dockerfiles for each component in root directory and subdirectory used by docker-compose.yml. We should only have the ones in root directory.

While pushing changes CheckPRs failed in e2e test in Liquidity provider because approve function in test was missing and also some changes in sdk and synchronization issues. 
Also I noticed that proposer3 was using the same address as liquidity provider and causing some issues about nonce.  I have removed the proposer3 tests and as there is an issue about refactoring e2e test al refactoring issues will be covered there.


test/e2e/nightfall-sdk.test.mjs
```diff
@@ -164,34 +164,50 @@ describe('Testing the Nightfall SDK', () => {
         web3,
       );
       // approve tokens to be advanced by liquidity provider in the instant withdraw
+      let txDataToSign;
       try {
-        await approve(
+        txDataToSign = await approve(
```
ChaitanyaKonda, line 169: Approval should only be done in bulk once per LP until the approved limit is reached and then can be done again.
Not for each time an instant withdrawal is requested because this would cost more gas

ChaitanyaKonda (changes requested):

test/e2e/nightfall-sdk.test.mjs
druiz0992: This is just in a test. Is it really necessary to change?

druiz0992 (approved):

test/e2e/nightfall-sdk.test.mjs
daveroga: Yes, it's only to test 1 instant withdraw in the e2e test that will be transferred by the LP.

ChaitanyaKonda: Test files are written with dual intent - for testing purposes as well as acting as a good reference for future/other developers. In the case of this test, that would be reference for SDK and how to build dapps for Nightfall correctly.
As a result, these have to be written to good standards just as any non test code we write.

Solving this would mean simply moving the `approve` out of event handling code to the phase when we register this LP.

ChaitanyaKonda:
This PR seems to cover more than just the removing docker files.
Could you please explain the other changes?

ChaitanyaKonda (changes requested):
This PR seems to cover more than just the removing docker files.
Could you please explain the other changes?

daveroga:
This PR cover more than just the removing docker files because e2e test failed when submitted changes for the PR and tried to solve them.

daveroga:
> 

It seems `./wallet` is used only for testing purposes talking with D.Ruiz. So no need to deploy with kubernetes.

test/e2e/nightfall-sdk.test.mjs
daveroga: You are right. `approve` in LP moved outside after initialization of the LP and ercAddress needed to approve.

daveroga: You are right. `approve` in LP moved outside after initialization of the LP and ercAddress needed to approve.

daveroga:
Solved merge conflicts with rebase master. Although e2e is not working because there is an issue #427 that should be solved first.

ChaitanyaKonda (approved):

Issue #388, opened 2021-12-27
Error in balances when transfer back between users


daveroga:
Error detected in balances when transfer from user1 to user2 account and then transfer back again from user2 to user1 (ERC20, ERC721 tested).
It seems that commitments of the second transfer (from user2 to user1) is not mined in any block.

Example:

- 6 deposits to user1 account
- 2 transfer from user1 to user2 account
- 2 transfer from user2 to user1 account
- 2 deposit to user1 account

Here are the final commitments in `nightfall_commitments` database in the client1 where you can see that commitment with id `0x04f6cf4f966d291666cb0032b546d4d4098af3bc003816a59dbba15bddf614ae`and commitment with id `0x068dd51952a30b4bd64df8da8375f035666232bfa41d136038858d0bc57fbdd1` are not on chain either in in any block, but yes all the previous and the subsequent from deposit txs:
```
[...]

PR #390, opened 2021-12-29
daveroga/blockProposedEventHandler not updating transfer commitment

daveroga:
Fix issue #388 .

It seems the problem is in nightfall-client in `block-proposed.mjs` when trying to store the commitment it throws an error of duplicate key for some reason.
- duplicate key error collection: nightfall_commitments.commitments index: _id_ dup key: { _id: "0x1259f9c3d5853a758c859eb274ca22c3019cea26f73e2318d1d90d4daf417792" } MongoError: E11000 duplicate key error collection: nightfall_commitments.commitments index: _id_ dup key: { _id: "0x1259f9c3d5853a758c859eb274ca22c3019cea26f73e2318d1d90d4daf417792" }

```
async function blockProposedEventHandler(data) {
...
  await Promise.all(storeCommitments);
  return Promise.all([
      markOnChain(nonZeroCommitments, block.blockNumberL2, data.blockNumber, data.transactionHash),
      markNullifiedOnChain(
        nonZeroNullifiers,
        block.blockNumberL2,
        data.blockNumber,
        data.transactionHash,
      ),
    ]);
...
}
```

This error is causing the second Promise.all that should mark the commitment as on chain and nullified on chain not being executed.
It can be solved by controlling the error in the first Promise as shown bellow.

```
async function blockProposedEventHandler(data) {
...
  await Promise.all(storeCommitments).catch(function (err) {
    logger.info(err);
  });
  return Promise.all([
      markOnChain(nonZeroCommitments, block.blockNumberL2, data.blockNumber, data.transactionHash),
      markNullifiedOnChain(
        nonZeroNullifiers,
        block.blockNumberL2,
        data.blockNumber,
        data.transactionHash,
      ),
    ]);
...
}
```

Other question is why here it throws the commitment as being duplicated and not in other transfer transactions that will be investigated later.

### DIFF
--- a/nightfall-client/src/event-handlers/block-proposed.mjs
+++ b/nightfall-client/src/event-handlers/block-proposed.mjs
@@ -62,7 +62,9 @@ async function blockProposedEventHandler(data) {
         }
       });
     }
-    await Promise.all(storeCommitments);
+    await Promise.all(storeCommitments).catch(function (err) {
+      logger.info(err);
+    }); // control errors when storing commitments in order to ensure next Promise being executed
     return Promise.all([
       markOnChain(nonZeroCommitments, block.blockNumberL2, data.blockNumber, data.transactionHash),
       markNullifiedOnChain(
--- a/nightfall-client/src/services/keys.mjs
+++ b/nightfall-client/src/services/keys.mjs
@@ -75,7 +75,10 @@ export async function generateKeys(mnemonic, path) {
 }
 
 export function storeMemoryKeysForDecryption(ivk, nsk) {
-  return Promise.all([ivks.push(...ivk), nsks.push(...nsk)]);
+  return Promise.all([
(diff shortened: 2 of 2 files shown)


daveroga:
Investigating the issue I've found that in the tests I use `nf3.subscribeToIncomingViewingKeys` explicitly in a test check, but this is already done in `setzkpKeys` or `setzkpKeysFromMnemonic` in nf3. 
The issue comes because when  subscribeToIncomingViewingKeys is called and with this the endpoing `/incoming-viewing-key`of the client, the `storeMemoryKeysForDecryption` function doesn't check if the key is already in the ivks collection and then is duplicated.

ChaitanyaKonda (approved):

IlyasRidhuan (approved):

PR #391, opened 2022-01-05
docs: add g16 malleability write up

### LINKED ISSUE #298: Implement knowledge of signature for public inputs for good practice against malleability
The results from this paper https://eprint.iacr.org/2020/811.pdf means that because an adversary should not be able to mess with the public input and malleability should not exist. To be on the safe side and follow good practices, it is recommended to prove knowledge of a signature on the public inputs inside the SNARK. This shouldn't add too much overhead and it also shouldn't interfere with privacy because the signature is never revealed in the clear. 
Thanking Mary Maller for her insight her.

MirandaWood:
Added some documentation to explain how NF is not vulnerable to the G16 malleability. Closes #298.


dwebchapey (approved):
Even if a front runner change the proof , witness data cannot be modified. Therefore the public inputs/outputs will be same. So correct fees will go to the original recipient even if malicious operator uses the modified proof.

IlyasRidhuan (approved):

Issue #355, opened 2021-12-07
Rationalise configs


Westlad:
The configuration for this application is becoming fragmented.  It now exists in:

- Some `constants.mjs` files.
-  Environment files
- `config` module files
- `env-cmd` module files

Probably it is sensible just to use one of these.

PR #334, opened 2021-11-29
configuration for hosted geth implemented

LijuJoseJJ:
this fixes #318 

Two temporary successful workflow run, which run integration test on hosted geth settings
https://github.com/EYBlockchain/nightfall_3/runs/4348622337?check_suite_focus=true
https://github.com/EYBlockchain/nightfall_3/runs/4341132049?check_suite_focus=true

PR #398, opened 2022-01-11
daveroga/rationalize configs

daveroga:
Fixes issue #355 .

The configuration for this application is becoming fragmented. It now exists in:

- Some constants.mjs files.
- Environment files
- config module files
- env-cmd module files

Probably it is sensible just to use one of these.

Steps to solve this:
- Used config module files for tests and removed constants.mjs.
- Removed env-cmd module files.
- Tried to use config module in cli but then found some problems with babel/webpack and the wallet. Some constants are cli specific and should be in the cli project.


daveroga:
- Used config module files for tests and removed constants.mjs.
- Removed env-cmd module files.
- Tried to use config module in cli but then found some problems with babel/webpack and the wallet. Couldn't still solve it.

config/default.js
```diff
@@ -126,4 +126,74 @@ module.exports = {
...
+      optimistWsUrl: '',
+      web3WsUrl: '',
+    },
+    ropsten: {
+      name: 'Ropsten',
+      chainId: 3,
+      clientApiUrl: 'https://client1.testnet.nightfall3.com',
+      optimistApiUrl: 'https://optimist1.testnet.nightfall3.com',
+      optimistWsUrl: 'wss://optimist1-ws.testnet.nightfall3.com',
+      web3WsUrl: 'wss://ropsten1-ws.testnet.nightfall3.com',
```
druiz0992, line 144: Can we remove web3WsUrl from code?

druiz0992 (changes requested):
I would like to keep web3 node url out of repo code. PR #334 proposes a way to do it. I suggest once it is approved, just reuse it

config/default.js
daveroga: It has been removed in last commit `fix: remove testnet Ropsten URL from config`

daveroga:
PR already solved but e2e test failing due to issue #427  that should be solved first.

daveroga:
ganache-test seems failing sometimes. Forced modify a file to test it again.

ChaitanyaKonda:
This seems to add 263 commits in total. Most of which are commits that already exist in the master.
Merging this has the risk of breaking the code and adding back logic that we removed. 

Could you revert this to just the commits from this PR please?

daveroga:
Wrong commits added after rebase solved.

signorecello (approved):
LGTM but just mind some tests are being removed by #490

PR #401, opened 2022-01-12
feat: -d or --dev flag

signorecello:
Added the -d flag so we can use the same start-nightfall script to deploy without bind mounts. This is useful for deploying to testnet from non-dev machines (essencially replicating the same as ping-pong's `./pong-nightfall` script)


docker-compose.dev.yml
```diff
@@ -0,0 +1,111 @@
+version: '3.5'
+# Use this script for running up nightfall_3 in 'developer' mode with local
+# bindings.  See the readme for more information.
+services:
+  client1:
+    build:
+      dockerfile: Dockerfile
+      context: ./nightfall-client
```
ChaitanyaKonda, line 8: These files were removed in this earlier [PR](https://github.com/EYBlockchain/nightfall_3/pull/387).
The context might need to be changed to root and the dockerfiles from root too

ChaitanyaKonda (changes requested):
As commented

ChaitanyaKonda (approved):

PR #402, opened 2022-01-13
fix: move initialclientsync to incomingviewingkey

### LINKED ISSUE #269: Revert Client syncing changes
We had to move `initialClientSync` to trigger on `/incoming-viewing-key` rather than on `main` ; however it’s now back in `main`. I’m not sure if this will work because this means that client will attempt to resync its local state before it has the necessary keys loaded to decrypt any potential commitments from the blocks it is resyncing.  We may need a new PR that splices the old and new changes together.

The change was initially made in #144  but then reverted in #223

dwebchapey:
This closes #269 .

The transactions are now stored in client database. So when new ivk(s) are received , client need not run past event handlers again instead it can fetch the transactions from TRANSACTIONS_COLLECTION and look for possible commitments.

The method `clientCommitmentSync` is called on new ivk arrival which fetches all tranasctions from db via `getAllTransactions` and look for commitments by passing transactions to `decryptCommitment`. `decryptCommitment` can also be used `blockProposedEventHandler` to decrypt commitments of new transactions.


ChaitanyaKonda:
It seems like we are queueing the client sync event after storing the new decryption keys.
Now this will still start a new syncing thread besides the one that is already running (started at the start of the client service), won't it?
In this case, the more the keys we will add, the more the number of syncs we will have.

The intialSync should be written such that if new keys were added, it needs to start sync for these new keys from a certain block number that will have the first commitment sent to this key's address. Ideally better if this sync function does not sync again for old keys from that block number.

ChaitanyaKonda:
> It seems like we are queueing the client sync event after storing the new decryption keys. Now this will still start a new syncing thread besides the one that is already running (started at the start of the client service), won't it? In this case, the more the keys we will add, the more the number of syncs we will have.
> 
> The intialSync should be written such that if new keys were added, it needs to start sync for these new keys from a certain block number that will have the first commitment sent to this key's address. Ideally better if this sync function does not sync again for old keys from that block number.

@dwebchapey

dwebchapey:
> It seems like we are queueing the client sync event after storing the new decryption keys. Now this will still start a new syncing thread besides the one that is already running (started at the start of the client service), won't it? In this case, the more the keys we will add, the more the number of syncs we will have.
> 
> The intialSync should be written such that if new keys were added, it needs to start sync for these new keys from a certain block number that will have the first commitment sent to this key's address. Ideally better if this sync function does not sync again for old keys from that block number.

Ok . I think I understand . In the `blockProposedEventHandler` instead of decrypting for all `incoming viewing keys `, it should do only for the new key ?  It will still start a new thread of `intialClientSync` for new ivks, but `Eventhandlers` will be ivk specific .

IlyasRidhuan:
> The intialSync should be written such that if new keys were added, it needs to start sync for these new keys from a certain block number that will have the first commitment sent to this key's address. Ideally better if this sync function does not sync again for old keys from that block number.

@dwebchapey  I don't see this suggested logic in the current PR. In fact, it looks like you removed logic, namely `initialSyncState`, that handles a "smarter" way to resync. Right now it looks like whenever you add incoming keys, we are syncing the entire state from scratch again?

dwebchapey:
> > The intialSync should be written such that if new keys were added, it needs to start sync for these new keys from a certain block number that will have the first commitment sent to this key's address. Ideally better if this sync function does not sync again for old keys from that block number.
> 
> @dwebchapey I don't see this suggested logic in the current PR. In fact, it looks like you removed logic, namely `initialSyncState`, that handles a "smarter" way to resync. Right now it looks like whenever you add incoming keys, we are syncing the entire state from scratch again?

With the current way initialClientSync is written , it does 2 checks . 

1.  If commitments in db , no sync at all
2. If commitments , sync only from the blocknumber of the earliest commitment


But if a new ivk is added , there could be commitments from earlier blocks that were not added to the db previously.

nightfall-client/src/services/commitment-sync.mjs
```diff
@@ -0,0 +1,56 @@
...
+    const decompressedSecrets = Secrets.decompressSecrets(transaction.compressedSecrets);
+    logger.info(`decompressedSecrets: ${decompressedSecrets}`);
+    try {
+      const commitment = Secrets.decryptSecrets(decompressedSecrets, key, nonZeroCommitments[0]);
+      if (Object.keys(commitment).length === 0)
+        logger.info("This encrypted message isn't for this recipient");
+      else {
+        // console.log('PUSHED', commitment, 'nsks', nsks[i]);
+        logger.info('This is encrpted and stored');
+        storeCommitment(commitment, nsk[j]);
```
IlyasRidhuan, line 30: I'm not entirely sure if calling `storeCommitment` (which is async) in a `forEach` will work the way we want it. You might need to push into an array that is then return a `Promise.all(...)` at the end or use a `map` and wrap that in a `Promise.all(..)`

dwebchapey: yeah . It did work in my limited testing . But I have added the necessary changes and pushed new commit.

nightfall-client/src/services/commitment-sync.mjs
```diff
@@ -0,0 +1,59 @@
...
+  });
+}
+
+/**
+Called when new ivk(s) are recieved , it fetches all available commitments 
+from commitments collection and decrypts commitments belonging to the new ivk(s).
+*/
+export async function clientCommitmentSync(ivk, nsk) {
+  const transactions = await getAllTransactions();
+  if (typeof transactions !== 'undefined') {
```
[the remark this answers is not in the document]
dwebchapey: It should not . It threw me an error once in an earlier version, so I kept as a safe check.

IlyasRidhuan (approved):

ChaitanyaKonda (approved):

PR #404, opened 2022-01-13
Build Zokrates in Dockerfile

### LINKED ISSUE #399: M1 Mac compatibility
As explained in #396, adding the `platform` keyword is a temporary fix, this is because relying on `qemu` to emulate an x86 environment is slower and can be buggy.

A more robust solution would be to enable the `zokrates` binary to be compiled natively to arm. [The latest 0.7.10](https://github.com/Zokrates/ZoKrates/releases) release of Zokrates appears to enable M1 compilation, but is not yet available as a docker image.

We could either dockerize our own 0.7.10 release and use it as we use 0.7.7 now, or build the 0.7.10 binary directly inside our docker file (this would likely be better as we only use the `zokrates` docker image so we can copy out the binary anyways)

IlyasRidhuan:
This PR closes #399 and superseded #396.

Zokrates 0.7.10 is now directly built in `optimist` and `worker` containers. This is achieved by using the rust image in the first build stage and cloning the appropriate Zokrates release.

There is a constraint when compiling rust binaries; in particular, glibc is dynamically linked at compile time and the binary expects to have the same version of glibc when the binary is compiled and when it is run. As we require a slightly later version of rust to compile Zokrates successfully for ARM, we also need an image in stage 2 of the build that has a later version of glibc  - the result is the following:

1. In `optimist` we use a later version of the mongo image (based on ubuntu `focal` - 20.04).
2. In `worker`, I'm hesitant to bump the node version past node@14, so instead we pull in the ubuntu:20.04 image and install node@14. This does lead to a slightly larger final image, but if that becomes a future issue we can use a slimmer version or `alpine`.

Finally, after Zokrates 0.7.9, the `--light` flag has been removed from the `compile` command as it is now default behaviour. It has been removed.


Westlad:
I made a PR #428 that creates a zokrates container image, from which the binary is copied.  Saves about 10 mins on my machine because the binary is only built once

Westlad (approved):

Westlad:
My PR breaks the wallet tests so maybe leave it for now.

IlyasRidhuan:
It's not clear to me why it is breaking the tests though.. I experienced the same thing with this PR so I have re-run the workflow just in case

Westlad:
ooh - it works now.  Must be a timing thing.  @druiz0992 has changed the timing in the wallet tests in PR #429 so hopefully this is a temporary issue and will go away when the ping-pong tests are merged.

ChaitanyaKonda (approved):

PR #432, opened 2022-01-27
fix: remove length check in invalid txn typecheck

### LINKED ISSUE #424: Bug: Good blocks can be successfully challenged for 'Invalid Transaction Type'
@dwebchapey has found a bug in which good blocks can be challenged for having an invalid transaction type.

The error stems from `ChallengesUtil.sol` incorrectly checking the _length_ of an array (e.g. expecting a single transfer transaction to have one new commitment). We do not differ array lengths by tx type, but rather use zero values.

One example is any correct single transfer failing since its transaction object has `commitments.length == 2`, with `commitments[0]` being the new commitment and `commitments[1]` being zero. This can be challenged since the contract believes single transfers should have `transactions.commitments.length == 2`.

Bug discovered when addressing #357 (adding this as a new issue since it's a bug, not part of new test infrastructure).

dwebchapey:
This fixes #424 
Good blocks were successfully challenged previously because of the length checks in libChallengeTransactionType() . 
Fixing the transaction type checks in contract accordingly.

### DIFF
--- a/nightfall-deployer/contracts/ChallengesUtil.sol
+++ b/nightfall-deployer/contracts/ChallengesUtil.sol
@@ -82,7 +82,7 @@ library ChallengesUtil {
                 transaction.commitments[1] != ZERO ||
                 transaction.nullifiers[0] != ZERO ||
                 transaction.nullifiers[1] != ZERO ||
-                transaction.compressedSecrets.length != nZeroCompressedSecrets ||
+                nZeroCompressedSecrets != 8 ||
                 nZeroProof == 4 || // We assume that 3 out of the 4 proof elements can be a valid ZERO. Deals with exception cases
                 transaction.historicRootBlockNumberL2[0] != 0 ||
                 transaction.historicRootBlockNumberL2[1] != 0,
@@ -168,7 +168,7 @@ library ChallengesUtil {
                 transaction.commitments[1] != ZERO ||
                 transaction.nullifiers[0] == ZERO ||
                 transaction.nullifiers[1] != ZERO ||
-                transaction.compressedSecrets.length != nZeroCompressedSecrets ||
+                nZeroCompressedSecrets != 8 ||
                 nZeroProof == 4 || // We assume that 3 out of the 4 proof elements can be a valid ZERO. Deals with exception cases
                 transaction.historicRootBlockNumberL2[1] != 0, // A withdraw has a similar constraint as a single transfer
             'This withdraw transaction type is valid'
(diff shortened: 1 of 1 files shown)


nightfall-deployer/contracts/ChallengesUtil.sol
```diff
@@ -76,9 +76,6 @@ library ChallengesUtil {
                 transaction.recipientAddress != ZERO ||
                 transaction.commitments[0] == ZERO ||
                 transaction.commitments[1] != ZERO ||
-                transaction.commitments.length != 1 ||
-                transaction.nullifiers.length != 0 ||
```
IlyasRidhuan, line 80: Does removing this nullifiers check mean we now need to re include `nullifiers[0] === ZERO` and `nullifiers[1] === ZERO` in this conditional?

nightfall-deployer/contracts/ChallengesUtil.sol
```diff
@@ -76,9 +76,6 @@ library ChallengesUtil {
...
-                transaction.compressedSecrets.length != 0 ||
```
IlyasRidhuan, line 81: Do we now need to check that the `compressedSecrets` array is full of `0x00..00`s

nightfall-deployer/contracts/ChallengesUtil.sol
dwebchapey: Added nullifier zero check

nightfall-deployer/contracts/ChallengesUtil.sol
dwebchapey: Added all zero compressedSecrets check

IlyasRidhuan (approved):

nightfall-deployer/contracts/ChallengesUtil.sol
```diff
@@ -156,15 +156,19 @@ library ChallengesUtil {
...
                 transaction.ercAddress == ZERO ||
                 transaction.recipientAddress == ZERO ||
-                transaction.commitments.length != 0 ||
+                transaction.commitments[0] != ZERO ||
+                transaction.commitments[1] != ZERO ||
                 transaction.nullifiers[0] == ZERO ||
                 transaction.nullifiers[1] != ZERO ||
-                transaction.nullifiers.length != 1 ||
-                transaction.compressedSecrets.length != 0 ||
+                transaction.compressedSecrets.length != nZeroCompressedSecrets ||
```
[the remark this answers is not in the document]
ChaitanyaKonda: Actually I take my comment back. The exception case I suggested is wrong because the length of the array is always 8 and will never be `transaction.compressedSecrets.length == 4`

But can I request that we do the straight forward check of `nZeroCompressedSecrets != 8 ||` here please?
This logic you wrote is quite clever but it takes a long time to interpret the negations and anyone looking at might be as confused to until a much deeper look. Apologies for the ask Hari :)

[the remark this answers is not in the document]
dwebchapey: No worries . Changes done.

ChaitanyaKonda (approved):

PR #455, opened 2022-02-02
pub-sub service to communicate new blocks

### LINKED ISSUE #431: Subscription system for client-less browsers
Research a system capable to broadcast Optimist's nightfall state so that  client-less wallets can update the state without the need to be constantly polling.

druiz0992:
Closes #431 

PR shows an example of  pub-sub service that uses AWS IOT service to deliver block proposed status updates to subscribers. It is thought as a nightfall service to deliver updates to client-less browser wallets.

The service is defined in test/pub-sub. It deploys the ping-pong test to have a pair of clients generating transactions. Then, there is a publisher node that detects that blocks have been generated, and submits a message using AWS  IOT service to all clients that have subscribed to the topic. There is also a subscriber that receives these messages.

Test can be launched by:
```
cd test/pub-sub
./ganache-standalone -s
./pub-nightfall -s -d
./pub-apps
```
You should see in the `subscriber` and `publisher` container logs the test result.

To stop:
```
./pub-down -d
```
You will also need to stop the ganache container manually.

Moving forward, I think that an Emitter function for Rollback Events needs to be added to the SDK (similarly to NewBlock Emiiter) so that the publisher can call it and receive Rollback events to alert subscribers.


ChaitanyaKonda:
This PR shows an example of pub-sub service using an AWS IoT.
This will be maintained as a branch instead of merging into master because master does not work with the example functionality.

PR #456, opened 2022-02-03
(nightfall-browser) remove sdk Nf3 init

### LINKED ISSUE #451: Remove last reference to NF3 SDK
In `wallet/index.jsx` we reference the sdk when initializing the nf3 state.
```
const nf3 = await new Nf3.Nf3(nf3Env.web3WsUrl, '', nf3Env);
await nf3.init();
```
This call sets the following environment variables from `nf3.mjs`
```
 constructor(web3WsUrl, ethereumSigningKey, environment = ENVIRONMENTS.localhost, zkpKeys) {
    this.clientBaseUrl = environment.clientApiUrl; ---> Not needed
    this.optimistBaseUrl = environment.optimistApiUrl;
    this.optimistWsUrl = environment.optimistWsUrl;
    this.web3WsUrl = web3WsUrl;
    this.ethereumSigningKey = ethereumSigningKey;
    this.zkpKeys = zkpKeys; --> already set in browser
  }
```

we should be able to replace this initialisation call and set these variables directly in the browser. In reality this information should be set at the time of webpage is bundled.

LijuJoseJJ:
this PR fixes #451 

Note: please test this PR by running UI


wallet/src/hooks/User/index.jsx
```diff
@@ -25,5 +33,88 @@ export const UserContext = React.createContext({
...
+    setupWebSocket();
+  }, [state.zkpKeys]);
+
+  React.useEffect(() => {
+    configureZkpKeys(state.mnemonic);
+  }, [state.mnemonic]);
+
+  return (
+    <UserContext.Provider value={[state, setState, configureMnemonic]}>
+      {isSyncComplete ? children : <div> Signing to MetaMask </div>}
```
IlyasRidhuan, line 117: Replace or comment out this check for now. It's a good idea but we should plan ahead with a loading modal rather than just a white page with text.

IlyasRidhuan (changes requested):

wallet/src/hooks/User/index.jsx
LijuJoseJJ: commented out for now

PR #477, opened 2022-02-08
Dev deployment

Westlad:
This PR creates a new directory containing scripts to deploy to the dev (`ohio.nightfall3.com`) server.  It's a fully containerised deployment and does not deploy any source code, other than scripts to start and stop the nightfall containers.  If the code is re-deployed, persistent volumes are preserved ie. the proving and verifying keys in `proving_files` and the contract addresses/abi in `build` are retained.  Thus, you should not normally need to update nightfall-browser etc.

The dev server is intended to work with `nightfall-browser` and therefore does not deploy `nightfall-client`.

The deployment to dev will only run when it is approved by one of the team.  This approval is separate from a PR approval and happens afterwards.  We would not normally want to deploy every merge into master because this may disrupt work that is using the dev server and, as the contract deployment is to Ropsten, it burns about 0.5 ETH.  Note also that it may fail if the gas price is >20 GWei.  This is to prevent excessive ETH consumption.

It's not possible to test this deployment locally.  You could push the dev branch and observe the action running.  You should reject the deployment however (to avoid disrupting work on the dev server) so it's not much of a test.


ChaitanyaKonda (approved):

ChaitanyaKonda:
> No it's not. The change in init is backwards compatible so older code will still run fine.

Oh yes, the second parameter defaults to undefined and it is handled later down in the function. 
My bad

Issue #474, opened 2022-02-07
Integrating PoS proposal into PNF3: proposer staking


daveroga:
Features:
- Add operations in the contracts for staking ETH. In the future decide if it's ETH or MATIC.
- Add endpoints for this proposers staking in optimist
- Add operations to sdk for this proposers staking
- This staking will be the base of the staking power when running spans of the weighted round robin algorithm for proposers selection in next task

Discuss:
- Tx fee payments could go to proposer's stake instead of adding it to proposer pending withdrawal.

PR #486, opened 2022-02-09
Integrating PoS proposal: proposers staking

daveroga:
Feat #474, #488 

- `registerProposer` moved to `stakeProposer(amount)`. This registers the proposer and if the proposer is already registered it increases the proposer stake amount.
- `deregisterProposer` moved to `unstakeProposer`. 
- No `REGISTRATION_BOND` now but a `MINIMUM_STAKE` (10 wei for testing now).
- `COOLING_OFF_PERIOD` moved to `CHALLENGE_PERIOD` (1 week).
- Changed messages from `require` statements in contracts to best practices used by Openzeppelin. `<contract name>: <message>`. So you know the contract where the message has been thrown. Useful now when debugging.
- `withdrawBond` moved to `withdrawStake` and used in the same way but with the stake of the proposer instead of the bond, so the proposer can do the withdraw of their stake. Previously to that, the proposer should call `unstakeProposer` to stop being proposer and wait this `CHALLENGE_PERIOD` to be allowed to do the withdraw.
- Moved `uint` to `uint256` in contracts to avoid errors and follow the same guide in all the contracts.
- `TimeLockedBond` moved to `TimeLockedStake` and added `challengeLocked` to store the amount locked by blocks proposed still in `CHALLENGE_PERIOD` and not claimed yet.
```
    struct TimeLockedStake {
        uint256 amount; // The amount held
        uint256 challengeLocked; // The amount locked by block proposed still in CHALLENGE_PERIOD
        uint256 time; // The time the funds were locked from
    }
```
- Need to request block payment for proposers. (#488).
- Created test for testing PoS features. `npm run test-pos`

Questions:
- Tx fee payments could go to proposer's stake instead of adding it to proposer pending withdrawal.

This PR is included in #499 .

### DIFF
--- a/cli/lib/nf3.mjs
+++ b/cli/lib/nf3.mjs
@@ -584,14 +584,14 @@ class Nf3 {
   }
 
   /**
-  Withdraw the bond left by the proposer.
-  It will use the address of the Ethereum Signing key that is holds to withdraw the bond.
+  Withdraw the stake left by the proposer.
+  It will use the address of the Ethereum Signing key that is holds to withdraw the stake.
   @method
   @async
   @returns {Promise} A promise that resolves to the Ethereum transaction receipt.
   */
-  async withdrawBond() {
-    const res = await axios.post(`${this.optimistBaseUrl}/proposer/withdrawBond`, {
+  async withdrawStake() {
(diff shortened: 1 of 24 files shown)


ChaitanyaKonda:
Closing because a new PR to feature POS will replace it

Issue #413, opened 2022-01-17
Refactor nightfall-sdk.test.mjs file because it is too big and hard to follow


ChaitanyaKonda:
1. Remove `ercmocks.test.mjs` because this is not related to NF3 testing. Would have been useful for dev purposes originally, I would think it is not required anymore 
2. Splitting nightfall-sdk.test.mjs. It is too big and the code is hard to follow today. 
The test files could be as defined below
    - Miscellaneous tests (Renamed to Health and Contract Checks) 
    - Proposer tests
    - Challenger tests
    - ERC20 tests (deposit, single transfer, double transfer, withdraw and instant withdraw; balance checks to be part of deposit, single transfer, double transfer, withdraw and instant withdraw sub tests not a separate test for balance. For example, after performing deposit, in the same test we can check for new balance in L2 or pending L2)
    - ERC721 tests (deposit, single transfer, NO double transfer and withdraw; balance checks to be part of deposit, single transfer and withdraw sub tests not a separate test for balance. For example, after performing deposit, in the same test we can check for new balance in L2 or pending L2)
    - ERC1155 tests (deposit, single transfer, double transfer and withdraw; balance checks to be part of deposit, single transfer, double transfer and withdraw sub tests not a separate test for balance. For example, after performing deposit, in the same test we can check for new balance in L2 or pending L2)
3. Remove `Get commitments tests` because this is inherently tested during transfer and withdraw
4. Remove `Get pending withdraw commitments tests` because this is tested when during `finaliseWithdrawal`. If `finaliseWithdrawal` withdrawal fails this too works. 
5. Rename `should allow instant withdraw of existing withdraw` to `should allow instant withdraw`. Instant withdrawal does not require and existing withdrawal. Normal withdraw has two levels of withdraw. To start withdraw and then to confirm withdraw after challenge period by retrieving balance to L1)
6. `Synchronize with block proposed` please explain the purpose of this test. Considering we are doing deposit tests below and this is the same, we could remove it.
7. Document code in more detailed where the logic is bit tricky to understand

PR #490, opened 2022-02-11
Splitting tests

signorecello:
So we needed e2e tests to be splitted as per #413 . I went ahead and made some more changes to improve them and make them easier to maintain on the long run. Here are some considerations to sustain this:

## Organization stuff
- I added a `no-await-in-loop` rule to `eslint.rc`, as we had about 70+ inline eslint rules exceptions. We can think of having them on a per-file basis though. As you know, we use await in loops to avoid nonce issues, mostly.
- I separated the e2e tests into `e2e-protocol` and `e2e-tokens`, don't mind the naming though. The idea is to have tests related to challengers, proposers, chain-reorgs, future PoS tests, whatever is protocol-related, inside this folder. The other one would be to test anything ERC-related.
- For each folder there's a `index.test.mjs` file that imports the tests from that folder. This allows us to skip specific files for development.
- I personally like tests to not have a lot of info logs, so it's easier to read the results without a lot of messages on the console. So I added a `VERBOSE` env variable you can pass to tests, to see all the info logs (ex. `VERBOSE=true npm run test-e2e-tokens`). Otherwise, only error or other very relevant messages show on the console.
- As discussed previously, I removed some unused tests like `chain-reorg`, etc.

## Big changes in tests
- We were doing some magic number stuff in tests that would break often, for example regarding transactionsPerBlock. I wanted tests to work with any transactionsPerBlock count by just setting an env variable. Also, tests would break when re-running them if there were any unprocessed transactions in the optimist mempool, as the L2 count would mismatch with the smart contract count, and transactions would fail. So I implemented a function to get the unprocessed transaction count, and another one that would attempt to empty the L2 mempool after every `describe`.
- We were using a `constants.mjs` file with a bunch of key-values, which was cumbersome to read. I created some `json` files to save them nice and tidy and easy to maintain.
- Web3 changes. So we had a bunch of things like `submitTransaction`, `getBalance`, `closeWeb3Connection`, etc that we would pass an instance of web3 to. I though it would make sense to make a class to abstract all the things we want with web3. So at the start of the tests we just get an instance of this class and use it without having to juggle with an web3 instance being passed around on every function call. It's also easy to clean up after each test. There's a getter `.getWeb3()` that will return you the web3 provider if you want to use the "raw" web3 provider.
- I don't like `logCounts` as it looks a bit naive for the uninitiated. I tried to remove it and I failed miserably, we really need it. As per the point above, I created a `subcribeTo` method on the new `Web3` class that hopefully abstracts the queueing of the logs. So `logCounts` is now `logs` and is an object with some arrays of logs inside, we just check the length if we actually need to count them. This way we can reuse the `subscribeLogs` method for every kind of blockchain event, `newBlockHeaders`, `transactionSubmitted`, `newCurrentProposer`, you name it. Easier to add some more events in the future.

## Some other weird changes
- As discussed previously, `getLayer2Balances` wasn't really working. I fixed that, and got rid of `filterByCompressedPkd` parameter as it wasn't being used anywhere. `getLayer2Balances` retrieves the nf3 instance's balance, and that's all. For the same reason I got rid of `/balances-details`, as we don't use it at all.


signorecello:
After talking with @druiz0992 I removed the wallet-test as it was failing inexplicably and the tests make little sense now given that we will have a brand new shiny wallet

daveroga:
We have to think about all config files related with tests. They were unified in the config module in the issue #398 about rationalizing configs.

signorecello:
> We have to think about all config files related with tests. They were unified in the config module in the issue #398 about rationalizing configs.

Hey, add me as a reviewer on that one, let's merge it as it's been here for a while. I'll solve the conflicts with this one. I'm having a hard time making the tests pass reliably anyway.

daveroga (approved):

cli/lib/nf3.mjs
```diff
@@ -186,7 +196,7 @@ class Nf3 {
     let tx;
     await this.nonceMutex.runExclusive(async () => {
       // if we don't have a nonce, we must get one from the ethereum client
-      if (!this.nonce) this.nonce = await this.web3.eth.getTransactionCount(this.ethereumAddress);
+      this.nonce = await this.web3.eth.getTransactionCount(this.ethereumAddress);
```
Westlad, line 199: If we do this, then we are not managing nonce increments ourselves. We used to think that web3 had trouble updating its nonces so we managed them locally, but given this is working, we were probably wrong and simply using a Mutex was sufficient.  Thus I think this code change is fine but then we should remove comments ln 198, 193

nightfall-client/src/app.mjs
```diff
@@ -25,7 +25,6 @@ app.use(cors());
 app.use(bodyParser.json({ limit: '2mb' }));
 app.use(bodyParser.urlencoded({ limit: '2mb', extended: true }));
 
-app.get('/healthcheck', (req, res) => res.sendStatus(200));
```
Westlad, line 28: Why remove this? We use it to check the server is up.  It will break the ping pong test if you remove it. I can easily fix that but it seems a helpful endpoint.

Westlad (commented):
Nice code!  Just a few minor comments.

cli/lib/nf3.mjs
```diff
@@ -164,6 +164,16 @@ class Nf3 {
     return this.subscribeToIncomingViewingKeys();
   }
 
+  /**
+  Gets the number of unprocessed transactions on the optimist
+  @async
+  */
+
+  async unprocessedTransactionCount() {
```
ChaitanyaKonda, line 172: Testing deposit, transfer, withdraw and instant withdrawal end to end is done from the perspective of user. 
Optimist is designed to be run by Proposers, Challengers and LPs and as such a user does not run their own instance of optimist other than wallet/nightfall-client.

The test uses this end point to query if the proposer has any transactions in the mempool before the user sends new transactions because the tests were written to work when re-running them. This meant unprocessed transactions for this user in the optimist mempool should not exist. In real usage, a user will never query a proposer for unprocessed transactions before sending new transactions. So we will have to remove this end point that the user can't use.

An equivalent approach to achieve this would be for the user to repeatedly poll their balance to see if their expected balance from the transactions they performed is equal to the balance their database has from processing Blocks Proposed. If not, they wait longer or propose. An example of this can be seen in `test/ping-pong/user-local/src/index.mjs` and in `test/adversary.test.mjs` in adversary2 branch.

test/e2e/configs.json
```diff
@@ -0,0 +1,7 @@
+{
+  "bond": 10,
```
ChaitanyaKonda, line 2: Could we use `config` in root directory for this?
It lets us set variables based on the environment by creating `environment.js` as defined by `config` module.

For some variables such as txPerBlock, we would want to pull this from the same config that optimist uses too as they will need to be aligned.

test/e2e/environments.json
```diff
@@ -0,0 +1,30 @@
+{
+  "mainnet": {
+    "name": "Mainnet",
+    "chainId": 1,
+    "clientApiUrl": "",
+    "optimistApiUrl": "",
+    "optimistWsUrl": ""
+  },
```
ChaitanyaKonda, line 8: This currently exists in config

ChaitanyaKonda (changes requested):
This refactor makes the test files more modular and clear to understand. 
This looks very clean and much easier to debug. 

I requested a few changes as suggested in earlier comments.

cli/lib/nf3.mjs
signorecello: web3 seems to work just fine managing nonces, at least I had 0 problems removing the local nonce management. Actually, some tests were failing exactly because of nonce conflicts. I'll just remove comments in 198, 193

nightfall-client/src/app.mjs
signorecello: I made this change as I was trying to make the github actions wait for a reaallly "healthy" service (as in, "after the client is synced"). Then i found out about `docker wait` which is a better solution. 

So... I didn't remove it, just put it [here](https://github.com/EYBlockchain/nightfall_3/blob/6ca1b30bf4d5a6dcea379cacafbd663b7a37e281/nightfall-client/src/index.mjs#L20). So the healthcheck only returns 200 after the client is synced. I thought it better reflects the "healthiness" of the service, so I left it there.

cli/lib/nf3.mjs
signorecello: Hi, I understand it makes little sense to be in `nf3`, but I wonder if we could just use `axios` to simply call the optimist's endpoint in the test itself? I didn't add that endpoint, it was there already, added by #137 .

cli/lib/nf3.mjs
Westlad: Yes, I think our current tests would expose any issues here so I'm good with this change.

nightfall-client/src/app.mjs
Westlad: My bad - hadn't spotted that.  This is a good improvement.

cli/lib/nf3.mjs
```diff
@@ -424,7 +438,7 @@ class Nf3 {
     });
     this.latestWithdrawHash = res.data.transaction.transactionHash;
     if (!offchain) {
-      const receiptPromise = this.submitTransaction(
+      const receiptPromise = await this.submitTransaction(
```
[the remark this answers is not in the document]
Westlad: I still think this await is not needed and could slow things?

Westlad (approved):

cli/lib/nf3.mjs
ChaitanyaKonda: When optimist is hosted in STAGING and PROD deployments, It will not expose any end point to the world.
This test simulates a non browser desktop application built using an SDK.
Since this application sits in the external world, it won't have access to optimist and will not be able to run against Staging and Prod deployments.

Duncan's architecture diagram would be useful to visualise this. 
image
Replace the nightfall browser with non browser application.

ChaitanyaKonda:
Fixes #504

cli/lib/nf3.mjs
ChaitanyaKonda: This will be solved in a separate issue #513

PR #491, opened 2022-02-11
docs: method for contract updating

Westlad:
Adds a description of how updates to contracts will work.  This will be required early on because otherwise we will not be able to bug-fix without losing users funds.


doc/contract-update.md
```diff
@@ -0,0 +1,56 @@
...
+contain the `upgrade` function.
+
+![contract interaction](./contract-upgrade.png)
+
+### upgrade function
+
+This function will, atomically, carry out the following actions:
+
+1. Repoint `State.sol`'s registered logic contracts to the new logic contracts;
+1. Repoint `Proxy.sol` to the new contracts;
```
signorecello, line 36: Just a quick note, I think `Proxy.sol` needs to exist for each contract, as all it does is a bit of assembly that points to another contract. So we will need multiple `Proxy.sol` contracts

Westlad: I was imagining one `Proxy.sol` that knows about all of your contracts. I suppose multiple contracts are more flexible if you wanted ever to add an entirely new contract to your fleet though.

doc/contract-update.md
```diff
@@ -0,0 +1,56 @@
...
+1. Change the addresses of the contracts registered with `State.sol` to the new contract addresses;
+1. Store the address of the old contracts and the block number at which the swap-over occurred in `State.sol`; this
+will be used to help nightfall applications parse historic events and calldata. This will be in the form of
+an array, added to each time the contracts are upgraded.
+1. Call `selfdestruct` on the old logic contracts (just to return gas);
```
signorecello, line 42: Not just gas, it's actually an **extremely important thing to do.** The address of the `State` contracts are [internal to the proxy contract](https://docs.openzeppelin.com/contracts/4.x/api/proxy#Proxy-_implementation--). So using the current implementation of `Proxy.sol` there is no way of getting the contract address... And very well so. We would need to add that functionality to `State.sol`

Let's say that I manage a token "ZP" at `ZPERC20.sol`, which is proxied by some `ZProxy` contract. If this contract exposed the `ZPERC20` address, which is at some point upgraded by me to fix a vulnerability, an attacker could use the previous address to "fake" a valid version of my ERC20 contract, thus exploiting the vulnerability although the contract is not "valid" anymore. I actually saw this happening with a real-life token, and was even tempted to "save" some tokens in an BNB address, before understanding that the tokens were being managed by an old version of a contract. They were still valid ERC20 tokens, but weren't accepted by the "community". The attacker was using the old contract's tokens as a honeypot. If the contract owners used the Proxy pattern, this wouldn't be a problem.

doc/contract-update.md
```diff
@@ -0,0 +1,56 @@
...
+1. Emit an event advertising the upgrade;
+
+We will ensure that a unique private key is needed to call `upgrade`.
+
+Note: The nightfall applications will need to be updated so that they can sync events and calldata on startup,
```
signorecello, line 47: This is the big thing. We use the `calldata` of the previous contracts. I think that `selfdestructing` the old contract should be enough, as no other clients would be able to use the old contract thus "faking" a valid version of NF. I think we should point this to the Security Audit guys, just to be sure that exposing the underlying proxied address and selfdestructing the old one after an update is safe

signorecello (commented):
Just some quick notes and questions! Great stuff @Westlad !

doc/contract-update.md
Westlad: We should, but I see no option here because one needs the calldata to sync back to the correct L2 state.  If for some reason that data is not available then the only option is a new deploy of the entire thing.

Westlad:
One other thing to add - we should use an ERC1967 proxy, exact implementation TBD.

Westlad:
I'll hoover up all the comments tomorrow and make a final commit, rather than do multiple ones.

doc/contract-update.md
daveroga: Discussed with @signorecello the solution to this data availability could be Polygon Avail and then in the code go for this data in Polygon Avail instead of the calldata.

signorecello: Yes we talked about this, although I'm not sure Avail will be _Avail_able (pun intended) for our mainnet launch... Not sure if we want to have that dependency 🤷

doc/contract-update.md
signorecello: I never implemented the Proxy pattern, so I really don't have a hands-on knowledge on this. I'm not referring to "our" `State` contract, but the `State` contracts the Proxy patters refer to. But I may just be incredibly confused right now 😄

Westlad: Ah well, I'm sure it will all become clear once we implement it.

doc/contract-update.md
```diff
@@ -0,0 +1,60 @@
...
+We would only migrate in the event of a key loss, which prevents upgrading contracts.
+
+## Upgrade contracts approach
+
+Our contracts are already mostly divided into contracts that contain data and contracts that
+contain logic. This is fortuitous for the pattern we are going to use.
+
+`State.sol` contains most of the state, with some in `Shield.sol`.
+This division could be made complete by moving the `proposeBlock` function out of `State.sol`
+and moving the `Shield.sol` data into `State.sol` (there may be a few other minor adjustments
```
ChaitanyaKonda, line 15: Minor adjustments would mean removing any function that has logic other than getting or setting a value, such as updateBondAccountTime, removeProposer, rewardChallenger, withdraw, emitRollback?

doc/contract-update.md
```diff
@@ -0,0 +1,60 @@
...
+TBC). `State.sol` will then only contain storage data and getters/setters for the same.
```
ChaitanyaKonda, line 16: Setters can also be considered logic and we might want to move this into logic contracts? 
For example, if a straight forward setter is in `State.sol` and a setter function in logic contract calls this `State,sol`, there are instances where this would pass even on a failed execution of setter in `State.sol`. In the event of logic contract pointing to the wrong or destroyed State.sol and if there is no check for done to see if the State contract being exists or has code, then this setting will return true even on a failed execution. One way to go around this, is to check if State.sol exists before calling any setter with in it. But it seems better coverage to have this in logic contract in case there are other considerations we haven't thought of yet.

doc/contract-update.md
```diff
@@ -0,0 +1,60 @@
...
+It will also need modifying so that the addresses of the logic contracts, which it allows to
+set state, can be updated by a registered contract containing an `upgrade` function (see later).
+
+After that is done, we will never upgrade the `State.sol` contract but will upgrade the
+stateless logic contracts. The logic contracts call `State.sol` and thus we avoid the use of
+`delegateCall`. If, in the future, `State.sol` becomes insufficient for our needs, another
+data contract can be deployed alongside, which adds the new storage variables needed.
```
ChaitanyaKonda, line 23: There might be instances where we would want to replace the current state variable other than adding new variables. For example, with the new PoS that David is implementing we replace `BondAccounts` with `stakeAccounts`. Thankfully they are of the same data types and hold similar data. So if we were to upgrade to the new PoS approach, then we would simply continue using this variable. But say if we were to make changes to PoS such that the bondAccounts variable's data type and data aren't suitable anymore, then we want to stop using the old variable and start using a new variable. Pointing the logic contracts to the new variable might suffice but sometimes it would also need data migration from old to new variable. In this example, the proposers can deregister and register with stakes again so they are reflected in the new state. A similar migration might be sufficient for proposer or challenger related variables. Even for users, the `shield.sol` only stores `withdraw,  advancedFeeWithdrawals, advancedWithdrawals, feebook` variables. So a migration by users for `withdraw,  advancedFeeWithdrawals, advancedWithdrawals` variables would mean pausing new advance withdrawals and for `feebook` would mean pausing deposits, transfers and withdraws , then waiting 7 days to confirm any remaining transactions. 

So we might want to consider adding a pause contract function?
This seems important even its not for migrating variables.

doc/contract-update.md
```diff
@@ -0,0 +1,60 @@
...
+
+We will create a set of contracts, which proxy calls to the logic contracts (one for each logic contract). This avoids users
```
ChaitanyaKonda, line 25: This question might be silly and straight forward.
How do we retain `msg.sender` for logic contracts if logic contracts are accessed via proxy contracts without using a delegate call?

doc/contract-update.md
Westlad: Agree.  This makes sense to me.  I shall remove all setters.

Westlad: Agree.  Let's remove named getters too - it seems silly to have one and not the other.  This means `State.sol` contains no functions (other than build-in getters).  We might need to have a `withdraw` function, only, to facilitate payment.

Westlad: But if the setter is not in `State.sol` how can I set a variable in `State.sol`?

doc/contract-update.md
Westlad: Yes, this makes sense. Using it would be a very bad user experience but we might one day find that it's our only recourse.

doc/contract-update.md
```diff
@@ -0,0 +1,60 @@
...
+contain the `upgrade` function (we could have a single `Proxy.sol` which deals with all contracts but this
+is less flexible if we want to add new logic contracts).
+
+![contract interaction](./contract-upgrade.png)
+
+### upgrade function
+
+This function will, atomically, carry out the following actions:
+
+1. Repoint `State.sol`'s registered logic contracts to the new logic contracts;
```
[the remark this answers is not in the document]
Westlad: So, `State.sol` will only accept calls from contracts that are registered with it: it holds their addresses.  This is to stop just anyone calling its setters, which would be bad.  Obviously, if you change the logic contract, you need to change these addresses. This is the re-pointing being referred to.

doc/contract-update.md
Westlad: Yes, although `updateBondAccountTime` should probably be considered a setter, and we need to leave `withdraw` there, because `State.sol` holds all of nightfall's money (that also being a type of state), so it's the only place that we can pay out from.

Westlad:
Closing until the upgrade work is complete - then I will add a definitive text describing how this was done.

PR #492, opened 2022-02-14
Adversary

ChaitanyaKonda:
This PR generates nigthfall adversary. Nightfall adversary is meant to be an adversarial block proposer that creates a combination of good and bad blocks. This is built for testing purposes. It is created using generative metaprogramming by copying nightfall-optimist code and injecting the code that creates bad blocks and bad transactions during block creation and transactions selection respectively. It creates blocks of type

- ValidBlock
- ValidTransaction
- IncorrectTreeRoot
- IncorrectLeafCount
- DuplicateTransaction
- DuplicateNullifier
- HistoricRootError
- IncorrectProof

**Test**
This functionality is tested by `adversary-test` job in check-PRs github action.

If the test is a success it means that a user has the correct balance considering the blocks proposed and rollbacks from bad blocks that contain their transaction. Bad block types to be generated are defined in files `test/adversary/adversary-code/database.mjs` and `test/adversary/adversary-code/block.mjs`. A challenger has spotted these bad blocks and successfully challenged them.

Removed `neg-http.mjs` because this implements the same functionality a lot cleaner

**Note** 
The nightfall adversary code is written such that it picks bad block or bad transaction type in the order as defined in `database.mjs` and `block.mjs` files. This order avoids the adversary from randomly picking 

- `DuplicateNullifier` when there are no spent transactions whose nullifiers can be duplicated
- `DuplicateTransaction` when there are no transactions to duplicate
- `IncorrectLeafCount` when there is no prior block
- `IncorrectTreeRoot` when there are no 2 prior blocks

This logic will later be replaced such that type of bad block to be generated will be picked automatically and randomly by nightfall adversary. This update depends on having a script that seeds the deployment with the first two L2 blocks and at least two transfers in the second blocks. This is being handled in the [issue](https://github.com/EYBlockchain/nightfall_3/issues/521)


ChaitanyaKonda:
Adding this test as a github action. 
This will be the proper implementation for doing the neg-http.mjs as will replace it.
Also the test need not be done manually during review.

package.json
```diff
@@ -11,7 +11,9 @@
     "test-e2e-tokens": "LOG_LEVEL=error mocha --timeout 0 --bail --exit test/e2e/tokens/*.test.mjs ",
     "lint": "eslint . --ext js,mjs,jsx && find-unused-exports",
     "prepare": "husky install",
-    "doc:build:sdk": "jsdoc -c jsdoc.json cli/lib/nf3.mjs"
+    "doc:build:sdk": "jsdoc -c jsdoc.json cli/lib/nf3.mjs",
+    "build-adversary": "node test/adversary/transpile-adversary.mjs",
+    "adversary-test": "mocha --timeout 0 --bail --exit test/adversary.test.mjs"
```
IlyasRidhuan, line 16: We should call `build-adversary` inside the `adversary-test` so that we don't have people run the test against "outdated" adversaries

test/adversary/adversary-code/block-assembler.mjs
```diff
@@ -0,0 +1,27 @@
...
+import { getMostProfitableTransactions } from './database.mjs';
+import Block from './block.mjs';
+
+const { TRANSACTIONS_PER_BLOCK } = config;
+
+let count = 0;
+async function makeBlock(proposer, number = TRANSACTIONS_PER_BLOCK) {
+  logger.debug('Block Assembler - about to make a new block');
+  // pick a random number between 0 to 6 (error length - 1)
+  const errorIndex = Math.floor(Math.random() * 0);
```
IlyasRidhuan, line 18: Not sure what `errorIndex` does here - it seems to always be 0. `count` is always 0 as well.

Does this mean that `getMostProfitableTransactions` always makes valid transactions?

ChaitanyaKonda: It's a global variable. So it is incremented with each block proposed.
Non valid blocks get created as it increments. 
This is a test run from github action
Screenshot 2022-03-07 at 09 27 57

IlyasRidhuan: ah ok - so `errorIndex` is always zero because we are not doing the random thing for now

package.json
ChaitanyaKonda: Adversary code is required at the stage of `./setup-nightfall` and `./start-nightfall`. Only after these two are completed do we run the tests. If there is no source code for adversary by `./start-nightfall `, it complains (don't remember if it fails, my memory says it does).

We could call `build-adversary` in the test. It would work if some adversary code already exists during build and start stage and this version of adversary is started. After this, `nodemon` will create the new code again during test and restart the adversary container. 

Considering we won't use `nodemon` in non dev deployments, I would recommend we get the developers used to calling `build-adversary` at the beginning of staging and prod deployments. Because they have to remember calling `./start-nightfall` with `-a`, as well as `./setup-nightfall` with `adversary` service too.

PS: We don't push `nightfall-adversary` to repo. It is part of `.gitignore` file

IlyasRidhuan (approved):

Westlad (approved):

PR #493, opened 2022-02-14
remove secrets from deploy script

Westlad:
The dev environment deploy script contains Ropsten private keys.  Although these are only for test purposes, it is better practice to create them as GitHub secrets and inject them into the script at deployment, because one day they may be real keys.

It's not possible to test fully without a deployment and so that will need to wait until we do a deploy to dev.  However it's a fairly small change so I don't anticipate problems. If it fails, we can push a fix then.


ChaitanyaKonda (approved):

ChaitanyaKonda:
Wallet tests are being removed in #490 because they are part of legacy code. 
Perhaps, restarting might solve this failed test run for now.

PR #499, opened 2022-02-15
Integrating PoS proposal: Weighted Round Robin proposers selection

### LINKED ISSUE #475: Integrating PoS proposal into PNF3: weighted round robin algorithm for proposer selection
- Implement weighted round robin algorithm based on the proposer staking for proposer selection
- changeCurrentProposer will be based on this new proposer selection
- Define first approach with a set of parameter numbers for testnet (minimum stake, proposer count, number of springs in a span, ...) that we should analyze the suitable numbers for mainnet.

daveroga:
Fix #475 

- Implement weighted round robin algorithm based on the proposer staking for proposer selection
- changeCurrentProposer will be based on this new proposer selection
- Define first approach with a set of parameter numbers for testnet (minimum stake, proposer count, number of springs in a span, ...) that we should analyze the suitable numbers for mainnet.

### DIFF
--- a/cli/lib/nf3.mjs
+++ b/cli/lib/nf3.mjs
@@ -580,7 +580,7 @@ class Nf3 {
     const res = await axios.get(`${this.optimistBaseUrl}/proposer/change`, {
       address: this.ethereumAddress,
     });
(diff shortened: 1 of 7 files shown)


ChaitanyaKonda:
Closing because a new PR to feature POS will replace it

PR #505, opened 2022-02-16
bug - submitTransaction in SDK not getting confirmations when revert

### LINKED ISSUE #504: submitTransaction in SDK never gets confirmations when a revert happens in the contract call
When calling submitTransaction in the SDK and a revert is being thown by the contract, there is no catch for the error and never get the confirmations in that case.

daveroga:
Fixes #504 

Currently, when calling submitTransaction in the SDK and a revert from the contract is thrown then confirmations are not being received. This is caused because the promise is not resolving or rejecting in that case. This PR fixes that bug by simply adding reject param and on 'error' when calling this.web3.eth.sendSignedTransactionsendSignedTransaction.

```
      return new Promise((resolve, reject) => {
        console.log(`Confirming transaction ${signed.transactionHash}`);
        this.notConfirmed++;
        this.web3.eth
          .sendSignedTransaction(signed.rawTransaction)
          .on('confirmation', (number, receipt) => {
            if (number === 12) {
              this.notConfirmed--;
              console.log(
                `Transaction ${receipt.transactionHash} has been confirmed ${number} times.`,
                `Number of unconfirmed transactions is ${this.notConfirmed}`,
              );
              resolve(receipt);
            }
          })
          .on('error', function (error) {
            reject(error);
          });
      });
```

Detected in other PR and a test has not being added in this PR as there is the PR #490  that should add a revert test tot test it.

### DIFF
--- a/cli/lib/nf3.mjs
+++ b/cli/lib/nf3.mjs
@@ -210,7 +210,7 @@ class Nf3 {
       // rather than waiting until we have a receipt, wait until we have enough confirmation blocks
       // then return the receipt.
       // TODO does this still work if there is a chain reorg or do we have to handle that?
-      return new Promise(resolve => {
+      return new Promise((resolve, reject) => {
         console.log(`Confirming transaction ${signed.transactionHash}`);
         this.notConfirmed++;
         this.web3.eth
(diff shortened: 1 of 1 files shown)


ChaitanyaKonda:
Closing this because it is covered by PR #490

PR #526, opened 2022-02-23
Westlad/ping pong fix

Westlad:
Fixes ping-pong tests after some recent PRs broke it.  Also runs an automated test on ping-pong to avoid future occurrences.

For instructions on running the ping-pong test, see `test/ping-pong/README.md`.


Westlad:
I wouldn't worry.  Tests always break eventually, unless they are part of the automated testing.

daveroga:
I see you have removed the Dockerfile inside `user-local` folder but it also could be removed the Dockerfile inside `proposer` folder as now the proposer docker file is in the root folder with `proposer.DockerFile`

Westlad:
Good point @daveroga

IlyasRidhuan (approved):

daveroga (approved):

PR #527, opened 2022-02-23
Implemented restriction for ERC20 tokens

### LINKED ISSUE #516: Restrict the value of deposits and withdraws
The objective is to minimise the loss in case of theft/loss of funds due to a bug. As long as we limit the total amount deposited/withdrawn by a user, we should be able to control the degree of loss for a user. This can be done at a contract level during deposit/withdraw for each ERC20 that we will support.

The user could accumulate more balance than the deposit threshold through receipt of funds from other users. But that is part of the design of privacy and should be restricted to an extent with deposit and withdraw thresholds. 

Also, the user could call deposit multiple times to accumulate more balance, in which case that is the going against the recommendations. The wallet’s UI should convey during deposit/withdraw this restriction on value and the reasoning that this is suggested to the user for the security of their funds during this restricted beta phase.

Definition of Done:
- Configurable threshold amount to deposit and withdraw for each ERC20 that will enforce a user to deposit/withdraw within these thresholds
- Enforce this in the wallet during deposits and withdraws
- Warn users in wallet UI about the thresholds with reasoning during deposits and withdraws

signorecello:
Closes #516 

Mind that most changes to `.sol` files are auto-formats by prettier. Major change is just the adding of `super.getRestrictions` [here](https://github.com/EYBlockchain/nightfall_3/blob/1e4bbcf164fb0e0524d55e3fbb79c2062f861290/nightfall-deployer/contracts/Shield.sol#L277) and [here](https://github.com/EYBlockchain/nightfall_3/blob/1e4bbcf164fb0e0524d55e3fbb79c2062f861290/nightfall-deployer/contracts/Shield.sol#L244)


daveroga (approved):

nightfall-deployer/contracts/Shield.sol
```diff
@@ -18,225 +18,280 @@ import './Config.sol';
...
+                uint256(t.value),
+                ''
+            );
+        } else {
+            revert('Invalid Token Type');
+        }
+    }
+
+    function payIn(Transaction memory t) internal {
+        address addr = address(uint160(uint256(t.ercAddress)));
```
ChaitanyaKonda, line 272: This is a comment for future reference, no action required

The following code 
```
ERCInterface tokenContract = ERCInterface(
    address(uint160(uint256(t.ercAddress)))
);
```
has been replaced by this code
```
address addr = address(uint160(uint256(t.ercAddress)));
ERCInterface tokenContract = ERCInterface(addr);
```

Contracts have been refactored in a prior PR to make them gas efficient to reduce the cost of transaction for user.

This change shouldn't cause any gas changes because it is using the stack location in both scenarios. 
Similarly, in the future gas efficiency should be prioritised while making any changes to the contract.

ChaitanyaKonda (changes requested):
The code looks good pending a small change to remove some code.

ChaitanyaKonda (approved):

PR #531, opened 2022-02-24
Automatically create first 2 blocks

### LINKED ISSUE #521: Automatically create first two blocks as valid blocks filled with valid transactions
Solving the issue #340 

The easiest way to solve the problem of not being able to challenge the first two blocks is to create the first two blocks as part of the initial deployment as valid blocks holding valid transactions. Every optimist instance can verify this when they join the network because they will pauses block creation if a bad block is identified.

## Definition of done
- automatically create transactions sufficient for the first two blocks which should be valid blocks filled with valid transactions 
- this will run as soon as the deployment of Nightfall deployer and proposer user is completed
- this script should be ephermeral lasting until these transactions are successfully proposed in 2 blocks
- should work with dev/staging/prod environments and run on ganache/ropsten
- create atleast 2 transfers amongst deposit (helps in testing adversarial proposer which will create different types of bad randomly)
- should work with browser based deployment only or client based deployment only or both

## Resolution
We can use `user-local` from ping-pong test wtih some modifications in code and parameters to get the first 2 blocks.
Also we need to run a container `client` to be able to do the calls through the SDK.

For mainnet there is a mock ERC20 called EY OpsCoin.
- https://etherscan.io/address/0x09b2d8b8741538abf56f47be76e37aed31f00e0d#code
- It is an ERC20 where anyone can mint unlimited number of EY OpsCoin tokens by calling mint function as can be seen in the contract tab of the Etherscan page.
- The deployer, can use this contract address. They will mint some amount of ERC20 by calling this public mint function and later using it for deposits in PNF.

daveroga:
Fixes #521 

The easiest way to solve the problem of not being able to challenge the first two blocks is to create the first two blocks as part of the initial deployment as valid blocks holding valid transactions. Every optimist instance can verify this when they join the network because they will pauses block creation if a bad block is identified.

## Definition of done
- automatically create transactions sufficient for the first two blocks which should be valid blocks filled with valid transactions
this will run as soon as the deployment of Nightfall deployer and proposer user is completed
this script should be ephermeral lasting until these transactions are successfully proposed in 2 blocks
should work with dev/staging/prod environments and run on ganache/ropsten
- create at least 2 transfers amongst deposit (helps in testing adversarial proposer which will create different types of bad randomly)
- should work with browser based deployment only or client based deployment only or both

## Resolution
We have used user-local from ping-pong test wtih some modifications in code and parameters to get the first 2 blocks.
Also we need to run a container client to be able to do the calls through the SDK.

Now `deployer-first-blocks` and `deployer-proposer` created in `./start-nightfall` after `deployer` exits and once they have created the first two blocks they exit.

For mainnet there is a mock ERC20 called EY OpsCoin.

https://etherscan.io/address/0x09b2d8b8741538abf56f47be76e37aed31f00e0d#code
It is an ERC20 where anyone can mint unlimited number of EY OpsCoin tokens by calling mint function as can be seen in the contract tab of the Etherscan page.
The deployer, can use this contract address. They will mint some amount of ERC20 by calling this public mint function and later using it for deposits in PNF.

### Start-Nightfall test
To test in general with start-nightfall in local `./start-nightfall -g -d`.
`genesis-blocks-deployer` and `proposer` created in `./start-nightfall`, they wait for contracts to be deployed and once they have created the first two blocks they exit.


ChaitanyaKonda (changes requested):
This PR creates the first two blocks under a new deployment which works when `./genesis-nightfall` is called. But this fix is required for every deployment so it should be created run when `./start-nightfall` is called. 

Once deployer finishes it's tasks and exit, this genesis deployer (renamed from user-local to make it easier to understand that this is a deployment step and not an application step) will run and exit after it is done.

deployer-first-blocks.Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM node:14.17
```
ChaitanyaKonda, line 1: To be consistent with the rest of the code, could we name this file `genesis-blocks-deployer` please?

genesis-blocks/docker-entrypoint.sh
```diff
@@ -0,0 +1,16 @@
+#! /bin/bash
+set -o errexit
+set -o pipefail
+
+# wait until there's a optimist instance up
+while ! nc -z ${CLIENT_HOST:-client} ${CLIENT_PORT:-80}; do sleep 3; done
+while true; do
```
ChaitanyaKonda, line 7: I might be missing something, why did we not use `nc` for this as below
`while ! nc -z ${DEPLOYER_HOST:-client} ${DEPLOYER_PORT:-80}; do sleep 3; done`

deployer-first-blocks.Dockerfile
daveroga: Done

docker-compose.yml
```diff
@@ -164,15 +164,44 @@ services:
...
+      - nightfall_network
+    environment:
+      OPTIMIST_HOST: optimist1
+      OPTIMIST_WS_PORT: 8080
+      OPTIMIST_HTTP_PORT: 80
+      BLOCKCHAIN_WS_HOST: blockchain1
+      BLOCKCHAIN_PORT: 8546
+      CLIENT_HOST: client1
+      DEPLOYER_HOST: deployer
+      TEST_BLOCKS: 2
```
ChaitanyaKonda, line 181: Can we call them `GENESIS_BLOCKS` because this is not just for test purposes?

docker-compose.yml
```diff
@@ -164,15 +164,44 @@ services:
...
+      OPTIMIST_WS_PORT: 8080
+      OPTIMIST_HTTP_PORT: 80
+      CLIENT_HOST: client1
+      DEPLOYER_HOST: deployer
+      CLIENT_PORT: 80
+      BLOCKCHAIN_WS_HOST: blockchain1
+      BLOCKCHAIN_PORT: 8546
+      IS_TEST_RUNNER: 'yes'
+      TEST_LENGTH: 1
+      TRANSACTIONS_PER_BLOCK: ${TRANSACTIONS_PER_BLOCK:-2}
```
ChaitanyaKonda, line 198: This is received from config to the user. Is there a reason why we added here?

deployer-proposer.Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+RUN npm ci
+WORKDIR /app/cli
+RUN npm ci
+
+WORKDIR /app/test/ping-pong/proposer
+RUN apt-get update -y
+RUN apt-get install -y netcat-openbsd
+COPY test/ping-pong/proposer/package*.json ./
+COPY test/ping-pong/proposer/src src
+COPY genesis-blocks/docker-entrypoint.sh docker-entrypoint.sh
```
ChaitanyaKonda, line 18: Similar to proposer in ping-pong where proposer is only waiting for optimist to start, this will also do the same. 
It does not need to wait for client and deployer to start because proposer will work independent of client.

In fact, we can reuse `proposer.Dockerfile`. This is not needed.

genesis-blocks-deployer.Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM node:14.17
```
ChaitanyaKonda, line 1: Same here too.
We can reuse `user-local.Dockerfile`

genesis-blocks/docker-compose.host.docker.internal.yml
```diff
@@ -0,0 +1,55 @@
+version: '3.5'
```
ChaitanyaKonda, line 1: The new `proposer` and `genesis-blocks-deployer` will be part of 
- docker-compose.yml
- docker-compose.images.example.yml
- docker-compose.dev.yml

This genesis blocks deployment is needed for all our current deployments.

So we don't need new docker-compose files in genesis-blocks.

genesis-blocks/ganache-standalone
```diff
@@ -0,0 +1,39 @@
+#! /bin/bash
+# This script will start a private gth blockchain with two miners and two node.
```
ChaitanyaKonda, line 2: I refer to my old comment here.

https://github.com/EYBlockchain/nightfall_3/pull/531#pullrequestreview-897095300

This needs to work with `./start-nightfall`. 
So we don't need these new scripts 
- `genesis-blocks/ganache-standalone`
- `genesis-blocks/genesis-down`
- `genesis-blocks/genesis-nightfall`

ChaitanyaKonda (changes requested):

druiz0992 (commented):
In my opinion this PR is way more complicated that what it needs to be. My impression is that we could simply modify the existing ping-pong test to include this functionality. I would only add the ERC20_ADDRESS option and the changes to adapt to any preconfigured number of transactions per block to the current ping pong test. The rest is duplicated code. I don’t want this comment to be enforced. Just wanted to share my opinion, and my fears that we are increasing the code base unnecessarily and the future effort to maintain code.

daveroga:
> In my opinion this PR is way more complicated that what it needs to be. My impression is that we could simply modify the existing ping-pong test to include this functionality. I would only add the ERC20_ADDRESS option and the changes to adapt to any preconfigured number of transactions per block to the current ping pong test. The rest is duplicated code. I don’t want this comment to be enforced. Ping pong test also will benefit from these changes, as it will be more robust by adapting to different number of transactions per block and the ability to use a preexisting erc20 token. I just wanted to share my opinion, and my fears that we are increasing the code base unnecessarily and the future effort to maintain code.

I totally agree and I'm going to fix all to keep it simple only with the ping pong test requirements and the minimum dockerfiles.

docker-compose.yml
```diff
@@ -164,15 +164,44 @@ services:
...
+    image: ghcr.io/eyblockchain/nightfall3-deployer-proposer:latest
+    networks:
+      - nightfall_network
+    environment:
+      OPTIMIST_HOST: optimist1
+      OPTIMIST_WS_PORT: 8080
+      OPTIMIST_HTTP_PORT: 80
+      BLOCKCHAIN_WS_HOST: blockchain1
+      BLOCKCHAIN_PORT: 8546
+      CLIENT_HOST: client1
```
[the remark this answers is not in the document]
ChaitanyaKonda: This is not resolved.
We do not need `CLIENT_HOST` here

test/ping-pong/user-local/docker-entrypoint.sh
```diff
@@ -1,6 +1,16 @@
 #! /bin/bash
 set -o errexit
 set -o pipefail
+DEPLOYER_HOST=nightfall_3_deployer_1
 # wait until there's a optimist instance up
 while ! nc -z ${CLIENT_HOST:-client} ${CLIENT_PORT:-80}; do sleep 3; done
+while true; do 
+  if ping -c 1 -W 1 "${DEPLOYER_HOST:-deployer}" > /dev/null 2>&1; then
```
ChaitanyaKonda, line 8: Ping pong test did not need to check for deployment because user-local was only started after deployer is done manually. In this case, because user and deployer are started with the same command, I can see why you are checking if deployer is done before starting user-local.

But I am not sure I understand the check.
`ping -c 1 -W 1 "${DEPLOYER_HOST:-deployer}" > /dev/null 2>&1` would not print error messsages to STDOUT or STDERR. But not sure what the if condition is checking here. If the output of ping is correctly redirected to /dev/null

test/ping-pong/user-local/src/index.mjs
```diff
@@ -41,8 +41,7 @@ async function localTest() {
   await nf3.init(zkpMnemonic);
   if (await nf3.healthcheck('client')) logger.info('Healthcheck passed');
   else throw new Error('Healthcheck failed');
-
-  const ercAddress = await nf3.getContractAddress(ERC20_NAME);
+  const ercAddress = ERC20_ADDRESS || (await nf3.getContractAddress(ERC20_NAME));
```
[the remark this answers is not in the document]
ChaitanyaKonda: Why is this resolved if nothing was removed or a comment was not provided about why it needs not be resolved?

docker-compose.yml
ChaitanyaKonda: Why is this resolved if nothing was removed or a comment was not provided about why it needs not be resolved?

daveroga:
Weird issue I'am not able to reply some comments directly in github.
I'll put the answer here:

- Regarding ERC20_ADDRESS is needed in testnet and production as we won't use the ERC20Mock and we will use an existing ERC20 passing this ERC20_ADDRESS.
- We do not need CLIENT_HOST. Removed.
- `ping -c 1 -W 1 "${DEPLOYER_HOST:-deployer}" > /dev/null 2>&1` from answers about hot to ping in linux for host (https://serverfault.com/questions/42021/how-to-ping-in-linux-until-host-is-known) as it has no ports exposed the `deployer`

config/default.js
```diff
@@ -142,7 +142,7 @@ module.exports = {
         : 'ws://localhost:8082',
       proposerBaseUrl: process.env.PROPOSER_HOST
         ? `http://${process.env.PROPOSER_HOST}:${process.env.PROPOSER_PORT}`
-        : 'http://172.16.238.1:8100',
```
druiz0992, line 145: Why did you remove this? How can i access the proposer from outside a container?

druiz0992 (approved):

druiz0992:
Now that we have mainnet deployment, do we need this?

Westlad:
I think not.  The need for this was always a bit questionable because it's easy to manually create the first two blocks on deployment.

PR #546, opened 2022-02-28
Proposer add logging and registration check

Westlad:
This PR fixes a bug in the dev environment (or indeed any environment) whereby restarting the `proposer` application will fail if it has already been registered with `optimist`.  It does that by requesting a list of registered proposers from `optimist` and only attempting to register if its address is not on the list.

This PR also adds some logging to the nf3 class.  Specifically, opening, closing, messaging and errors are all now logged for the proposer websocket, although actual log details are limited because of the way websockets work.

Note that to add the logger, some the rather brittle common-files psuedo-packaging has been removed from `proposer` and `user-local` as this did not work well when importing both `common-files` and `cli`.  It is unnecessary in any case now the `Dockerfiles` have been moved to the repository root.

The e2e test script runners in `package.json` have been modified to disable logging when these are run, otherwise the test output is spammed with log data.

To test, run the `ping-pong` tests (see the readme in `tests/ping-pong`). You will see the new log output as the test proceeds. If you wish to test duplicate registration, the easiest way is to add `await nf3.registerProposer();` at line 24 of `index.mjs`.  This will force the proposer to attempt to register twice.  A warning will be logged and the second attempt will not proceed.


package.json
```diff
@@ -6,9 +6,9 @@
   "scripts": {
     "neg-test": "mocha --timeout 0 --bail --exit test/neg-http.mjs",
     "neg-test-ropsten": "mocha --timeout 0 --bail --exit test/neg-http.mjs",
-    "test-e2e-protocol": "mocha --timeout 0 --bail --parallel --exit test/e2e/protocol/*.test.mjs ",
+    "test-e2e-protocol": "LOG_LEVEL=error mocha --timeout 0 --bail --parallel --exit test/e2e/protocol/*.test.mjs ",
     "test-gas": "mocha --timeout 0 --bail --exit test/e2e/gas.test.mjs ",
-    "test-e2e-tokens": "mocha --timeout 0 --bail --exit test/e2e/tokens/*.test.mjs ",
+    "test-e2e-tokens": "LOG_LEVEL=error mocha --timeout 0 --bail --exit test/e2e/tokens/*.test.mjs ",
```
signorecello, line 11: Hey, currently mocha gets ENV variables from the command line, so that's why you see a bunch of

```js
if (process.env.VERBOSE) console.log("whatever")
```

I think adding this env variable in ```package.json``` will be confusing as we now have two "log levels" variables. I'd propose : 
1. either removing this LOG_LEVEL and starting the e2e process with ```LOG_LEVEL="error" npm run test-e2e-tokens", 
2. or changing ```package.json``` to ```VERBOSE=true LOG_LEVEL=error mocha etc etc etc```

I personally prefer using it the former way, because we don't need to change anything in the code if we want to change log levels, you just add or remove the env variable from ```npm``` cli command

signorecello (changes requested):
Just a shout out so we have consistent env variable locations

ChaitanyaKonda (approved):

daveroga (approved):

signorecello (approved):

Issue #513, opened 2022-02-18
Update user's e2e such that deposit, transfer and withdraw tests use the state that the user's instance will have access to


ChaitanyaKonda:
Definition of done:
After deposit, transfer and withdraw, the user's balance should reflect the expected balance as on chain or pending to be on chain.  Which depends if the transaction has been submitted by Block proposed or not. In the event of a rollback of the block that hold's user transaction, the transaction will end in the mempool of the proposers and will be submitted eventually again. Some examples for user testing can be seen in `test/ping-pong/user-local/src/index.mjs` and `test/adversary.test.mjs` in `adversary2` branch.  This check could be done differently too but should be done from the data that the user will have access to according to the architecture below.

To note:
- Checking a balance as soon as the transactions have been submitted might not always be sufficient because this transaction might be part of a rollback eventually and the balance is lost once again. In reality, the confirmation requires to wait the challenge period. For the sake of the test, we might want to shorten this much further and test the total expected balance from all the transactions at the end after waiting a good enough period. 
- The above could be handled if some logic is added in to `nightfall-client/wallet` that has a flag for the transaction in the transactions collection such as `isChallengeable` which adds `true` if the transaction is part of a bad block or there any prior bad blocks. A future enhancement issue. 
- `waitForSufficientBalance` used in `test/ping-pong/user-local/src/index.mjs` or `test/adversary.test.mjs` checks if the accumulated balance of the user is greater than or equal to the value that the user wants to transfer. But this function is not correct (incomplete) because in order to transfer, the user either requires a commitment of the exact value or exactly 2 commitments when put together hold a value greater than or equal to the value that needs to be transferred. 
![image](https://user-images.githubusercontent.com/36898048/154673272-fb2b5061-f64b-4dde-94a8-d7542a1a39b6.png)

PR #548, opened 2022-03-01
(nightfall-browser) e2e test implemented

LijuJoseJJ:


LijuJoseJJ:
currently branch `wallet/nf-browser-v2` has bug, when `optimist ` build from image `ghcr.io/eyblockchain/nightfall3-optimist:latest` seems like websocket does not properly setups
when browser initiate `sync`,  log `SUBSCRIBING TO PROPOSEDBLOCK` never appeared
Screenshot 2022-03-01 at 2 28 45 PM

this change somehow fixes it, I add this this change for debugging issue.

Note: I do not have any changes in nightfall-optimist

IlyasRidhuan:
This works for me - @LijuJoseJJ  is this suppose to replace the existing Selenium test folder in `wallet`. It is not my preference to have two different test directories (`test/` and `tests/`)

LijuJoseJJ:
> This works for me - @LijuJoseJJ is this suppose to replace the existing Selenium test folder in `wallet`. It is not my preference to have two different test directories (`test/` and `tests/`)

old Selenium e2e test code removed

wallet/tests/e2e/specs/e2e-spec.js
```diff
@@ -0,0 +1,252 @@
...
+      cy.spy(win.console, 'log');
+      cy.spy(win.console, 'error');
+      cy.spy(win.console, 'warn');
+    });
+  });
+
+  context('MetaMask', () => {
+    it('getNetwork should return network by default', () => {
+      cy.getNetwork().then(network => {
+        expect(network.networkName).to.be.equal('ganache-nightfall');
```
ChaitanyaKonda, line 26: This test should work with non ganache deployment too.
So this shouldn't be hardcoded.

wallet/tests/e2e/specs/e2e-spec.js
```diff
@@ -0,0 +1,252 @@
...
+    it('generate Mnemonic and set state', () => {
+      cy.contains('Polygon Nightfall Wallet').click();
+      cy.get('button').contains('Generate Mnemonic').click();
+      cy.get('button').contains('Create Wallet').click();
+      cy.confirmMetamaskSignatureRequest().then(confirmed => expect(confirmed).to.be.true);
+      cy.get('#TokenItem_tokenDepositMATIC').click();
+      cy.url().should('include', '/bridge');
+      cy.contains('Nightfall Assets').click();
+      cy.url().should('include', '/wallet');
+      cy.get('button').contains('Generate Mnemonic').should('not.exist');
```
ChaitanyaKonda, line 51: Not clear why how this checks if the state has been set.
Comments might help

wallet/tests/e2e/specs/e2e-spec.js
```diff
@@ -0,0 +1,252 @@
...
+      cy.get('#Bridge_modal_continueTransferButton').click();
+      cy.wait(20000);
+      cy.confirmMetamaskPermissionToSpend().then(approved => expect(approved).to.be.true);
+      cy.wait(30000);
+      cy.confirmMetamaskTransaction().then(confirmed => expect(confirmed).to.be.true);
+      cy.wait(50000);
+      cy.get('.btn-close').click();
+    });
+
+    it(`second deposit of value ${depositValue}`, () => {
```
ChaitanyaKonda, line 71: Where are we setting the deposit values in the subsequent tests?

wallet/tests/e2e/specs/e2e-spec.js
LijuJoseJJ: actually its just the environment variable check, i set here
 ```
 "e2e-test": "NETWORK_NAME=ganache-nightfall RPC_URL=http://localhost:8546 CHAIN_ID=1337 PRIVATE_KEY=0x4775af73d6dc84a0ae76f8726bda4b9ecf187c377229cb39e1afa7a18236a69e npx synpress run"
 ```
 code is in wallet/package.json

wallet/tests/e2e/specs/e2e-spec.js
```diff
@@ -0,0 +1,252 @@
...
+
+    /*
+     * dummy pkd of user who does not exist
+     * Note: even though we are passing recipientPkd but in code
+     *   it gets override by sender's pdk for now
+     *   hence two check
+     *   check1: before Block Proposed event
+     *   check2: after Block Proposed event
+     */
+    const recipientPkd = '0x90ff185f7fa35ddae731ddad18a958d55d45bb973c16735018f6bc6f3798a7e1';
```
ChaitanyaKonda, line 153: Similar test for a user using `nightfall-client` via the `SDK` is done in `test/ping-pong/user-local/src/index.mjs`. 
The approach might be slightly different in wallet tests but functionally they are the similar (minus the withdraw).
Can we replicate similar assignment to variables without hardcoding them?
Please also refer to `test/e2e/tokens/erc20.test.mjs` whilst bearing in mind the upcoming changes to these tests under (issue)[https://github.com/EYBlockchain/nightfall_3/issues/513]

wallet/tests/e2e/specs/e2e-spec.js
```diff
@@ -0,0 +1,252 @@
...
+      cy.get('button').contains('Continue').click();
+      cy.wait(50000);
+      cy.confirmMetamaskTransaction().then(confirmed => expect(confirmed).to.be.true);
+      cy.get('.btn-close').click();
+      cy.contains('L2 Bridge').click();
+      cy.contains('Nightfall Assets').click();
+    });
+
+    // check1
+    it(`check token balance after transfer - before block proposed event`, () => {
```
ChaitanyaKonda, line 168: The test should be written from the perspective of user who will not track block proposed events or know the size of L2 blocks. They can only check if their balance is pending or finalised. 
References on how to test without considering block proposed in `test/ping-pong/user-local/src/index.mjs` and `test/adversary.test.mjs`

ChaitanyaKonda (changes requested):
Comments above

wallet/tests/e2e/specs/e2e-spec.js
```diff
@@ -0,0 +1,252 @@
...
+    it('getNetwork should return network by default', () => {
+      cy.getNetwork().then(network => {
+        expect(network.networkName).to.be.equal('ganache-nightfall');
+        expect(network.networkId).to.be.equal(1337);
+      });
+    });
+
+    it('getMetamaskWalletAddress should return wallet address of current metamask account', () => {
+      cy.getMetamaskWalletAddress().then(address => {
+        expect(address).to.be.equal('0x9C8B2276D490141Ae1440Da660E470E7C0349C63');
```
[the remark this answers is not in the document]
LijuJoseJJ: same as above, 
since we are configuring metamask it need to connect to network and also need an account.

wallet/tests/e2e/specs/e2e-spec.js
```diff
@@ -0,0 +1,252 @@
...
+      cy.wait(50000);
+      cy.get('.btn-close').click();
+      cy.contains('Nightfall Assets').click();
+      cy.wait(20000);
+    });
+
+    it(`check token balance equal to ${depositValue * 4}`, () => {
+      cy.get('#TokenItem_tokenBalanceMATIC').should($div => {
+        const totalBalance = Number($div.text());
+        expect(totalBalance).to.equal(depositValue * 4);
```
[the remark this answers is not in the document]
LijuJoseJJ: changed

ChaitanyaKonda (approved):

PR #514, opened 2022-02-18
David/staging

druiz0992:
Close #543 

Minor changes for staging/production deployments. No change required or done for existing deployments. 

- `config/default.js` :  optimistBaseUrl,   optimistWsUrl and  web3WsUrl now depend on DOMAIN_NAME
- nightfall-deployer/truffle-config.js : Added staging and mainnet networks. I am using staging temporarily which uses ganache
- added /healthcheck endpoint to propsoser. I need this because in AWS deployments will bring container down if there isn't an endpoint where healthcheck can be checked
- modified `common-files/utils/mongo.mjs` to work with both mongoDb and DocumentDb. 
- DocumentDb is accessed via some stored secrets in AWS secret manager.  Code needs to be part of nightfall.

PR #549, opened 2022-03-01
David/proposer offchain

### LINKED ISSUE #524: Users should be able to send transfer and withdraw transactions directly to the current and subsequent few proposers
Definition of Done:
- During proposer registration, the proposers will also register their URL that the users can use to send direct transactions to. To save gas, the URL registered does not need to the specify the path `/proposer/offchain-transaction`.
- A proposer should be able to update the URL in the contract.
- A user should be able to send a transaction to the current proposer and the subsequent few proposers by querying the contract for this information.

Note:
Securely (rate throttled, protected against  bot attacks, NoSQL injection etc using WAF) expose this endpoint in Proposer's optimist node to the outside world

druiz0992:
Fixes #524 

This PR includes a proposer that exports two endpoints:
- GET /healthcheck
- POST /proposer/offchain-transaction to receive offchain transactions from users. The proposer will forward transactions received through this endpoint to optimist endpoint

Additional changes included are:
- When a proposer registers, it now needs to provide its API URL so that the URL is recorded in the `proposers` contract. The proposer needs to have two endpoints implemented:
   - GET /healthcheck
   -  POST /proposer/offchain-transaction
- nightfall-client/src/routes/peers have been deleted, as the functionality is not used anymore. However, there is a function in nightfall-client/src/services/peers.mjs `getProposers` that return the next N proposers and the URLs so that a user can forward transactions directly. The URLs from proposers are extracted from `proposers.sol` contract directly.
- Off chain transactions have been added to cli
- Increased gas limit in proposer tests as a result of adding a url 
- New updateProposer(url) method added to SDK to update proposer's URL
- Proposer can not be launched with two optional parameters
  - --proposerUrl : URL where API is located (http://172.16.238.1 by default, so that containers can communicate with it directly)
  - --proposerPort: Port where proposer is launched (8100 by default)

To test changes, simply run ping-pong test. If has been modified to alternate between onchain and offchain transactions. You can also test direct transactions with cli tools (launching `./nf` and requesting offchain transactions).

NOTE: I had to merge PR with #514 because there were changes to ping pong test that were required.


signorecello:
Hey the tests failed because it turns out we were really over the limit on gas. Add another 0 to the gas cost and they should run fine :sweat_smile:

config/default.js
```diff
@@ -36,9 +38,21 @@ module.exports = {
   OPTIMIST_HOST: process.env.OPTIMIST_HOST || 'optimist',
   OPTIMIST_PORT: process.env.OPTIMIST_PORT || 80,
   clientBaseUrl: `http://${process.env.CLIENT_HOST}:${process.env.CLIENT_PORT}`,
-  optimistBaseUrl: `http://${process.env.OPTIMIST_HOST}:${process.env.OPTIMIST_HTTP_PORT}`,
-  optimistWsUrl: `ws://${process.env.OPTIMIST_HOST}:${process.env.OPTIMIST_WS_PORT}`,
-  web3WsUrl: `ws://${process.env.BLOCKCHAIN_WS_HOST}:${process.env.BLOCKCHAIN_PORT}`,
+  // Define Urls and Url format (http/ws vs https/wss, with vs without port) depending on whether a DOMAIN_NAME has been defined.
+  // In production and staging environements, we require https/wss and no port, as traffic will be routed to the correct service
+  // given a URL.
+  optimistBaseUrl:
```
ChaitanyaKonda, line 44: Can we define this in the `ENVIRONMENTS` variables below where it already exists?

nightfall-deployer/contracts/Proposers.sol
```diff
@@ -29,11 +30,12 @@ contract Proposers is Stateful, Structures, Config {
 
 
   //add the proposer to the circular linked list
-  function registerProposer() external payable {
+  function registerProposer(string memory url) external payable {
```
ChaitanyaKonda, line 33: `State.sol` holds the list of all proposers and any associated details of each proposer as below
```  
mapping(address => LinkedAddress) public proposers;
```

Can we add url to `LinkedAddress` struct that holds proposer's info?
```
    struct LinkedAddress {
        address thisAddress;
        address previousAddress;
->    string url;
        address nextAddress;
    }
```

test/ping-pong/proposer/src/routes/proposer.mjs
```diff
@@ -0,0 +1,32 @@
...
+module but handle the entire request here.
+*/
+import express from 'express';
+import axios from 'axios';
+import logger from '../../../../../common-files/utils/logger.mjs';
+
+let optimistUrl = '';
+const router = express.Router();
+
+router.post('/offchain-transaction', async req => {
```
ChaitanyaKonda, line 13: Similar to how proposer interacts with optimist today via SDK, this will also be a function in SDK for the proposer to call the optimist. Since the SDK has the optimist url from nf3 class, it will be able to call this easily.

test/ping-pong/proposer/src/app.mjs
```diff
@@ -0,0 +1,32 @@
...
+  }),
+);
+
+app.get('/healthcheck', (req, res) => {
+  res.sendStatus(200);
+});
+
+app.use('/proposer', proposer.router);
+
+function setOptimistUrl(url) {
```
ChaitanyaKonda, line 28: proposer already is setting this URL when it calls 

```
new Nf3(ethereumSigningKey, {
    web3WsUrl,
    clientApiUrl,
    optimistApiUrl,
    optimistWsUrl,
  });
```

ChaitanyaKonda (changes requested):
This is implemented for`nightfall-client`.
I'm assuming you planned implement this for wallet after wallet's merge into master?

ChaitanyaKonda:
BREAKING CHANGE: Because registerProposer function signature changes

config/default.js
druiz0992: `ENVIRONMENTS` is only used in internal testing. We should not touch that.

test/ping-pong/proposer/src/app.mjs
druiz0992: My issue is that i need to access the optimist URL from a file within the proposer structure (proposer/routes/proposer.mjs) so that the proposer forwards the call to optimist. This route file doesn't have access to the URL, and that is why i am passing it down. 
If i don't do it like this, I fail to understand how to do it. Do you have a suggestion?

ChaitanyaKonda: My suggestion is to implement `sendOffchainTransaction` in SDK similar to the other proposer functions that SDK implements already `registerProposer`, `deregisterProposer`, `changeCurrentProposer`, `getCurrentProposer`, `getProposers`. 

The proposer's route file will then call this function from SDK. 
The objective of the SDK is to abstract the logic of implementation from a proposer.
The proposer will only have to think about calling the relevant SDK with configuration that they prefer by having configureable variables. One example I use to explain it easily is how a user uses geth JSON RPC to configure their block chain node without implementing any functional logic for the geth node. In our case, it will be SDK based REST API. If the SDK is accessed from route file or proposer.mjs file, it is still the proposer that is using it.

signorecello (approved):

ChaitanyaKonda (approved):

PR #552, opened 2022-03-01
updateable contracts part 1

Westlad:
DO NOT RELY ON contract-update.md for reviewing this PR. The pattern used here is quite different.  Instead, read the Openzeppelin Upgades link below.  We'll update the documentation when these changes are finalised.

This PR deploys proxied contracts that can be upgraded.  To keep PRs small, this is the first of (probably) two PRs. This PR simply adds upgradable contracts, it does not provide any means to carry out an upgrade; that will follow in a second PR.

The PR leans heavily on Openzepplin's [Upgrades](https://docs.openzeppelin.com/upgrades-plugins/1.x/) library. Nevertheless, some changes are required to make contracts upgradeable.  This is mainly:
- proxies are oblivious to constructors and thus constructors are replaced by `initialize()` functions (note the US spelling or expect hard to find bugs). The `initializer` modifier ensures they can only be called once.
- `.transfer` no longer works because it has heavily restricted gas to prevent reentrancy.  We replace it with `.call`. This increases the risk of reentrancy attacks and so functions use a checks-effects-interactions pattern normally, but all are additionally protected by the Openzepplin `RentrancyGuard` modifier which prevents a modified function being called reenetrantly in any case.

Testing is as normal but you will need to run `./setup-nightfall` to incorporate changes to the packages used. Note that the contracts are actually being called transparently through their proxies in the tests.


test/e2e/protocol/proposer.test.mjs
```diff
@@ -26,14 +26,19 @@ const web3Client = new Web3Client();
 describe('Basic Proposer tests', () => {
   before(async () => {
     for (const prop of testProposers) {
+      console.log('INITIALISING');
```
signorecello, line 29: just a shout out to remove these ```console.logs``` :smile:

signorecello (changes requested):
Great stuff! I would just remove the ```console.logs``` used for debugging, though

test/e2e/protocol/proposer.test.mjs
Westlad: Oh yes, sorry, I forgot those.

signorecello (approved):

nightfall-deployer/migrations/4_upgrade_upgradeable.js
```diff
@@ -0,0 +1,32 @@
...
+  await deployer.link(Verifier, [Challenges,ChallengesUtil]);
+  await deployer.deploy(MiMC);
+  await deployer.link(MiMC, MerkleTree_Stateless);
+  await deployer.deploy(MerkleTree_Stateless);
+  await deployer.link(MerkleTree_Stateless, [Challenges,ChallengesUtil]);
+    await deployer.deploy(Utils);
+  await deployer.link(Utils, [Shield,Challenges,ChallengesUtil]);
+  await deployer.deploy(ChallengesUtil);
+  await deployer.link(ChallengesUtil, Challenges);
+  await deployProxy(Structures, [], {deployer, unsafeAllowLinkedLibraries: true });
```
ChaitanyaKonda, line 26: This is the same file as `nightfall-deployer/migrations/2_deploy_upgradeable.js`.
I am assuming this is WIP that will be tied up in upgrade PR 2, which will use `upgradeProxy` instead of `deployProxy` here?

Westlad: All of these are libraries, i.e. they contain only pure functions and they're only called by other contracts. Thus there didn't seem to be any point in proxying them because we can just replace them if we need to, and link the new versions to the more stateful contracts.
I may change my mind but that's my current thinking.

doc/contract-update.md
```diff
@@ -0,0 +1,68 @@
+# Upgrading and Migrating contracts
```
ChaitanyaKonda, line 1: Can we remove this please?
Will be confusing because we are implementing the delegateCall approach now

ChaitanyaKonda:
This PR looks good. 

To understand what I should review, I relied on the plugin documentation link you provided.
I presume the following changes as suggested in [upgrade plugin](https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable) will come in the next PR as suggested in your description.

> The PR leans heavily on Openzepplin's [Upgrades](https://docs.openzeppelin.com/upgrades-plugins/1.x/) library. Nevertheless, some changes are required to make contracts upgradeable.

- `Config.sol` defining initial values for field declarations 
- `Stateful.sol` to initialise `state.sol` inside `initialize()`
- Adding `Upgradeable` suffix to contract names
- `Utils.sol` defining initial values for field declarations

doc/contract-update.md
ChaitanyaKonda: Can be handled in PR2

ChaitanyaKonda (approved):

Issue #478, opened 2022-02-08
Proposer to propose multiple L2 blocks at once


ChaitanyaKonda:
#426 adds confirmations for processing L2 events. This will slow block assembly. We should be able to modify [conditionalMakeBlock](https://github.com/EYBlockchain/nightfall_3/blob/aede5aa7865faf58c53589ee3733bef0766dc24a/nightfall-optimist/src/services/block-assembler.mjs#L43) to create `numberOfUnprocessedTransactions / TRANSACTIONS_PER_BLOCK` blocks instead of just `1`. This should allow us to build multiple blocks before having to process the events with their 12 confirmation wait in 0th queue.

ChaitanyaKonda:
Once the current proposer changes, the new current proposer can submit multiple blocks at once.

- If there are not enough unprocessed transactions for more than 1 block: it is similar to proposing 1 block at a time -> for each L2 block proposal, it will have to wait for prior events in the queue to be processed with their 12 confirmation wait included.
- If there are too many unprocessed transactions: an upper limit for maximum number of L2 blocks to propose is required such that the proposer does not lose gas on all the L2 blocks that did not make it to L1 when they were the current proposer. `MAX_L2_BLOCKS_TO_PROPOSE = Number of L1 blocks a proposer gets * the number of L2 blocks can fit in each L1`. This assumes that L1 blocks are completely filled with NF's L2 block proposed transactions, this might need to be corrected by a factor

PR #553, opened 2022-03-01
Proposer to propose multiple L2 blocks at once

daveroga:
Feature #478

## Feature
Once the current proposer changes, the new current proposer can submit multiple blocks at once.

- If there are not enough unprocessed transactions for more than 1 block: it is similar to proposing 1 block at a time -> for each L2 block proposal, it will have to wait for prior events in the queue to be processed with their 12 confirmation wait included.
- If there are too many unprocessed transactions: an upper limit for maximum number of L2 blocks to propose is required such that the proposer does not lose gas on all the L2 blocks that did not make it to L1 when they were the current proposer. `MAX_L2_BLOCKS_TO_PROPOSE = Number of L1 blocks a proposer gets * the number of L2 blocks can fit in each L1`. This assumes that L1 blocks are completely filled with NF's L2 block proposed transactions, this might need to be corrected by a factor

## Actions
[conditionalMakeBlock](https://github.com/EYBlockchain/nightfall_3/blob/aede5aa7865faf58c53589ee3733bef0766dc24a/nightfall-optimist/src/services/block-assembler.mjs#L43) to create `numberOfUnprocessedTransactions / TRANSACTIONS_PER_BLOCK` blocks instead of just 1. This should allow us to build multiple blocks before having to process the events with their 12 confirmation wait in 0th queue.

- Remove async from `Block.build` and pass the parameters needed instead.
- `makeBlock` to return list of blocks and transactions per block instead of single block.
- ws.send 'block' will send list of unsigned propose block transactions so that the proposer can send the list of unsigned tx to the state contract with `proposeBlock`.

## Test
Created a test for this functionality. A user will create some deposits without a proposer and then a proposer will register and start to get all pending L2 blocks proposed txs to be proposed at once.
- Start Nightfall with `./start-nightfall -g -s -d`
- Run the test `npm run test-e2e-protocol` that includes `proposer should propose multiple L2 blocks after deposits` test.


test/e2e/multiple-proposed-blocks.test.mjs
```diff
@@ -0,0 +1,132 @@
+/* eslint-disable no-await-in-loop */
+import chai from 'chai';
```
IlyasRidhuan, line 2: It's not clear to me why we have a specific test for this? This type of behaviour should just be testable and (in theory) already covered as part of the standard test suite.

IlyasRidhuan (changes requested):

cli/lib/nf3.mjs
```diff
@@ -661,15 +661,30 @@ class Nf3 {
...
+      const { type, txDataToSignList } = msg;
       logger.debug(`Proposer received websocket message of type ${type}`);
       if (type === 'block') {
-        const res = await this.submitTransaction(
-          txDataToSign,
-          this.stateContractAddress,
-          this.BLOCK_STAKE,
-        );
-        newGasBlockEmitter.emit('gascost', res.gasUsed);
+        if (process.env.VERBOSE) console.log(`Found ${txDataToSignList.length} blocks to process`);
```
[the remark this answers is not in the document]
daveroga: You are right. Now that is available logger I will change VERBOSE for logger.

cli/lib/nf3.mjs
```diff
@@ -661,15 +661,30 @@ class Nf3 {
...
+            txDataToSignList[i],
+            this.stateContractAddress,
+            this.BLOCK_STAKE,
+          );
+          submitTxList.push(submitTx);
+        }
+
+        try {
+          // wait for all tx to be confirmed (12 confirmations)
+          const resultSubmitTxs = await Promise.all(submitTxList);
```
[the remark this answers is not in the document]
daveroga: The txs has been sent previously in order. This `Promise.all` is to wait for 12 confirmations in each of them.

cli/lib/nf3.mjs
```diff
@@ -194,9 +192,10 @@ class Nf3 {
     // we need a Mutex so that we don't get a nonce-updating race.
 
     let tx;
+    let nonce;
     await this.nonceMutex.runExclusive(async () => {
-      this.nonce = await this.web3.eth.getTransactionCount(this.ethereumAddress);
-
+      nonce = await this.web3.eth.getTransactionCount(this.ethereumAddress);
+      nonce += this.txsNotConfirmed; // we need to add the number of transactions not confirmed yet
```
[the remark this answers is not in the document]
daveroga: In a previous PR it was introduced the change of `this.web3.eth.getTransactionCount(this.ethereumAddress)` for `this.nonce`every time we submit a transaction. So `this.nonce` then has no sense as it always is being set to `this.web3.eth.getTransactionCount(this.ethereumAddress)`. That's why I removed this from the class and put it as local variable. If you look at master branch you will see it. I think this was introduced because errors when revert from contract, etc. This `this.nonce` was not decremented and produced nonce errors because nonce account was different from `this.nonce` that was previously incremented.

nightfall-optimist/src/classes/block.mjs
```diff
@@ -124,16 +121,19 @@ class Block {
...
-      transactionHashes: transactions.map(t => t.transactionHash),
-      leafCount: timber.leafCount,
-      root: updatedTimber.root,
-      blockHash,
-      nCommitments,
-      blockNumberL2,
-      previousBlockHash,
-    });
+    return {
+      block: new Block({
```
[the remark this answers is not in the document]
daveroga: I return the `block` because now there is another parameter `updatedTimber` that is the updated timber in order to generate subsequent block proposed transactions until the unprocessed txs, that is needed in the `makeBlock` function that does the `Block.build` for every block.

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -23,15 +25,37 @@ export function setBlockAssembledWebSocketConnection(_ws) {
   ws = _ws;
 }
 
-async function makeBlock(proposer, number = TRANSACTIONS_PER_BLOCK) {
+async function makeBlock(proposer, unprocessed, number = TRANSACTIONS_PER_BLOCK) {
```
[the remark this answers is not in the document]
daveroga: I think `number` was the transactions we want to build in the block. It's different that unprocessed transactions, that is the total transactions in the mempool that are not processed yet. So `unprocessed / number` makes the number of blocks the proposer can build for all this unprocessed transactions at once.

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -23,15 +25,37 @@ export function setBlockAssembledWebSocketConnection(_ws) {
...
+  let { blockNumberL2, blockHash } = await getLatestBlockInfo();
+  let latestTree = await getLatestTree();
+  const blockList = [];
+  const transactionsList = [];
+
   // then we make new block objects until we run out of unprocessed
   // transactions
-  const block = await Block.build({ proposer, transactions });
-  return { block, transactions };
+  for (let i = 0; i < unprocessedTransactions.length / number; i++) {
```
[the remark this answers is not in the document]
daveroga: We can rid of this `number` parameter that was in this function previously to this PR as it's not being used, and use directly TRANSACTIONS_PER_BLOCK. It will be more clear.

test/e2e/multiple-proposed-blocks.test.mjs
daveroga: Yes, I was testing this independently before but it could be added now as part of the standard test suite.

signorecello (approved):
lgtm

cli/lib/nf3.mjs
```diff
@@ -212,6 +213,7 @@ class Nf3 {
         nonce: this.nonce,
       };
       this.nonce++;
+      this.txsNotConfirmed++;
```
ChaitanyaKonda, line 216: Why did we replace `notConfirmed` with `txsNotConfirmed` with the additional increment here?

ChaitanyaKonda: Incrementing `txsNotConfirmed/notConfirmed` in the function's global scope on top of incrementing it inside the `if` condition later means that we will always print that there is an unconfirmed transaction even when there is none. Because we incrementing twice but decrementing only once. 

To dig deeper, if the function successfully executes `if (this.ethereumSigningKey) { }` condition, then we will return before executing `this.txsNotConfirmed--;` in line `240`. So `Number of unconfirmed transactions is ${this.notConfirmed}` is getting printed with wrong value.
 
To explain what the function is meant to do:
`if (this.ethereumSigningKey) { }` is used below to differentiate if the user is using a locked account or an unlocked account. Duncan added this logic to be backwards compatible with David's old wallet code. Today we won't be running into a situation where line 248 with `return this.web3.eth.sendTransaction(tx);` is reached because we don't use unlocked accounts. So Duncan did not implement lines 224 - 244 for an unlocked account too because they don't get executed. 

Can we revert this function's code to the old code please?


or both scenarions - to proposer or to blockchain. This is wrong because `Number of unconfirmed transactions is ${this.notConfirmed}` 
Could we revert this file to older stage and make no changes here?
The logic is correct

cli/lib/nf3.mjs
```diff
@@ -661,15 +665,20 @@ class Nf3 {
...
       if (type === 'block') {
-        const res = await this.submitTransaction(
-          txDataToSign,
-          this.stateContractAddress,
-          this.BLOCK_STAKE,
-        );
-        newGasBlockEmitter.emit('gascost', res.gasUsed);
+        logger.debug(`Found ${txDataToSignList.length} blocks to process`);
+
+        txDataToSignList.reduce((seq, txDataToSign) => {
```
ChaitanyaKonda, line 673: Is the `reduce` method used here to submit transactions sequentially and not to reduce an array to a value?
If so, this is new to me :) I usually use `for..of` loop to do this.

daveroga: Yes, reduce pattern. I learned it from @IlyasRidhuan suggestion and after reviewing some information.

IlyasRidhuan: Yes the reduce pattern is preferred if we need to sequentially resolve promise. It avoids having to prevent the `eslint` warning about `awaiting in a for loop`

ChaitanyaKonda: `for await...of` can be used for synchronous iteration for an async/sync iterable objects

signorecello: I confirm `for... of` also allows synchronous iteration, didn't know about the `reduce` pattern

test/e2e/protocol/proposer.test.mjs
```diff
@@ -29,10 +38,40 @@ const testProposersUrl = [
...
 describe('Basic Proposer tests', () => {
   before(async () => {
+    await nf3User.init(mnemonics.user1);
+    erc20Address = await nf3User.getContractAddress('ERC20Mock');
+    stateAddress = await nf3User.stateContractAddress;
+    web3Client.subscribeTo('logs', eventLogs, { address: stateAddress });
+  });
+
+  it('proposer should propose multiple L2 blocks after deposits', async function () {
+    let currentPkdBalance;
```
druiz0992, line 58: How does this test verify that multiple blocks are generated?

ChaitanyaKonda: Same question too, the rest looks good

daveroga:
During the tests I checked the message sent by `logger.debug('Found ${txDataToSignList.length} blocks to process');` but it's true that I have to pass some check directly to the text and check it there. I've updated the test sending another parameter of the number of blocks proposed in the 'gasCost' event.

test/e2e/protocol/proposer.test.mjs
ChaitanyaKonda: > During the tests I checked the message sent by `logger.debug('Found ${txDataToSignList.length} blocks to process');` but it's true that I have to pass some check directly to the test and check it there. I've updated the test sending another parameter with the number of blocks proposed in the 'gasCost' event.
> 
> `expect(blocksReceivedToPropose).to.be.equal(Math.floor(totalDeposits / txPerBlock));`

How is `blocksReceivedToPropose` defined ?

ChaitanyaKonda:
> During the tests I checked the message sent by `logger.debug('Found ${txDataToSignList.length} blocks to process');` but it's true that I have to pass some check directly to the test and check it there. I've updated the test sending another parameter with the number of blocks proposed in the 'gasCost' event.
> 
> `expect(blocksReceivedToPropose).to.be.equal(Math.floor(totalDeposits / txPerBlock));`

How is `blocksReceivedToPropose` defined ?

daveroga:
> 
It's the `txDataToSignList.length`. Length of the list of `txDataToSign`. Data to sign of blocks to propose that the proposer receive with the data of type 'block' from the optimist.

PR #564, opened 2022-03-04
Westlad/fix restart nightfall

### LINKED ISSUE #375: Ping-pong test is not idempotent
It should be possible to re-run the ping-pong test without incident and yet it causes a duplicate keys error.  Currently, this means that a re-deployment is required to re-run the test.

Westlad:
fixes #375
When nightfall is stopped and restarted, its containers should re-sync with the blockchain and it should carry on working.  This was buggy for two reasons, which this PR fixes:

1) If `optimist` stops and loses its database, it does not know which proposers are registered with it. Therefore it will not make blocks for them when they are the current proposer.  The proposers are still registered with the blockchain, however, and so should be able to make blocks when it's their turn in the round-robin.  This PR adds a check to the `/register` endpoint.  It checks whether the proposer is (a) already registered with the blockchain and (b) already registered with the `optimist` instance.  It then registers on an as-needed basis.  This is transparent to the caller.
2) `client` did not sync if its commitment database had been dropped (or a new instance was started).  This meant that its Timber Merkle tree was out of sync with the blockchain and it was unable to perform `transfer` L2 transactions.

One result of the above fix is that the ping-pong test can now be run repeatedly without resetting nightfall.

One can demonstrate the effectiveness with a ping-pong test:
- Start and run the ping pong test as described in the readme.
- When the test has finished bring the containers down: `./pong-down` or `docker-compose down` (the former will drop `client`'s DB, the latter will not).  *Do not* do `./pong-down -v` because this will remove the trusted setup and smart contract addresses., effectively starting again with a clean environment.
- Restart nightfall with either `./pong-nightfall -s` or `./pong-nightfall -d -s`.  The former will run up nightfall without running `deployer` first, thus guaranteeing no changes to the blockchain or trusted setup; the second will run `deployer` but deployer will work out that it has no work to do and will eventually exit without making changes. They are therefore equivalent.
- Wait for nightfall `optimist` and `client` to finish syncing with the blockchain (maybe we should provide a `syncing` endpoint so we can test for this programmatically).
- Re-run `./pong-apps`.  The test will succeed, whereas previously it would have failed.


cli/lib/nf3.mjs
```diff
@@ -570,6 +570,25 @@ class Nf3 {
...
+  if the proposer is already registered on the blockchain (has paid their bond) and
+  for some reason the Optimist instance does not know about them, e.g. a new instance
+  has been created. The method 'registerProposer' will both register the proposer
+  with the blockchain and register locally with the optimist instance. So, if
+  that method has been used successfully, there is no need to also call this method
+  @method
+  @async
+  @returns {Promise} A promise that resolves to the Ethereum transaction receipt.
+  */
+  async registerProposerLocally() {
```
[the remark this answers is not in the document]
Westlad: It's for consistency of return.  When we call `registerProposer` we're expecting the proposer receive, sign and submit an Ethereum transaction, or fail. When we call `registerProposerLocally, we're expecting a simple success or fail report.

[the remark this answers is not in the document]
signorecello: I see... Can I suggest we instead return the raw transaction if it needs signing, and nothing (or a boolean or something) if it's registered locally? Then we could just make something like:

```js
// pseudo-code obviously
const { rawTx } = await nf3proposer.registerProposer()
if (rawTx) web3.submitTransaction(rawTx)
```

I really don't like that the developer needs to know the internals of optimist to understand why `registerProposer` reverts, and why there is a `registerProposerLocally`. I'd make this as abstracted as possible...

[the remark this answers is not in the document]
Westlad: I'm happy to do that if it makes more sense to you.  So you'd have a single endpoint and logic to process the return, i.e.:
A transaction for a first time registration
A `true` for a successful local registration
A `false` or `Error` for an attempted but un-needed re-registration.
This would be a breaking change but if it makes more sense to you like that then it probably makes more sense to everyone else.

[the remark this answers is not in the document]
signorecello: I can't speak for everyone else, so let's discuss this offline

cli/lib/nf3.mjs
```diff
@@ -563,13 +563,33 @@ class Nf3 {
...
+    return axios.post(`${this.optimistBaseUrl}/proposer/registerlocally`, {
```
druiz0992, line 588: Proposer registration has changed so that now it includes a URL. Does `registerLocally` need to pass the URL as well? Probably not needed as this info is stored in the smart contract

nightfall-client/src/services/database.mjs
```diff
@@ -109,7 +109,8 @@ export async function saveBlock(_block) {
   if (!existing || !existing.blockNumber) {
     return db.collection(SUBMITTED_BLOCKS_COLLECTION).updateOne(query, update, { upsert: true });
   }
-  throw new Error('Attempted to replay existing layer 2 block');
+  logger.warn('Attempted to replay existing layer 2 block.  This is expected if we are syncing');
```
druiz0992, line 112: Resyncing at the optimist will still generate an exception? Why is it different behavior?

nightfall-optimist/src/services/database.mjs
```diff
@@ -218,8 +219,8 @@ export async function setRegisteredProposerAddress(address) {
   const connection = await mongo.connection(MONGO_URL);
   const db = connection.db(OPTIMIST_DB);
   logger.debug(`Saving proposer address ${address}`);
-  const data = { proposer: address };
-  return db.collection(METADATA_COLLECTION).insertOne(data);
+  const data = { _id: address };
```
druiz0992, line 222: There are some conflicts in this file due to changes when adding a URL to the proposer. I thought I should update the url in the database as well (although I did not use this new info), but if you think it’s not needed, you can get rid of my changes.

nightfall-client/src/services/database.mjs
Westlad: Well it's not an error, it's expected behaviour so now it logs it and doesn't throw.

cli/lib/nf3.mjs
Westlad: This endpoint is removed in the latest code push

nightfall-optimist/src/services/database.mjs
Westlad: This seems like a good change though, so I will be sure to retain the URL.

druiz0992 (commented):
Will the some chances you added here  he also required for a challenger registration?

Westlad:
I think not because challengers don't register on-chain - but it remains to be seen.

Westlad:
@druiz0992 I think not because challengers don't register on-chain - but it remains to be seen.

druiz0992 (approved):

IlyasRidhuan (approved):

PR #568, opened 2022-03-09
upgrade contracts demonstration

### LINKED ISSUE #515: Upgradeable contracts while preserving user's funds and access to funds
Implement a method to upgrade contracts to solve a significant bug or to add a feature

Westlad:
fixes #515

This PR incorporates a test/demo for the upgradeable contracts that were previously merged in #552, to show that they are indeed upgradeable. 

To test, run the test script. No other setup is required
```sh
cd test/ping-pong
docker-compose build
./upgrade-test.sh
```
It will run the ping-pong test twice, upgrading the `State` contract in between (from `State.sol` to `StateV2.sol`. The only difference between the two contracts is a constant value `version`, which gets incremented. The first time
the ping-pong test runs, look for the `info` log output `State contract version is 0.0.1`.  The second time the test
runs, the contract's version will be `0.0.2`. The test takes a little while (~5 mins) so be patient.
The console output should provide helpful hints about progress. 

Note that to make the test run, an endpoint has been added to `client` to enable a contract's version to be queried.  This is a significant change just to enable a test but it is likely to be useful to track contract versions in future. It also requires us to add a contract (`StateV2`) that is only needed for test purposes. We may wish to remove this in future. 

The PR also adds instructions for how to do an upgrade.


Westlad:
Closed for now. Will revisit after initial deployment.

PR #570, opened 2022-03-10
David/staging2

druiz0992:
Minor changes to be compatible with staging/prod.

- heartbeat function to keep websocket alive. AWS closes idle connections after 1 minute
- Some additional logging to track websockets


cli/lib/nf3.mjs
```diff
@@ -666,9 +666,17 @@ class Nf3 {
     const newGasBlockEmitter = new EventEmitter();
     const connection = new WebSocket(this.optimistWsUrl);
     this.websockets.push(connection); // save so we can close it properly later
+    // Heartbeat function to keep WS open. Send beat every 15 seconds
```
ChaitanyaKonda, line 669: There is a standard implementation in most websocket implementations called `ping` and `pong` frames for just this purpose. Could we use that https://github.com/websockets/ws ?

ChaitanyaKonda (changes requested):
minor change requested

cli/lib/nf3.mjs
druiz0992: I think this is what you asked?

daveroga (approved):

ChaitanyaKonda (approved):

PR #580, opened 2022-03-15
(nightfall-browser) web-worker comlink implement to fetch circuits files from s3 bucket 

LijuJoseJJ:
this worker thread fetch/download all files in s3 bucket and insert it in IndexedDB. - run at startup script


LijuJoseJJ:
will create new PR from master as base branch

Issue #579, opened 2022-03-14
Miscellaneous Browser Tasks


IlyasRidhuan:
### Fees + Payment
- [x] Switch networks (mainnet / polygon) when deposit / (send + withdraw) `Priority 0` - Ze
- [x] Collect fees in Matic when sending `Priority 0` - Ze
- [ ] Coinbase wallet support `Priority 1`
### Storage + Security
- [x] Encryption of secrets in the browser. `Priority 0`
- [ ] Export/Reimport IndexedDB button. `Priority 3`
- [x] Anonymity set `Priority 1`
### Usability and QoL
- [x] Decimal Handling in browser `Priority 0` #539 
- [x] Exchange rates and dollar values `Priority 1`
### Infrastructure
- [x] Stream and store circuits and proving keys `Priority 0` #469 
- [ ] Web workers for multi-threading. `Priority 1`
- [ ] Off load proof computations to web worker. `Priority 1`
- [ ] Rollback event handler. `Priority 0` #523 
- [ ] Integration in Polygon's website `Priority -1`
- [ ] Websocket keep-alive in wallet `Priority 0`

### Refactor
- [ ] Remove LOCAL_PROPOSER for wallet `Priority 4`

IlyasRidhuan:
Closing, any open tasks will be migrated to V2

PR #587, opened 2022-03-18
feat: adding chaindetection and change for two networks 🪐

signorecello:
As part of #579 we need to detect in which chain are we:

If the user wants to deposit some ERC20 tokens he or she needs to deposit in the Ethereum network so it's bridged to Polygon. There's obviously no way of doing it offchain.

However for transfers or withdraws, the user can send the transaction directly to the proposer, which will relay it to the optimist to include in a block. But the proposer needs to be paid a fee. We decided it didn't make much sense to pay that fee in ETH as the gas fees make it extremely expensive. Polygon Nightfall will then interact with a different chain such as Polygon PoS chain. This means the user will pay the fee in MATIC (on Polygon chain).

For that, the browser needs to detect which chain the user is connected to. If he or she wants to deposit, the browser should make web3 prompt for a chain change to Ethereum. Otherwise it will ask to change to Polygon.

An env variable is set to distinguish between different environments. For local testing, `docker-compose.ganache.yml` will simply boot two ganache containers with different chainIds, and web3 (on the browser) will simply change between them. For testnet, it will switch between Ropsten (for Ethereum) and Goerli (for Polygon).

The counterpart of this work needs to be done in the backend.


wallet/src/components/BridgeComponent/index.tsx
```diff
@@ -145,6 +146,15 @@ const BridgeComponent = (props: any) => {
     return true;
   }
 
+  useEffect(() => {
```
IlyasRidhuan, line 149: Not sure if adding it here and refreshing based on `txType` is a good experience. It means that this pops up every time you click on the toggle between `Deposit` and `Withdraw` even if you are not actually performing any of the operations

wallet/src/components/TokenItem/index.jsx
```diff
@@ -160,6 +161,7 @@ export default function TokenItem({
               className={styles.tokenListButton}
               id={tokenSendId}
               onClick={() => {
+                changeChain('polygon');
```
IlyasRidhuan, line 164: Do we want to `await` this call. Right now a user is free to just ignore this as the sendModal still shows. If they hit send the off-chain transaction will still go through.

Of course the transaction will later not be picked up, but their database will be in a bad state.

signorecello:
Closing this PR, will PR into #596 for integration testing

PR #606, opened 2022-04-06
remove browser e2e test GitHub actions workflow

LijuJoseJJ:
This PR comment out github actions workflow which test browser e2e test, for two reason.
1.  currently in master browser e2e test are broken because significant change in UI.
2. since we are tight with deliverable it is getting bit tedious to keep changing e2e test workflow with change in code.

Understanding is, later this will be reverted back

### DIFF
--- a/.github/workflows/browser-e2e-test.yml
+++ b/.github/workflows/browser-e2e-test.yml
@@ -1,80 +1,80 @@
-name: Browser e2e test
-
-on:
-  pull_request:
-    branch:
-      - master
-
-jobs:
-  e2e-test:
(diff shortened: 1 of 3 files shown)


Westlad:
What are the change in `wallet/src/nightfall-browser/services/withdraw.js` for?

IlyasRidhuan:
It looks like linter updates (with a rename from `items` -> `withdrawParams`) ;

I'll have a check of the eslint for the `wallet` in a later PR, it looks like it needs some work

IlyasRidhuan (approved):

PR #609, opened 2022-04-08
feat: completing the MPC ceremony 🙏

### LINKED ISSUE #282: Create functionality for MPC Ceremony using Zokrates
This issue should work on the recommendation from the issue #191.
Currently ZoKrates team is looking into writing this functionality.
This issue should be updated with progress every week, so that we can decisions accordingly.

signorecello:
I can't commit the radix files because they're over 500mb in size. We'll think of a different solution. In the meantime I'm creating this PR so we can see the diff.

Closes #282

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -102,6 +102,7 @@ module.exports = {
     U: BigInt(5), // non square in Fp
   },
   MAX_QUEUE: 5,
+  RADIX_FILES_URL: 'https://nightfallv3-proving-files.s3.eu-west-1.amazonaws.com/radix',
   ENVIRONMENTS: {
     mainnet: {
       name: 'Mainnet',
--- a/config/default.js
+++ b/config/default.js
@@ -102,7 +102,9 @@ module.exports = {
     U: BigInt(5), // non square in Fp
   },
   MAX_QUEUE: 5,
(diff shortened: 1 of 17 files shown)


signorecello:
Just pushed a new version that will fetch radix files from a public s3 repo

signorecello:
Here's exactly what is changed:

- Radix files are stored in an S3 bucket, which link is set in an `env` variable. They're ignored in gitignore and nodemon.
- I added a port in `docker-compose.dev.yml` so we can (in dev) access the different endpoints without having to run the whole `./start-nightfall` in future development
- At `index.mjs`, node checks if the radix files are already present in the volume. If not, it will get them from the provided url.
- At `generateKeys.mjs` the phase2 starts. It computes the `mpc.params` from the radix files. Then makes a secure pseudo-random number using `crypto` and contributes to it. It then generates another random number and applies a random beacon to it. The result is `final.params`. It then takes these params and creates the proving and verifying keys. After that, the process continues as previous.
- I obviously removed the old `setup` command and file, as it's replaced with this trusted setup now

signorecello:
So far I haven't seen any problem regarding RAM constraints, and I was able to run the MPC successfully (although quite visibly slower) with 8gb RAM

Westlad (approved):

daveroga (approved):
LGTM

PR #612, opened 2022-04-08
Westlad/enhanced gas test

Westlad:
This PR enhances the coverage of the test-gas test.  This test will now measure the L1 and L2 gas usage for all four types of nightfall transaction (deposit, single transfer, double transfer, withdraw) and also for the finalise withdraw.  The latter is already covered by the e2e tests, but it uses minimal resources, and so I included it here so that this test provides a complete measure of all gas used.

To use this test locally, you will need to set the environment variable `TRANSACTIONS_PER_BLOCK` to the value that you want to use in the test run.  For example:
```sh
export TRANSACTIONS_PER_BLOCK=32
```
Do this in both the window from which you start nightfall and the window from which you run the test. _Do not_ forget to set it back to 2 when you have finished (same command as above but with 2, rather than 32) otherwise you'll spend half a day wondering why none of your other tests pass any more. Closing the windows will also work as the export does not persist past logout.

Other than that, the test is completely standard `./start-nightfall -g -s -d` and `npm run test-gas`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,17 +108,37 @@ make the standard tests fast.
 
 In reality, a value of two transactions per block, although convenient for testing, wouldn't make
 very efficient use of Optimism. A more realistic value is 32 transactions per layer 2 block. This
-value can be configured by the environment variable `TRANSACTIONS_PER_BLOCK` in the
-`docker-compose.yml` file (part of the `optimist` service). This is important for the Block Gas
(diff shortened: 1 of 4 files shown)


signorecello (approved):
LGTM 🙂

druiz0992 (approved):

PR #615, opened 2022-04-12
Local Development Environment For Wallet

IlyasRidhuan:
This PR adds a local development environment for the wallet that now starts up using `npm run start`.

It does this using an environment variable `REACT_APP_MODE`- the prefix `REACT_APP` is a requirement for browser environments. This environment determines the list of tokens to be used in the wallet - `supported-token-list`.

Note that the value for the `ERC20Mock` is hardcoded for now, it will always be the same deployed address provided we do not make modifications to the deployment process we currently have. Doing this dynamically using `getContractAddress` from `optimist` is a possibility but ends up being messier due to the `promise` handling.

This PR also includes some `eslint` fixes and `alias` for some deep directories (reducing the unwieldiness of long `../../..` imports) as well as updates to the README.

### To Test
- `npm ci`
- `npm run start`
- When the browser starts, there should be a single token using the deployed `ERC20Mock` address.


ChaitanyaKonda:
I run the wallet up following all the instructions and I run into this error

Screenshot 2022-04-13 at 18 49 54

ChaitanyaKonda:
After running the wallet up following all the instructions and I run into this error

Screenshot 2022-04-13 at 20 55 19

Westlad (approved):

Westlad:
Clearing indexDb, local storage and reseting the account in Metamask cleared the error for me.

ChaitanyaKonda (approved):

PR #619, opened 2022-04-13
Update timber paths in browser and client

### LINKED ISSUE #540: Root for old commitments used in transfer and withdraw should not reveal the L2 block the transaction was aprt

IlyasRidhuan:
This PR closes #540.

This adds a new collections `sibling_path` to the databases. This stores the:
- base sibling path ( An object containing sibling information from the block that the commitment is mined in)
- diffPaths ( An array of sibling path objects representing the updated sibling paths).

To ensure that the diffPaths array does not have unbounded growth, we restrict it to the last 4 blocks. If a rollback was to occur that is deeper than 4 blocks, we would have to revert to the base sibling path.

This can be tested by just running the standard tests.

### Notes
- The storage of diffPaths is currently not optimal, having a more compressed storage method will enable us to extend the maximum length of this array.
- siblingPaths for already spent commitments are still updated, this is because there is no clear indication of when a spent commitment is past the finalised window.
- This PR has not been tested with rollbacks, this will need to be revisited when rollbacks are enabled.
- There are numerous performance optimisations that can be tackled to make the update process quicker.

### DIFF
--- a/nightfall-client/src/event-handlers/block-proposed.mjs
+++ b/nightfall-client/src/event-handlers/block-proposed.mjs
@@ -84,6 +84,8 @@ async function blockProposedEventHandler(data) {
         s.leafIndex,
         s._id,
         siblingPath,
+        HASH_TYPE,
+        TIMBER_HEIGHT,
       );
       return updateSiblingPath(s._id, block.blockNumberL2, updatedPath, updatedTimber.root);
     }),
--- a/nightfall-client/src/event-handlers/block-proposed.mjs
+++ b/nightfall-client/src/event-handlers/block-proposed.mjs
@@ -5,7 +5,6 @@ import {
   markNullifiedOnChain,
   markOnChain,
   countCommitments,
-  setSiblingInfo,
   countWithdrawTransactionHashes,
   isTransactionHashWithdraw,
 } from '../services/commitment-storage.mjs';
@@ -17,6 +16,9 @@ import {
   saveTransaction,
   saveBlock,
   setTransactionHashSiblingInfo,
+  getAllSiblingPaths,
+  updateSiblingPath,
+  saveSiblingPath,
 } from '../services/database.mjs';
 import { decryptCommitment } from '../services/commitment-sync.mjs';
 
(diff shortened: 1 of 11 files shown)


wallet/src/nightfall-browser/services/database.js
```diff
@@ -268,3 +270,81 @@ export async function markWithdrawState(transactionHash, withdrawState) {
...
 }
+
+/*
+Functions for manipulating the sibling path collection
+*/
+export async function getAllSiblingPaths() {
+  const db = await connectDB();
+  const commitments = await db.getAllKeys(SIBLING_COLLECTION);
+  return Promise.all(
+    commitments.map(async c => {
```
Westlad, line 281: This seems to me to iterate over the collection to get the sibling paths.  Would not an atomic `find()` to return an array of `siblingPaths` be more efficient?

nightfall-client/src/event-handlers/block-proposed.mjs
```diff
@@ -57,14 +64,37 @@ async function blockProposedEventHandler(data) {
   await Promise.all(dbUpdates);
   const updatedTimber = Timber.statelessUpdate(latestTree, blockCommitments);
   await saveTree(transactionHashL1, block.blockNumberL2, updatedTimber);
+
+  // We update all sibling paths to account for the newly inserted leaves.
+  const currentPaths = await getAllSiblingPaths();
```
Westlad, line 69: Would this attempt to pull all sibling paths from the DB each time a block is proposed and process them all?  Is there an issue when we have millions of commits? I'm worried that the processing of a block event will become too slow.

wallet/src/nightfall-browser/services/database.js
IlyasRidhuan: It would but I don't know of a `find-like` call in indexedDB.

The reason I have to do a `getAllKeys` and iterate is that the indexedDB schema for this collection is keyed by commitmentHash. Unfortunately, `get` in indexedDB doesn't also return the key, just the value. As I need the commitmentHash as part of the return object I think I'm forced to iterate (so I can re-pair up the commitmentHash to the corresponding stored values).

Alternatively I can duplicate commitmentHash in the key and value stored - but that violates DRY.

nightfall-client/src/event-handlers/block-proposed.mjs
Westlad: Fair enough

wallet/src/nightfall-browser/services/database.js
Westlad: Fair enough

nightfall-client/src/event-handlers/block-proposed.mjs
Westlad: If it's restricted to just the commits that the client is concerned with then it's not really an issue.

IlyasRidhuan: Yep it will. This will definitely slow down block event processing but unfortunately sibling path updates need to happen every block.

There are a couple of ways to mitigate this (dumping them here for reference later)

- Optimisations we can (need to) do
    1. Improve the performance of the sibling path update (i.e the batching and "dead-heading" techniques I described in the show-n-tell).
    2. Only update the sibling paths of commitments "in-play" (i.e. inside the challenge window)
    3. More efficient storage and retrieval patterns for updated paths.

- We can decide on `MAX_UPDATES` to be performed on a sibling path based on some analysis of the anonymity set. In my mind, there should be a "sweet spot" where "I have updated my sibling path for x number of blocks and there are y total blocks in the blockchain" such that the improvement in anonymity of updating to x+1 is minimal.

- For the client/browser slow block processing isn't terminal, you just might take longer to notice when you have received or successfully spent a commitment. 

I agree that if a user owns a million commitments today then updating the siblingPath of all these commitments will be probably be slow.

Westlad (approved):

ChaitanyaKonda:
Will need to be done for the new wallet version

PR #626, opened 2022-04-23
double spend bugfix

Westlad:
fixes https://github.com/EYBlockchain/nightfall_3_private/issues/38

It is possible to double spend by using a nullifier which is congruent but not equal to one that has been used before; the proof will still verify because it works over the BN128 group order.  Simply constraining the value to be less than the group order in `Verifier.sol` will not work because this is only used in the event of a challenge.

There are really three solutions.  The more complex approach is to challenge any values which are greater than the group order.  This has the advantage that it is easy to extend to other variables but is more time-consuming to implement. Alternatively, one can simply drop any transactions which have values greater than the group order, which will work only while we have total control of block creation.

The approach taken here however was to note that when we generate a proof, we actually truncate the nullifier hash to 31 bytes anyway. We can use this to our advantage and simply make the nullifier hash 31 bytes everywhere, including in the `Transaction` struct.  This prevents anyone proposing a block containing a transaction with a nullifier >31 bytes long.  This obviates the need for a challenge mechanism. To assist the proposer, we also check that a Transaction has a nullifier not exceeding 31 bytes, so that any such will be dropped.

### DIFF
--- a/common-files/classes/transaction.mjs
+++ b/common-files/classes/transaction.mjs
@@ -25,7 +25,7 @@ function keccak(preimage) {
     { t: 'bytes32', v: preimage.ercAddress },
     { t: 'bytes32', v: preimage.recipientAddress },
     ...preimage.commitments.map(ch => ({ t: 'bytes32', v: ch })),
-    ...preimage.nullifiers.map(nh => ({ t: 'bytes32', v: nh })),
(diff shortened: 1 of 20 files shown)


Westlad:
Closed as we've thought of a better way to tackle this.

PR #630, opened 2022-04-27
blocks can be made at any point in time

Westlad:
This PR creates a new `optimist` endpoint `/block/make-now` (and wrappers it in the Nf3 class method `makeBlockNow`).  When called, `optimist` will make a 'short' block using whatever transactions it has in its database, even if that is less than `TRANSACTIONS_PER_BLOCK`. The short block is still made by queuing `conditionalMakeBlock` in the usual way, so the queue is never bypassed. The new code will not make a block if there are no transactions: there must be at least one. If there are already >=`TRANSACTIONS_PER_BLOCK` transactions, then calling this endpoint has no effect.

To test, an additional test stanza has been added to the ERC1155 tests, whereby `makeBlockNow` is called and a single deposit transaction is made.  The test then waits for a block to be created.  Normally it would wait forever (well, until it times out) but because the end point has been called, a block with a single transaction will be made.  you can double check this by looking at the logs and you will find a block with a `transactionHashes` of length 1, rather than 2.

### DIFF
--- a/cli/lib/nf3.mjs
+++ b/cli/lib/nf3.mjs
@@ -182,12 +182,23 @@ class Nf3 {
 
   /**
     Gets the number of unprocessed transactions on the optimist
+    @method
     @async
     */
 
   async unprocessedTransactionCount() {
     const { result: mempool } = (await axios.get(`${this.optimistBaseUrl}/proposer/mempool`)).data;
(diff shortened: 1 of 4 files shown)


daveroga (approved):
Tested successfully

PR #631, opened 2022-04-27
transaction hashes merkle tree

ChaitanyaKonda:


IlyasRidhuan:
I think `eslint` is failing because you have deleted the `package-lock.json` from `wallet/`. So it cannot run `npm ci`

ChaitanyaKonda:
> I think `eslint` is failing because you have deleted the `package-lock.json` from `wallet/`. So it cannot run `npm ci`

Oh yes, that was a mistake, I am pushing it

nightfall-deployer/contracts/Shield.sol
```diff
@@ -129,37 +124,37 @@ contract Shield is Stateful, Structures, Config, Key_Registry, ReentrancyGuardUp
...
-        uint256 index
+        bytes32 transactionsHash,
+        Transaction calldata t,
+        uint256 index,
+        bytes32[6] memory siblingPath
     ) external {
-        // check this block is a real one, in the queue, not something made up.
-        state.isBlockReal(b, ts, blockNumberL2);
+        // check this block is a real one, in the queue, not something made up and that the transaction exists in the block
+        /* state.areBlockAndTransactionValid(b, transactionsHash, t, index, siblingPath); */
```
IlyasRidhuan, line 140: Should this be uncommented? We would need this check in `finaliseWithdrawal`?

nightfall-deployer/contracts/Shield.sol
```diff
@@ -129,37 +124,37 @@ contract Shield is Stateful, Structures, Config, Key_Registry, ReentrancyGuardUp
...
 
     function finaliseWithdrawal(
         Block calldata b,
-        uint256 blockNumberL2,
-        Transaction[] calldata ts,
-        uint256 index
+        bytes32 transactionsHash,
+        Transaction calldata t,
+        uint256 index,
+        bytes32[6] memory siblingPath
```
IlyasRidhuan, line 137: Is fixing the length of the `siblingPath` problematic if we have blocks of different size? I guess this should be dynamic based on the number of transactions in a block.

PR #633, opened 2022-04-28
feat: scripts for MPC + removing MPC from the worker process 🔨

signorecello:
This PR implements phase2 for the MPC ceremony.

We use phase1 from the perpetual powers of tau. This phase2 is circuit-specific.

You can run ``npm run ceremony`` to start a new ceremony, which will pick the compiled circuits, fetch the radix files, and produce some ``mpc params``. If you don't need to start a new ceremony and you want to contribute to the existent one (which is probably the most interesting usage), you can run ``sh MPC/mpc_contribute.sh`` which will start up a bash script that will download and install Docker if you don't have it, then contribute to the existent one.

This contribution will download the existent mpc params from our S3 bucket, and will contribute with the source of entropy that is asked in the bash script. After finishing, you should have your new mpc params in the output folder `params/out``. That's the one you should send us, to be replaced in the S3 bucket and handed over to the next contributor


MPC/contribution.js
```diff
@@ -0,0 +1,52 @@
...
+      resolve();
+    });
+  });
+}
+
+const main = async () => {
+  try {
+    if (!process.env.RANDOM) throw new Error('Please provide source of randomness');
+    if (!process.env.NAME) throw new Error('What is... your name? What is... your quest?');
+
```
druiz0992, line 36: If you print the input randomness, you will see it does not match what you entered.

MPC/mpc_contribute.sh
```diff
@@ -0,0 +1,21 @@
...
+read RANDOM
+
+echo "Thanks, hang on a bit!"
+
+if ! docker --version; then
+    curl -fsSL https://get.docker.com -o get-docker.sh
+    sh get-docker.sh
+fi
+
+docker build -t contribute --build-arg NAME=${NAME} --build-arg RANDOM=${RANDOM} .
```
druiz0992, line 16: Add "${NAME}" and "${RANDOM}" instead. Otherwise, it will fail if one enters a space as one of the characters

druiz0992 (changes requested):
I would like to have some instructions on how to run this MPC in the main README section

Westlad (approved):

signorecello:
For clarity, the MPC ceremony has been moved to a new repo, here: https://github.com/maticnetwork/nightfall_phase2ceremony

PR #634, opened 2022-04-29
add amount restrictions for mainnet and goerli ERC contracts

Westlad:
This PR extends the test code that put restrictions on the amount that can be deposited or withdrawn.
image
It selects the contract addresses to put limits on via the environment variable `ETH_NETWORK` which is the same one used to select the truffle configuration for deployment, so it's transparent to the deployer.  Value restrictions are created in the contract migration.

Test in the normal way, although obviously it's not possible to directly test mainnet and goerli values.  However the symmetry between those and the ganache (`ETH_NETWORK=blockchain1`) case.

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -290,6 +290,28 @@ module.exports = {
           amount: '1000000000000',
         },
       ],
+      development: [
+        {
+          name: 'MockERC20',
+          address: '0xB5Acbe9a0F1F8B98F3fC04471F7fE5d2c222cB44',
+          amount: 200,
+        },
+        {
+          name: 'Test-Eth',
+          address: '0x3f152B63Ec5CA5831061B2DccFb29a874C317502',
+          amount: '10000000000000000000000',
+        },
+        {
+          name: 'MATIC',
+          address: '0x499d11E0b6eAC7c0593d8Fb292DCBbF815Fb29Ae',
+          amount: '10000000000000000000000',
+        },
+        {
+          name: 'USDC',
+          address: '0x07865c6e87b9f70255377e024ace6630c1eaa37f',
+          amount: '1000000000000',
+        },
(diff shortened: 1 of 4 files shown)


config/default.js
```diff
@@ -267,28 +267,64 @@ module.exports = {
...
+          address: '0x499d11E0b6eAC7c0593d8Fb292DCBbF815Fb29Ae',
+          amount: '10000000000000000000000',
+        },
+        {
+          name: 'USDC',
+          address: '0x07865c6e87b9f70255377e024ace6630c1eaa37f',
+          amount: '1000000000000',
+        },
+      ],
+      mainnet: [
```
druiz0992, line 293: Why aren't USDT and DAI in the list of restricted tokens?

druiz0992 (changes requested):
minor change request

config/default.js
Westlad: The browser doesn't support them as far as I know.

config/default.js
```diff
@@ -267,28 +267,64 @@ module.exports = {
...
-        amount: '10000000000000000000000',
-      },
-      {
-        name: 'USDC',
-        address: '0x07865c6e87b9f70255377e024ace6630c1eaa37f',
-        amount: '1000000000000',
-      },
-    ],
+    tokens: {
+      blockchain1: [
```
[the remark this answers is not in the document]
Westlad: Yes, if that would be helpful, I'll push an update

config/default.js
Westlad: I can add them though if you like, it just increases the deployment cost a little bit.

druiz0992 (approved):

daveroga (approved):

PR #640, opened 2022-05-06
feat: removing MPC flag, now the setup only runs for stubs 🏁

signorecello:
The worker now exports the keys from the MPC ceremony, unless you choose to use stubs, in which case it will run the normal setup


Westlad:
If I change the circuits, how would I test that locally, given that the MPC-generated keys will now be invalid?

signorecello:
> 

That's a good question... I guess we need to keep the MPC flag then

PR #646, opened 2022-05-09
sync fix

Westlad:
If we find a transaction in a block that we have not seen before then we save it but make sure its marked as not in the mempool because it must now be onchain.  This achieves that be enabling a new parameter for calling transactionSubmittedEventHandler, so that `mempool` can be set (true or false, false in this case).

We also have to set the `blockNumberL2` of the block that this transaction is in (by default it will be assigned `-1`)


ChaitanyaKonda (approved):

druiz0992 (dismissed):

Westlad:
Closed because code incorporated in later PR

PR #648, opened 2022-05-10
Update restrictions

ChaitanyaKonda:
Description:
If a user does two deposits with the max deposit amount, then a transfer, the user already owns a commitment of value bigger than the max withdraw limit. To control this, this PR introduces max deposit amount to be a quarter of max withdraw amount. 

Other fixes:
- It also fixes restriction tests which always pass despite depositing/withdrawing a value over or under the restriction limit because
  - For a successful test, the test expects to catch an error message `Transaction has been reverted by the EVM`. This error message is also thrown by the test by calling `expect.fail('Transaction has not been reverted by the EVM');` Which meant the test would work even if deposit/withdraw calls were commented out because the error was always thrown and caught as expected.
  - Also, the values used for deposit or withdraw tests were always below the max limit. So the tests created valid deposit and withdraw transactions which did not revert on the blockchain as they were under the max deposit and withdraw restrictions
- Updated the contracts such that if 1000 MATIC is set as the withdraw limit, 1000 MATIC can be withdrawn successfully whilst 1001 will fail. Earlier only values `<1000` MATIC were accepted and values equal to and over 1000 MATIC reverted. 

Test:
Automatically done by Github Action.
For a further review, look in blockchain logs of ganache test for transactions with revert messages
`Value is above current restrictions for deposits` and `Value is above current restrictions for withdrawals`

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -285,9 +285,9 @@ module.exports = {
     tokens: {
       blockchain1: [
         {
-          name: 'MockERC20',
+          name: 'ERC20Mock',
           address: '0xB5Acbe9a0F1F8B98F3fC04471F7fE5d2c222cB44',
-          amount: 200,
+          amount: '200',
         },
         {
           name: 'Test-Eth',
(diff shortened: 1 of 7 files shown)


druiz0992 (commented):
Why are we limiting withdraws at all? Is it some type of safeguard?

ChaitanyaKonda:
> 

It is. In the event of an attack, this will control how much can be withdrawn at once.

ChaitanyaKonda:
> Why are we limiting withdraws at all? Is it some type of safeguard?

It is. In the event of an attack, this will control how much can be withdrawn at once.

Westlad (approved):

druiz0992 (approved):

PR #651, opened 2022-05-10
Fix upgrade contracts

daveroga:
State contract reference was missing in the upgrade script


Westlad (changes requested):
Should there not be a line to deploy an `upgradeProxy` for `State`?

daveroga:
I thought the state couldn't be migrated with the plugin. But I read that the state is preserved. So yes, upgradeProxy for State was missing. Added.

ChaitanyaKonda (approved):

Westlad (approved):

PR #657, opened 2022-05-11
resync cleanup

Westlad:
Currently,  when we restart optimist, it will resync before it starts listening to events. It: pauses processing of its input queues; replays all existing transactions; restarts processing its queues; subscribes to new events from the point of the last L2 block in its freshly-resynced database.

This can cause errors to display because any transactions which arrived after the last L2 block are played twice.  Optimist will drop the second event, so no harm is done, but it will show an error in the logs.

Additionally, this process is overly complex: we stop processing events in the queues while resyncing.  This means we could simply start our event listeners as our first action;  they won't hurt the resync because any events they hear will just join the queue and wait until the resync is completed.  At that point the queue will start moving again and they'll be processed. Thus there is no need for the complex 'start-listening-from-where-the-resync-stopped', at all.

This PR starts the event listeners as soon as the container is up.

Testing is a little tricky.  Fire up the browser locally and make a set of transactions to give some complete blocks.  Then make another transaction _but don't press 'send'_.
In a terminal remove the optimist1 container: `docker compose -f docker-compose.yml -f docker-compose.ganache.yml -f docker-compose.stubs.yml -f docker-compose.dev.yml rm optimist1`
Then up it again (same command but replace `rm` with `up`).  You should see the resync happening as `optimist1` comes up.  While the resync is still running, press 'send' to throw in your transaction. When the rsync is finished, you should see your transaction added neatly afterwards, once the queues start moving.

### DIFF
--- a/nightfall-optimist/src/event-handlers/subscribe.mjs
+++ b/nightfall-optimist/src/event-handlers/subscribe.mjs
@@ -87,7 +87,7 @@ export async function waitForContract(contractName) {
  * @param arg - List of arguments to be passed to callback, the first element must be the event-handler functions
  * @returns = List of emitters from each contract.
  */
-export async function startEventQueue(lastSyncedBlock, callback, ...arg) {
+export async function startEventQueue(callback, ...arg) {
   const contractNames = [
     STATE_CONTRACT_NAME,
     SHIELD_CONTRACT_NAME,
@@ -96,10 +96,7 @@ export async function startEventQueue(lastSyncedBlock, callback, ...arg) {
   ];
   const contracts = await Promise.all(contractNames.map(c => waitForContract(c)));
   const emitters = contracts.map(e => {
-    let emitterC;
-    // if we've just resynced, start from the end of the sync:
-    if (lastSyncedBlock) emitterC = e.events.allEvents({ fromBlock: lastSyncedBlock + 1 });
-    else emitterC = e.events.allEvents();
+    const emitterC = e.events.allEvents();
     emitterC.on('changed', event => callback(event, arg));
     emitterC.on('data', event => callback(event, arg));
     return emitterC;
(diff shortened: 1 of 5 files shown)


nightfall-optimist/src/index.mjs
```diff
@@ -29,24 +29,24 @@ const main = async () => {
...
-          return enqueueEvent(conditionalMakeBlock, 0, proposer);
-        }
-        // eslint-disable-next-line no-void, no-useless-return
-        return void false; // This is here to satisfy consistent return rules, we do nothing.
-      });
-      // We enqueue a message so that we can actualy trigger the queue.end call even if we havent received anything.
-      // This helps in the case that we restart client and we are the current proposer.
-      await enqueueEvent(() => logger.info('Start Queue'), 0);
+    // start the event queue
+    await startEventQueue(queueManager, eventHandlers, proposer);
```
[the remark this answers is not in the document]
Westlad: Yes, this is a good catch.  I've fixed it in the way you suggested.

nightfall-optimist/src/event-handlers/subscribe.mjs
```diff
@@ -96,10 +96,7 @@ export async function startEventQueue(lastSyncedBlock, callback, ...arg) {
   ];
   const contracts = await Promise.all(contractNames.map(c => waitForContract(c)));
   const emitters = contracts.map(e => {
-    let emitterC;
-    // if we've just resynced, start from the end of the sync:
-    if (lastSyncedBlock) emitterC = e.events.allEvents({ fromBlock: lastSyncedBlock + 1 });
-    else emitterC = e.events.allEvents();
+    const emitterC = e.events.allEvents();
```
ChaitanyaKonda, line 99: Every time optimist is restarted but the database is intact, it starts syncing from the start of the contract ?
This would be inefficient if syncing after a few years

Westlad: No, that won't happen.  Without any argument, `.allEvents` will start listening for events from the current block onwards.  We're just telling it to subscribe to all events that come along.

ChaitanyaKonda (approved):

PR #661, opened 2022-05-13
Adds some functionality and fixes to wallet

IlyasRidhuan:
1) Syncing indicator for Nightfall to show when the wallet has completely synced the circuits and chain state

2) Chain detection, wallet will check to see the user is currently connected to Ethereum main net and disallow deposits otherwise.

3) User inputs in deposit and withdraw - raises a warning if the amounts are greater than the current restrictions.

4) Updated the txinfo modal that pops up when clicking on a transaction in the 'transactionPage'


daveroga (approved):

imagobea (approved):
Just leaving comments for future iterations. These are certainly nice additions, thank you.

1. `Syncing indicator` - We could improve this by taking timber into account, which is what takes longer (the indicator was green, but I was still receiving blocks)
2. `Chain detection` - Works fine but Metamask is always difficult (dialogs were not 'popping up' and that was conflicting with the sync indicator...)
3. `L2 Bridge page validations` - I would disable the `Transfer` button altogether (looks like you can carry on with the tx). Also, we should expand validations for balance and `0` value.

wallet/src/nightfall-browser/event-handlers/block-proposed.js
```diff
@@ -56,7 +56,7 @@ async function blockProposedEventHandler(data, ivks, nsks) {
     const storeCommitments = [];
     const tempTransactionStore = [];
     if (
-      (transaction.transactionType === '1' || transaction.transactionType === '2') &&
+      (Number(transaction.transactionType) === 1 || Number(transaction.transactionType) === '2') &&
```
daveroga, line 59: Number(transaction.transactionType) === '2' should be
Number(transaction.transactionType) === 2

IlyasRidhuan: good catch!

PR #662, opened 2022-05-13
fix: erc20 test race condition

ChaitanyaKonda:
There is a race condition problem where a transfer needed commitment from prior transaction which has not been written to the client’s database yet. Adding some waiting time solves it. So it is just the test and not the functionality of restricted withdrawals that is broken

### DIFF
--- a/test/e2e/tokens/erc20.test.mjs
+++ b/test/e2e/tokens/erc20.test.mjs
@@ -529,7 +529,7 @@ describe('ERC20 tests', () => {
               fee,
             );
             await emptyL2(nf3Users[0]);
-            await new Promise(resolve => setTimeout(resolve, 15000));
+            await new Promise(resolve => setTimeout(resolve, 30000));
           }
 
           // console.log('withdrawing', trnsferValue * 6);


Westlad (approved):

test/e2e/tokens/erc20.test.mjs
```diff
@@ -529,7 +529,7 @@ describe('ERC20 tests', () => {
               fee,
             );
             await emptyL2(nf3Users[0]);
-            await new Promise(resolve => setTimeout(resolve, 15000));
+            await new Promise(resolve => setTimeout(resolve, 30000));
```
druiz0992, line 532: Why increasing this timer to 30 seconds solves the issue? Doesn’t it simply reduce the possibility of it happening rather than eliminating the problem?

ChaitanyaKonda: I'm not sure I understand the question. The description explains why we added longer wait.

In theory, the proper way to do these tests would be to test for success from updates to balance after a transaction as suggested in #513 Ze wanted to pick this as a different issue instead of in the e2e tests PR he created.

druiz0992: That was my question i guess. This is not so much a permanent fix but a workaround to minimize the probability of it happening.

druiz0992 (approved):

PR #671, opened 2022-05-19
Better estimating of proposer gas price and gas limit

### LINKED ISSUE #666: Improve proposer gas estimation
It's important for smooth running that Proposer is able to propose its blocks.  However its current gas-pricing mechanism is overly pessimistic and that results in more cost than necessary (never falling below 200 GWei).  A mechanism similar to that used by the wallet (with a suitable price multiplier) would probably be a better solution.

Westlad:
fixes #666 

The proposer uses the nf3 class to submit transactions.  The way that this set the gasPrice and gasLimit was quite crude and results in the proposer paying over the odds for gas and using such a high gas limit that there is not enough in the proposer's account to submit the transaction.  This PR fixes that and gives much better values for both.  The strategy used is essentially the same as for the wallet.

For gasPrice:
- it uses Etherescan to get an average value for the last several block.  If that fails...
- it uses web3js to find the gasPrice for the previous block,  If that fails...
- it uses a fixed value (set at 100 GWei, currently).
In each case, the resulting figure is multiplied by a fixed multiplier (`GAS_PRICE_MULTIPLIER`) to give high confidence of a successful transaction.

For gasLimit:
- it uses web3.estimate gas.  If that fails (it often does)...
- it uses a fixed value (set at 4MGas).
In each case, the resulting figure is multiplied by a fixed multiplier (`GAS_MULTIPLIER`) to give high confidence of a successful transaction.

To test.  Use the standard `npm t` test but temporarily edit the `test` script in `package.json` so that `LOG_LEVEL=debug`.  You can then see the gas price and gas limit chosen, as the test proceeds.

### DIFF
--- a/cli/lib/nf3.mjs
+++ b/cli/lib/nf3.mjs
@@ -18,6 +18,8 @@ import {
   GAS_MULTIPLIER,
   GAS,
   GAS_PRICE,
+  GAS_PRICE_MULTIPLIER,
+  GAS_ESTIMATE_ENDPOINT,
 } from './constants.mjs';
 
 // TODO when SDK is refactored such that these functions are split by user, proposer and challenger,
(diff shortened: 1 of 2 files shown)


cli/lib/nf3.mjs
```diff
@@ -204,47 +206,72 @@ class Nf3 {
...
+      gasLimit = GAS; // backup if estimateGas failed
+    }
+    return Math.ceil(Number(gasLimit) * GAS_MULTIPLIER); // 50% seems a more than reasonable buffer.
+  }
+
+  async estimateGasPrice() {
+    let proposedGasPrice;
+    try {
+      // Call the endpoint to estimate the gas fee.
+      const res = (await axios.get(GAS_ESTIMATE_ENDPOINT)).data.result;
```
druiz0992, line 229: Why are we using a lamba function? Can’t the call to Etherscan be embedded directly?

IlyasRidhuan: Mostly for ease-of-use. The etherscan call requires an API_KEY that needs to be kept private. For the `sdk` we could extract this secret information from an environment variable; however, the wallet also needs this and there is no way to keep this information secret in the browser.

So we put it behind a gateway and lambda (where the API_KEY is stored in the secrets manager).

druiz0992 (approved):

deauna (commented):
cheers

PR #678, opened 2022-05-26
remove unnecessary deploys

Westlad:
We deploy `Structures.sol` and `Config.sol` but it's unnecessary.  These contracts are inherited by several other contracts but they're never called by anything.


daveroga (approved):

druiz0992 (commented):
How would i deploy this new version of contracts?

druiz0992 (approved):

PR #683, opened 2022-05-30
refactor proposer to be in main folder

druiz0992:
Proposer sits in a test folder, and it results very inconvenient to import it and use it outside tests as part of infrastructure code. We also expect to release scripts to help external parties deploy their own version of proposers. Therefore, it seems a better option to have a proposer folder in the main code structure and remove it from tests.

There is no change in functionality.


israelboudoux:
LGTM

Westlad (approved):

daveroga (approved):
LGTM

PR #689, opened 2022-05-30
Refactor challenger to be in main folder

daveroga:
Change Challenger to an independent folder in the root of the solution. Expecting to release scripts to help external parties deploy their own version of challengers. Remove it from cli folder.

There is no change in functionality.


druiz0992 (approved):

ChaitanyaKonda:
Similar to user inside test file or user as wallet, challenger/proposer apps are not core nightfall and are part of and only used for testing. So challenger/proposer app should go under `test` folder?

Usually, developers of nightfall would build their own version of challenger/proposer app using nightfall SDK and use it with the core nightfall (nightfall-client/wallet, nightfall-optimist).

signorecello:
I 200% agree with @druiz0992 , even if we don't consider proposer/challenger as core parties, a `test` folder isn't definitely the right place for them to be. We'll definitely find a better solution but for now I 200% agree we should have them in a place where external parties can easily identify and use

daveroga:
Without a proposer you can't make blocks. So it's not the core but very important for the users to understand how to build a proposer and `test` folder doesn't seem the proper folder.
Challengers and Liquidity providers were not in `test` but `cli` folder. This `cli` folder problably is the one that should be renamed as `app` or `roles` as sugested by @druiz0992 for this purpose and structured in the proper way.

Issue #681, opened 2022-05-30
Client reports error replaying Layer2 block


druiz0992:
Client reports a replay Layer2 block error if it starts with a non-empty mongoDb.

PR #698, opened 2022-06-02
Client deployment to use with the sdk

daveroga:
Fixes: #711,  #681

Build client to be able to test with the SDK for testnet (goerli) and production (mainnet).
Based in docker-compose and script to start client.

- Updates in `nightfall-client` and `worker` to download files from the public S3 bucket of the contracts abi in json and output of circuits based on some parameters.
- Updated README in `nightfall-client` that was outdated and added information about to configure and run the client.

To test the client run:
- Configure your `client.env` in the root folder of `nightfall-client`
- Run the script `./start-client -d` in the root folder of `nightfall-client`
- Run the test in `client` folder with correct values and connect:
  ```
  TRANSFER_VALUE=1 USER1_MNEMONIC="xxx xxx xxx xxx xxx xxx xxx xxx xxx xxx xxx xxx" 
  USER1_KEY="xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx" TRANSACTIONS_PER_BLOCK=32 
  BLOCKCHAIN_WS_HOST="your goerli wss provider url" npm run test-erc20-cli
  ```

Now the client is ready to persist the data in the volume and when you start again with `./start-client -d` it syncs again but without removing the data from the volumes.
If you want to remove the volumes first you can run `./start-client -d -r` with the `-r` for removing the volumes first.

Tasks
- [x] Build client service based on docker-compose
- [x] Run some tests with the SDK to test client in testnet environment
- [x] Fix persistent database error duplicated key when syncing

### DIFF
--- a/common-files/package-lock.json
+++ b/common-files/package-lock.json
@@ -365,6 +365,14 @@
       "resolved": "https://registry.npmjs.org/aws4/-/aws4-1.11.0.tgz",
       "integrity": "sha512-xh1Rl34h6Fi1DC2WWKfxUTVqRsNnr6LsKz2+hfwDxQJWmrx8+c7ylaqBMcHfl1U1r2dsifOvKX3LQuLNZ+XSvA=="
     },
+    "axios": {
+      "version": "0.21.4",
+      "resolved": "https://registry.npmjs.org/axios/-/axios-0.21.4.tgz",
+      "integrity": "sha512-ut5vewkiu8jjGBdqpM44XxjuCjq9LAKeHVmoVfHVzy8eHgxxq8SbAVQNovDA8mVi05kP0Ea/n/UzcSHcTJQfNg==",
+      "requires": {
+        "follow-redirects": "^1.14.0"
+      }
+    },
     "base-x": {
       "version": "3.0.9",
       "resolved": "https://registry.npmjs.org/base-x/-/base-x-3.0.9.tgz",
@@ -644,14 +652,6 @@
       "integrity": "sha512-yjLXh88P599UOyPTFX0POsd7WxnbsVsGohcwzHOLspIhhpalPw1BcqED8NblyZLKcGrL8dTgMlcaZxV2jAD41Q==",
       "requires": {
         "mimic-response": "^1.0.0"
-      }
-    },
-    "axios": {
-      "version": "0.21.4",
-      "resolved": "https://registry.npmjs.org/axios/-/axios-0.21.4.tgz",
-      "integrity": "sha512-ut5vewkiu8jjGBdqpM44XxjuCjq9LAKeHVmoVfHVzy8eHgxxq8SbAVQNovDA8mVi05kP0Ea/n/UzcSHcTJQfNg==",
-      "requires": {
-        "follow-redirects": "^1.14.0"
       }
     },
     "color": {
@@ -1269,6 +1269,11 @@
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/fn.name/-/fn.name-1.1.0.tgz",
       "integrity": "sha512-GRnmB5gPyJpAhTQdSZTSp9uaPSvl09KoYcMQtsB9rQoOmzs9dH6ffeccH+Z+cv6P68Hu5bC6JjRh4Ah/mHSNRw=="
+    },
+    "follow-redirects": {
+      "version": "1.15.1",
+      "resolved": "https://registry.npmjs.org/follow-redirects/-/follow-redirects-1.15.1.tgz",
+      "integrity": "sha512-yLAMQs+k0b2m7cVxpS1VKJVvoz7SS9Td1zss3XRwXj+ZDH00RJgnuLx7E44wx02kQLrdM3aOOy+FpzS7+8OizA=="
     },
     "for-each": {
       "version": "0.3.3",
@@ -2200,15 +2205,6 @@
       "resolved": "https://registry.npmjs.org/psl/-/psl-1.8.0.tgz",
       "integrity": "sha512-RIdOzyoavK+hA18OGGWDqUTsCLhtA7IcZ/6NCs4fFJaHBDab+pDDmDIByWFRQJq2Cd7r1OoQxBGKOaztq+hjIQ=="
     },
-    "follow-redirects": {
-      "version": "1.15.1",
-      "resolved": "https://registry.npmjs.org/follow-redirects/-/follow-redirects-1.15.1.tgz",
-      "integrity": "sha512-yLAMQs+k0b2m7cVxpS1VKJVvoz7SS9Td1zss3XRwXj+ZDH00RJgnuLx7E44wx02kQLrdM3aOOy+FpzS7+8OizA=="
-    },
-    "inherits": {
-      "version": "2.0.4",
-      "resolved": "https://registry.npmjs.org/inherits/-/inherits-2.0.4.tgz",
-      "integrity": "sha512-k/vGaX4/Yla3WzyMCvTQOXYeIHvqOKtnqBduzTHpzpQZzAskKMhZ2K+EnBiSM9zGSoIFeMpXKxa4dYeZIQqewQ=="
     "public-encrypt": {
       "version": "4.0.3",
       "resolved": "https://registry.npmjs.org/public-encrypt/-/public-encrypt-4.0.3.tgz",


druiz0992 (changes requested):
- add README how to deploy client
- I’d move start-client inside client
- I think you need a `client.env` where one can configure parameters as Web3Url and proposerUrl. Also, i would add the location of s3 bucket in a variable in case it changes.

daveroga:
Added suggested changes

nightfall-client/README.md
```diff
@@ -8,7 +8,7 @@
 
 # nightfall-client
 
-This code generates a containerised application that can be used to interact with Nightfall_2 via
+This code generates a containerised application that can be used to interact with Nightfall_3 via
```
druiz0992, line 11: Let's call this Polygon Nightfall from now on

nightfall-client/README.md
```diff
@@ -19,79 +19,39 @@ nightfall-client requires a number of services to be present for it to work. The
 instructions explain how to run all of these up, in a similar 'developer' mode with local file
```
druiz0992, line 19: This README is intended for outside people trying to use the SDK. So, it needs to be very clear. Wha'ts developer mode?

nightfall-client/README.md
```diff
@@ -19,79 +19,39 @@ nightfall-client requires a number of services to be present for it to work. The
...
 system bindings.
 
-If you just want to run Nightfall_2 then an easier way is to use
-[nightfall-scripts](https://github.com/EYBlockchain/nightfall-scripts), which does a similar job but
-uses docker images that do not rely on your local file system and therefore requires no other code
-repositories to be downloaded.
+If you just want to run Nightfall_3 then an easier way is to use
```
druiz0992, line 22: Polygon Nightfall

nightfall-client/README.md
```diff
@@ -19,79 +19,39 @@ nightfall-client requires a number of services to be present for it to work. The
...
-
-Make sure you put all of these repositories under a common root in your file system because the
-docker-compose yaml assumes they are accessible via `../<application>`. Then, in the
-`nightfall-client` root directory, locally install the files you'll need for testing via:
-
-```sh
-$ npm i
+ETH_NETWORK=goerli or mainnet
+BLOCKCHAIN_URL=your web3 url provider to access the blockchain
+CIRCUIT_FILES_URL=url of the repository for the circuit files
```
druiz0992, line 37: You will need to provide this URL. A external user doesn't know what this is.

nightfall-client/client.env
```diff
@@ -0,0 +1,4 @@
+ETH_NETWORK=goerli
+BLOCKCHAIN_URL=your web3 url provider to access the blockchain
```
druiz0992, line 2: Can you provide an example to understand format

druiz0992, line 2: Don't we need a proposer URL?

docker-compose.client.yml
```diff
@@ -0,0 +1,72 @@
+version: '3.5'
```
druiz0992, line 1: There are some requirements on the docker-compose version required to be installed. Can you clarify?

nightfall-client/client.env
```diff
@@ -0,0 +1,4 @@
+ETH_NETWORK=goerli
```
druiz0992, line 1: Call this file something else `client-example.env`. Add ´client.env` to .gitignore. In the instructions, then specify that one should copy `client-exmaple.env` to `client.env` and add their own information

nightfall-client/README.md
```diff
@@ -19,79 +19,39 @@ nightfall-client requires a number of services to be present for it to work. The
...
-- [nightfall-deployer](https://github.com/EYBlockchain/nightfall-deployer)
-- [nightfall-timber](https://github.com/EYBlockchain/nightfall-timber)
-
-Make sure you put all of these repositories under a common root in your file system because the
-docker-compose yaml assumes they are accessible via `../<application>`. Then, in the
-`nightfall-client` root directory, locally install the files you'll need for testing via:
-
-```sh
-$ npm i
+ETH_NETWORK=goerli or mainnet
```
druiz0992, line 35: It needs to be clear that this client is to connect to existing testnet and mainnet deployments. Also, can you comment on how to deploy 2+ clients? And if one wants to launch a client vs a local environment. How would this be done?

druiz0992 (changes requested):
Left my comments in the review

nightfall-client/client.env
daveroga: The proposer url now is got in the `nightfall-client` inside `transfer` or `withdraw` when an off-chain transaction. It gets the url from the proposer structure of the contracts.

nightfall-client/README.md
```diff
@@ -19,79 +19,39 @@ nightfall-client requires a number of services to be present for it to work. The
...
 
-Build the docker images (you only need do this once unless you makes changes to package.json). To do
-this you will need a personal access token with repo permissions. This is because one of the build
-stages of `zokrates-worker` needs to pull a GPR package:
+To run the script with existing images:
 
-```sh
-$ docker-compose build --build-arg GPR_TOKEN=<paste your token here>
+```
+./start-client -c
```
druiz0992, line 44: How do you stop a client?

docker-compose.client.yml
daveroga: I choose the same version and format of the docker-compose files of the root folder for other purposes.

druiz0992: What i meant is that there should be a requirements section in the README where we explain what sw is needed to run this. For example, we need nodeJS v xxx, docker-compose v.xxx,... Again, this README needs to be enough for a user to spin a client without the need to go to the main README. All information required should be here.

docker-compose.client.yml
```diff
@@ -0,0 +1,72 @@
...
+# Use this script for running up nightfall_3 client in 'developer' mode with local
+# bindings.  See the readme for more information.
+services:
+  client1:
+    image: ghcr.io/eyblockchain/nightfall3-client:latest
```
druiz0992, line 6: i think you should change the image name.... its strange that its called `eyblockchain`

nightfall-client/start-client
```diff
@@ -0,0 +1,71 @@
...
+
+usage()
+{
+  echo "Usage:"
+  echo "  -c or --client; for client service"
+  echo "  -d or --dev; to bind mount the filesystem and use it for development"
+}
+
+# select a Geth or Ganache client
+if [ -z "$1" ]; then
```
druiz0992, line 23: What is this for?

nightfall-client/README.md
daveroga: With ctrl+c as start-nightfall. It shuts down the containers because the `trap` in the `start-client` script catch the event and runs a docker-compose down.

nightfall-client/start-client
```diff
@@ -0,0 +1,71 @@
...
+  exit 1
+fi
+
+VOLUME_LIST=$(docker volume ls -q)
+FILE=
+
+usage()
+{
+  echo "Usage:"
+  echo "  -c or --client; for client service"
```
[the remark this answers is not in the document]
daveroga: True. Changed to always run this part by default.

nightfall-client/start-client
daveroga: Removed. Expected at least 1 argument. With -c by default and after removal then it has no sense.

nightfall-client/README.md
daveroga: Some comments in README

docker-compose.client.yml
daveroga: Yes, but this applies to all images that we are building now in the Dockerfiles in the root folder and not only this PR.
So we have to do it in all images I think.
- ghcr.io/eyblockchain/nightfall3-deployer:latest
- ghcr.io/eyblockchain/nightfall3-worker:latest
- ghcr.io/eyblockchain/nightfall3-client:latest
- ghcr.io/eyblockchain/nightfall3-optimist:latest
- ghcr.io/eyblockchain/nightfall3-proposer:latest
- ghcr.io/eyblockchain/nightfall3-user-local:latest

daveroga:
Added `TRANSFER_VALUE` in config to be able to pass the amount to be transferred in tests.
```
transferValue: process.env.TRANSFER_VALUE || 10,
```

druiz0992 (approved):

nightfall-client/start-client
```diff
@@ -0,0 +1,71 @@
...
+}
+
+# select a Geth or Ganache client
+if [ -z "$1" ]; then
+  usage
+  exit 1
+fi
+
+# delete env file
+rm -f ${ENV_FILE}
```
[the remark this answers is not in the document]
daveroga: This is from copy paste. Removed.

nightfall-client/start-client
```diff
@@ -0,0 +1,71 @@
...
+if [ -f client.env ]; then
+  grep -v '^#' client.env
+  # Export env vars
+  export $(grep -v '^#' client.env | xargs)
+else
+  echo "client.env does not exist. You need to define the client.env with the needed variables to run the client."
+  exit 1
+fi
+
+VOLUME_LIST=$(docker volume ls -q)
```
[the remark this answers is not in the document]
daveroga: The docker volume gets initialized with `start-client` in the docker-compose `docker-compose.client.yml`

nightfall-client/src/services/state-sync.mjs
```diff
@@ -44,14 +53,73 @@ const syncState = async (fromBlock = 'earliest', toBlock = 'latest', eventFilter
...
+    await Promise.all(
+      files.map(async f => {
+        if (f) {
+          try {
+            await downloadFile(
+              `${baseUrl}/build/contracts/${f.split('  ')[1]}`,
+              `${config.CONTRACT_ARTIFACTS}/${f.split('  ')[1]}`,
+            );
+          } catch (e) {
+            console.error(`ERROR downloading ${f.split('  ')[1]}`);
```
israelboudoux, line 111: Here we just log that an error happened. Is the download of a contract required for the app to work as expected? if so, I think we should rethrow the error.

test/utils.mjs
```diff
@@ -339,9 +339,31 @@ export const expectTransaction = res => {
...
+        res = await nf3.deposit(ercAddress, tokenType, value, tokenId, fee);
+        count = 0;
+      } catch (e) {
+        if (
+          e.message.includes('nonce too low') ||
+          e.message.includes('replacement transaction underpriced')
+        ) {
+          count -= 1;
+          logger.debug(`Transaction failed. Trying again...${count} tries left`);
+          await new Promise(resolving => setTimeout(resolving, 10000));
```
israelboudoux, line 356: I think we could create a utility function into some of our `Utils` js files to have a timeout function where we just pass over the desired `timeout` (e.g. `waitForTimeout(timeoutInMs)`), and replace it everywhere this construction is being used.

test/client/erc20.test.mjs
```diff
@@ -0,0 +1,124 @@
...
+    await nf3Users[1].init(mnemonics.user2);
+    erc20Address = '0x499d11e0b6eac7c0593d8fb292dcbbf815fb29ae'; // MATIC
+
+    stateAddress = await nf3Users[0].stateContractAddress;
+    web3Client.subscribeTo('logs', eventLogs, { address: stateAddress });
+  });
+
+  describe('Deposits', () => {
+    it('should increment the balance after deposit some ERC20 crypto', async function () {
+      const before = (await nf3Users[0].getLayer2Balances())[erc20Address]?.[0].balance || 0;
```
israelboudoux, line 50: Please, rename `before` and `after` variables to `balanceBefore` and `balanceAfter`, respectively.

test/client/erc20.test.mjs
```diff
@@ -0,0 +1,124 @@
...
+        tokenType,
+        transferValue,
+        tokenId,
+        fee,
+      );
+      eventLogs = await web3Client.waitForEvent(eventLogs, ['blockProposed']);
+      // wait a bit to blockProposed be processed by the client
+      await new Promise(resolving => setTimeout(resolving, 10000));
+      const after = (await nf3Users[0].getLayer2Balances())[erc20Address]?.[0].balance || 0;
+      remainingBalance = txPerBlock * transferValue - (after - before);
```
israelboudoux, line 67: It isn't an issue, but I'm not sure if I understood the remainingBalance calculation, wouldn't it be `remainingBalance = balanceAfter - txPerBlock * transferValue`?

test/client/erc20.test.mjs
```diff
@@ -0,0 +1,124 @@
...
+        nf3Users[1].zkpKeys.compressedPkd,
+        fee,
+      );
+      expect(res).to.be.equal(200);
+
+      const after = (await nf3Users[0].getLayer2Balances())[erc20Address]?.[0].balance || 0;
+      logger.debug(after, before, before + remainingBalance - transferValue);
+      expect(after).to.be.lessThan(before);
+      expect(after).to.satisfy(balance => {
+        return balance < before || balance === before + remainingBalance - transferValue;
```
israelboudoux, line 93: You're relying on that the sequence of the tests will be preserved by the testing library when running them, I'm not sure if this is the best approach, what if someone changes the ordering of the test methods?

zokrates-worker/src/index.mjs
```diff
@@ -8,10 +10,80 @@ import logger from './utils/logger.mjs';
...
 const {
   MPC: { MPC_PARAMS_URL },
+  DEPLOYMENT_FILES_URL: { DEFAULT_CIRCUIT_FILES_URL },
 } = config;
 
+const { ETH_NETWORK, CIRCUIT_FILES_URL } = process.env;
+
+const finished = promisify(stream.finished);
+
+const downloadFile = async (fileUrl, outputLocationPath) => {
```
israelboudoux, line 20: This method is duplicated, it is a serious candidate to go to a library like `httpUtils`. What do you think?

zokrates-worker/src/index.mjs
```diff
@@ -8,10 +10,80 @@ import logger from './utils/logger.mjs';
...
+const checkCircuitsOutput = async () => {
+  let env = '';
+  switch (ETH_NETWORK) {
+    case 'goerli':
+      env = 'testnet';
+      break;
+    case 'mainnet':
+      env = 'production';
+      break;
+    default:
```
israelboudoux, line 41: The `default` case can be removed here, the `env` variable is already blank.

zokrates-worker/src/index.mjs
```diff
@@ -8,10 +10,80 @@ import logger from './utils/logger.mjs';
...
+            fs.mkdirSync(`${outputPath}/${circuit}`);
+          }
+
+          try {
+            await downloadFile(
+              `${baseUrl}/proving_files/${circuit}/${f.split('  ')[1]}`,
+              `${outputPath}/${circuit}/${f.split('  ')[1]}`,
+            );
+          } catch (e) {
+            console.error(`ERROR downloading ${f.split('  ')[1]}`);
```
israelboudoux, line 74: The same as the comment I added before, if the contract strictly required for the app to run well, we should let the error go up.

israelboudoux (changes requested):
Hey @daveroga, I think you've missed some comments, please take a look on them.

israelboudoux (commented):
Looks better now, thanks!

daveroga (commented):
Reviewed changes and suggestions

israelboudoux (approved):
Looks good to me!

PR #710, opened 2022-06-02
Adding a synchronizer and refactoring stuff

signorecello:
So the idea is to have a root folder `apps` holding all the... hum... apps. Like `proposer`, `challenger`, and the new `synchronizer`. This new service simply exists by now and doesn't do much, but the idea is to expose some endpoints to get information from the underlying optimist without exposing the optimist itself


apps/proposer/src/app.mjs
```diff
@@ -1,10 +1,21 @@
...
+import Nf3 from '../../../cli/lib/nf3.mjs';
+
+const { signingKeys } = config.TEST_OPTIONS;
+const environment = config.ENVIRONMENTS[process.env.ENVIRONMENT] || config.ENVIRONMENTS.localhost;
+const { PROPOSER_PORT } = config;
 
 const app = express();
+const nf3 = new Nf3(signingKeys.proposer1, environment);
+
+app.set('nf3', nf3);
```
druiz0992, line 17: I love this `app.set('nf3', nf3);`. I wish i knew how to program better.

apps/synchronizer/src/app.mjs
```diff
@@ -0,0 +1,38 @@
...
+import { contracts } from './routes/index.mjs';
+import Nf3 from '../../../cli/lib/nf3.mjs';
+import startSynchronizer from './synchronizer.mjs';
+
+const { signingKeys } = config.TEST_OPTIONS;
+const environment = config.ENVIRONMENTS[process.env.ENVIRONMENT] || config.ENVIRONMENTS.localhost;
+const { SYNCHRONIZER_PORT } = config;
+
+const app = express();
+const nf3 = new Nf3(signingKeys.synchronizer, environment);
```
druiz0992, line 15: Why does synchronizer require a key? It doesn't need to sign anything

config/default.js
```diff
@@ -51,7 +51,8 @@ module.exports = {
   OPTIMIST_HOST: process.env.OPTIMIST_HOST || 'optimist',
   OPTIMIST_PORT: process.env.OPTIMIST_PORT || 80,
   PROPOSER_HOST: process.env.PROPOSER_HOST || 'proposer',
-  PROPOSER_PORT: process.env.PROPOSER_PORT || 8080,
+  PROPOSER_PORT: process.env.PROPOSER_PORT || 8090,
```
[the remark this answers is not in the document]
signorecello: I had some conflicts since client1 is running on the same port (check docker-compose). Now, why didn't I get this conflict before? I think only the Pope may know why. 

Anyway, changing ports fixes this

druiz0992 (changes requested):
I left some questions

config/default.js
```diff
@@ -226,6 +227,9 @@ module.exports = {
       liquidityProvider:
         process.env.LIQUIDITY_PROVIDER_KEY ||
         '0xfbc1ee1c7332e2e5a76a99956f50b3ba2639aff73d56477e877ef8390c41e0c6',
+      synchronizer:
+        process.env.SYNCHRONIZER_KEY ||
```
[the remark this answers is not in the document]
signorecello: Well we shouldn't need, but nf3 expects one. I thought to change nf3 to somehow allow starting without one but since the SDK will be refactored anyway, that'd probably only create conflicts.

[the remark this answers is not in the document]
druiz0992: Does it need to be a valid key? can we put garbage, or an empty variable?

apps/proposer/config/default.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  SIGNING_KEY:
+    process.env.SIGNING_KEY || '0x4775af73d6dc84a0ae76f8726bda4b9ecf187c377229cb39e1afa7a18236a69d',
+  PROPOSER_HOST: process.env.PROPOSER_HOST || 'localhost',
+  PROPOSER_PORT: process.env.PROPOSER_PORT || 8092,
+  CLIENT_HOST: process.env.CLIENT_HOST || 'localhost',
```
druiz0992, line 6: Why do we need a CLIENT?

druiz0992, line 6: Why do we need a CLIENT?

signorecello: Because `nf3` expects one to be passed to the constructor. Again, we can then tweak the SDK to not need it, but doing so now would cause conflicts in the future. We can add an issue to the SDK repo though

daveroga (approved):
ILGTM

PR #715, opened 2022-06-02
Luizoamorim/import and export commitments

luizoamorim:
### Export
- [x] Export transactions and commitments from indexedDB, add them for a file and allow the user to do a download of this backup.
- [x] When saving the file, you should ask where to save the file
  **- For this is necessary the user for itself change this setting in its browser settings.**

#### Test
- [x] Unit tests

image

### Import
- [x] Upload the backup file.
- [x] Allow the user to insert an old mnemonic.
- [x] Back this objects to indexedDB.
- [x] unless you select a file, import button should be disabled
- [x] If file is incorrect format, it should be handled better than just crashing
- [x] If mnenonic is incorrect, it should be handled better than just crashing

#### Test
- [x] Unit tests

### Verification between commitments compressedKeys and derived keys
- [x] function
- [x] web interface

#### Test
- [x] unit test

#### Rebase
- [x] rebase my PR with master

#### Bugs found
- [x] Fix the problem for the following flow:
 - User insert a mnemonic to recover an old wallet;
 - User upload a file;
 - User try to recover;
 - A modal shows saying that the mnemonic derived keys don't match with the commitments compressed_pkd in backup file.
 - So the main modal to generate or recover a wallet is showed again;
 - Now, when the user insert a mnemonic and a file, the button for recover doesn't works more. Is some problem when passing the props between modals.
- [x] After the wrong flow above. When the app comeback to the main modal. Is necessary not just delete the indexedDB nightfall_commitments database, but is necessary to drop this database.
image


luizoamorim:
# With this last commit is possible to test the new feature

## Steps
- If you don't have the last mnemonic saved;
  - Create a new wallet;
  - Save the mnemonic;
  - Do some transactions;
  - If you already have your mnemonic go directly for transactions page.
- Go to transactions page;
- Click in export transactions button;
- Save the file;
- Clear the database or go to other browser;
- Open the wallet;
- In the Generate Mnemonic modal, click in the button to recover wallet;
- In recover wallet modal:
  -  Insert your 12 words;
  - Upload the file that you've downloaded;
  - Click in recover wallet;

After this you should have your commitments available.

druiz0992 (changes requested):
Some comments:
- When saving the file, you should ask where to save the file. (agree)
- When importing the file:
    -  unless you select a file, import button should be disabled (agree)
    - If file is incorrect format, it should be handled better than just crashing (agree)
    - If mnenonic is incorrect, it should be handled better than just crashing (agree)
    - Why can't you reuse the same modal to enter mnemonic for importing than for creating new wallet? (**For this we'll need to follow the new designs, so I don't know if make sense to spend time in this now. Unless if is necessary to already have something better for the user in this moment**)

@druiz0992 I give some comments here.

wallet/src/static/supported-token-lists/supported-tokens-local.ts
```diff
@@ -5,7 +5,7 @@ const tokensList = {
       name: 'Ether - ERC20 Mock',
       symbol: 'ETH',
       decimals: 9,
-      address: '0x17E1009e6Eab141615811104795D98c316d97AC0',
+      address: '0x9b7bD670D87C3Dd5C808ba627c75ba7E88aD066f',
```
signorecello, line 8: I'm confused, why is this?

signorecello (changes requested):

wallet/src/components/Transactions/index.jsx
```diff
@@ -157,40 +158,63 @@ const Transactions = () => {
     setTxs(mappedTxs);
   }, delay);
 
+  const handleExportIndedexDB = async () => {
+    const exportedDB = await exportIndexdDB('nightfall_commitments');
```
israelboudoux, line 162: Hey @luizoamorim, instead of exporting the data straight away, what do you think of confirming before doing it?

wallet/src/views/wallet/index.jsx
```diff
@@ -32,6 +38,192 @@ const { ethereum } = global;
...
+    });
+  };
+
+  const recoverWallet = async () => {
+    let mnemonic = await concatAllWords();
+    await deriveAccounts(mnemonic, DEFAULT_ACCOUNT_NUM);
+
+    /**
+     * TODO
+     * Conditional to verify if the commitments keys match with the keys derivated
```
israelboudoux, line 100: Hey @luizoamorim, is this commented code still needed?

israelboudoux:
I would add some unit tests, for these cases only the happy paths. We should start writing unit tests as soon as possible.

luizoamorim:
The last commit insert some unit tests for this flow. Was necessary use a framework like Jest to test some stuffs related with Blob files and IndexedDB.

wallet/src/components/Transactions/index.jsx
luizoamorim: Like a modal after the user click in the button? @israelboudoux

wallet/src/static/supported-token-lists/supported-tokens-local.ts
luizoamorim: I was giving an error with the old address. So Ilyas asked me to change for this new address. I understood some new mock smart contract was deployed.

luizoamorim:
The last commit add the function and the test to verify if all the commitments match with one of the derived keys.

druiz0992 (changes requested):
When generating a new wallet, it takes me to import wallet modal.

luizoamorim:
# With the last commit you can test
## Steps
#### Export
- Export your commitments and transactions in Transactions page'
- Save your file.
#### Import
- Select recover waller in the initial modal
- insert a valid mnemonic
- upload a valid json file
- click in recover
#### Verification
- if the keys derived from the mnemonic inserted don't match with the commitments preimage compressedPkd an message will be showed for the user and the user can do the flow again.
- if the keys match a successfully message will be showed.
- user needs to insert a valid mnemonic and a valid json file

luizoamorim:
@druiz0992 @israelboudoux 
I just found some bugs that I added to the description of this PR. So I think is better I solve this first and after we do the merge.

druiz0992 (changes requested):
When saving the file, you should ask where to save the file. This functionality is still missing

druiz0992:
> @druiz0992 @israelboudoux I just found some bugs that I added to the description of this PR. So I think is better I solve this first and after we do the merge.

@israelboudoux  I cant reproduce the issue you report

luizoamorim:
## package-lock.json
The error that eslint was triggering was because of a problem in package-lock.json format.
So I deleted it and ran npm install again. I don't know if it is the best approach, but was what I found to solve the problem.

luizoamorim:
## Bug fixes
To fix the follow bug:
image
I'm refreshing the browser (window.location.reload()) after the user click in OK.
Again, I think it is not the best approach, but is something that we can refactor later.

luizoamorim:
## Bug fixes
To fix the follow bug:
image
I've created a function called clearSiteData that clear the localStorage and SessionStorage and delete the indexedDB database.

wallet/tests/unit/useCases/CommitmentsBackup/export.test.js
```diff
@@ -0,0 +1,32 @@
+// 0: {table: 'blocks', rows: Array(0)}
```
israelboudoux, line 1: I think these comments aren't needed right?

wallet/tests/unit/useCases/CommitmentsBackup/import.test.js
```diff
@@ -0,0 +1,95 @@
...
+      { key: '2', value: { _id: '2', value: '2' } },
+      { key: '3', value: { _id: '3', value: '3' } },
+    ],
+  },
+];
+
+describe('Tests about file', () => {
+  test('Should convert a blob file in a object', async () => {
+    const file = new Blob([JSON.stringify(mockObject)], { type: 'text/plain' });
+    expect(file).not.toEqual(mockObject);
```
israelboudoux, line 39: You don't need this, since `file` is an instance of Blob and `mockObject` is an array of objects, it is straightforward to see they aren't equal.

wallet/tests/unit/useCases/CommitmentsBackup/import.test.js
```diff
@@ -0,0 +1,95 @@
...
+      // store the result of opening the database in the db variable.
+      // This is used a lot below
+      const db = request.result;
+      const transaction = db.transaction([OBJ_STORE_NAME], 'readwrite');
+      const objectStore = transaction.objectStore(OBJ_STORE_NAME);
+      const objectStoreRequest = objectStore.get('1');
+      objectStoreRequest.onsuccess = function () {
+        // report the success of our request
+        const objResult = objectStoreRequest.result;
+        expect(objResult.key).toEqual('1');
```
israelboudoux, line 91: Here you could test another attribute (e.g. `value._id`) instead of the `key`, since you are searching by `key`, it is expected that it should be equal.

israelboudoux (changes requested):
Hey @luizoamorim, I think we could move `CommitimentsBackup` to another place like `utils`. Using the folder `useCases` sounds a bit strange, I wouldn't use it, because Use Cases are normally related to functions a User can perform but they are usually grouped as `methods` in `Classes`.

wallet/tests/unit/useCases/CommitmentsBackup/export.test.js
luizoamorim: Right! Removed!

wallet/tests/unit/useCases/CommitmentsBackup/import.test.js
luizoamorim: Really. I don't know what I left it there. Removed line 39.

wallet/tests/unit/useCases/CommitmentsBackup/import.test.js
luizoamorim: Ok. Could be better. Changed.

luizoamorim:
> Hey @luizoamorim, I think we could move `CommitimentsBackup` to another place like `utils`. Using the folder `useCases` sounds a bit strange, I wouldn't use it, because Use Cases are normally related to functions a User can perform but they are usually grouped as `methods` in `Classes`.

I'm understanding your vision focused by object oriented approach. I'll move it and if in the future we find some better way we change again. Tks.

druiz0992 (approved):

wallet/src/views/wallet/index.jsx
```diff
@@ -14,77 +14,427 @@ import Header from '@Components/Header/header.jsx';
...
+
+    new Promise((resolve, reject) => {
+      let flag = false;
+      let myIndexedDB;
+      // const request = indexedDB.open('MyDatabase', 1);
+
+      while (!flag) {
+        try {
+          myIndexedDB = indexedDB.open('nightfall_commitments', 1);
+          flag = true;
```
israelboudoux, line 134: Hey @luizoamorim, we can remove the `flag` variable and just use a `break` statement, it has the same effect you want.

wallet/src/views/wallet/index.jsx
```diff
@@ -14,77 +14,427 @@ import Header from '@Components/Header/header.jsx';
...
+
+      myIndexedDB.onsuccess = () => {
+        const OBJ_STORE_NAME = 'keys';
+        const db = myIndexedDB.result;
+        let transaction;
+        flag = false;
+        while (!flag) {
+          transaction = db.transaction([OBJ_STORE_NAME], 'readwrite');
+          if (transaction) {
+            flag = true;
```
israelboudoux, line 152: The `flag` variable can be removed here as well and use the `break` statement.

wallet/src/views/wallet/index.jsx
```diff
@@ -14,77 +14,427 @@ import Header from '@Components/Header/header.jsx';
...
+
+    if (!backupFile) {
+      return;
+    }
+
+    setIsMnemonicValid(isValid);
+    await deriveAccounts(mnemonic, DEFAULT_ACCOUNT_NUM);
+
+    new Promise((resolve, reject) => {
+      let flag = false;
```
israelboudoux, line 127: The code in this promise is so long that I think it deserves a method just for it.

wallet/src/views/wallet/index.jsx
```diff
@@ -14,77 +14,427 @@ import Header from '@Components/Header/header.jsx';
...
 
 const { ethereum } = global;
 
-/**
-This is a modal to detect if a wallet has been initialized
-*/
-
 function WalletModal(props) {
   const [, , deriveAccounts] = useContext(UserContext);
   const [screenMnemonic, setScreenMnemonic] = useState();
```
israelboudoux, line 42: Hey @luizoamorim, I think it is the first time I see this file, I'm sorry if I missed it, but I have a concern about it since it is storing the mnemonic in attributes/variables, which is a vulnerability. Is there how to get rid of these fields which store them?
This kind of sensitive information should be very very short-lived, I mean, as soon as we use it, we must blank/nullify the variable/attribute so that it doesn't hang on the memory.

luizoamorim: At this first moment I don't see other way to handle this. Because the user will generate it from the screen and we need to manipulate this info. 
What I did, in the last commit, was clear the variable state as soon that its is used. Anyways, after the modal is closed the variable is lost, because it is a local variable.
Anyways we can open an other issue to investigate a better way to track this. Maybe we can use an env var with some word to encrypt and decrypt this field between it is handled.

israelboudoux: > 

Yes, I understand that the object will be garbage collected at some moment, but the value of the variables/attributes will still hold in memory until their respective memory addresses are used by another processes.

israelboudoux (approved):
Hey @luizoamorim, please, create the ticket with the pending issues when possible.

PR #716, opened 2022-06-02
avoid storing redundant tx from blockpropose event

### LINKED ISSUE #667: Wallet to stop saving transactions it doesn't care about.
Since finalise withdrawal now only requires the transaction index and a siblingPath, there is no longer any need for all other transactions in a block to be stored.

We should remove this to save storage overhead in the browser.

LijuJoseJJ:
this PR for now fixes wallet/browser and nightfall_client
- only storing relevant transaction
- also only storing block object for which any transaction is stored to db

fixes #667

### DIFF
--- a/nightfall-client/src/event-handlers/block-proposed.mjs
+++ b/nightfall-client/src/event-handlers/block-proposed.mjs
@@ -98,7 +98,6 @@ async function blockProposedEventHandler(data, syncing) {
   });
 
   await Promise.all(dbUpdates).then(async updateReturn => {
-
     // only save block if any transaction in it is saved/stored to db
     const saveBlockToDb = updateReturn.map(d => d[0]);
     if (saveBlockToDb.includes(true)) {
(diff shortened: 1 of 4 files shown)


druiz0992 (changes requested):
You will need to rebase with master and failing tests should now pass

druiz0992 (commented):
Do we need to apply these changes to client-nightfall?

LijuJoseJJ:
> Do we need to apply these changes to client-nightfall?

yes I believe, that why I refactored client as well in this PR

druiz0992 (commented):
Can you think of how we can test this?

LijuJoseJJ:
> Can you think of how we can test this?

the best way to test is (and the way i did is) to customise `erc20.test.mjs` for all scenario, I also run wallet app for case where client is not hosting a particular user (i.e keys are not store in client).

israelboudoux:
Hey @LijuJoseJJ, are you still working on this PR?

LijuJoseJJ:
> Hey @LijuJoseJJ, are you still working on this PR?

didn't notice merge conflict, will fix by tomorrow

druiz0992 (approved):

daveroga (approved):
LGTM

PR #758, opened 2022-06-07
David/optimist standalone

druiz0992:
Launch standalone optimist.  one issue has been found and resolve. Optimist cannot sync if there are existing blocks at the moment of launching optimist. It seems the issue is caused because queues are double paused. To prevent this, before pausing a queue, we now check the state.

To test it:
```
./setup-nightfall.sh
./start-nightfall -g -d -s
```

Then go to `nightfall-optimist` and copy `optimist.copy.env` to `optimist.env`

Launch test `./test-optimist.sh`

A tmux session will be created, `optimist-test`, with test logs.


israelboudoux (approved):
Looks good to me @druiz0992, the only thing that I don't like much is the extensive use of `sh` files, maybe in the future we could try to migrate to use javascript, so we would be uniform with the project and everyone is kind of used to it.

druiz0992:
@israelboudoux sh files are needed to launch different containers. As soon as we have the client, optimist stand-alone and SDK PRs approved, we can writes test in jabascript

daveroga (approved):
Looks good to me

PR #761, opened 2022-06-13
David/ping pong

druiz0992:
Fixes ping pong test 
- sets correct proposer port and host
- fixes number of transfers to be equal to number of transactions per block - 1
- github action now checks that test passes rather than exit status
- manages unused exports in `apps/`, which makes eslint check fail


daveroga (approved):

signorecello (approved):
LGTM

Westlad (approved):

PR #766, opened 2022-06-17
Correct on-chain merkle hashing

IlyasRidhuan:
This PR updates some issues with the on-chain hashing of `hashTransactionHashes`. The issue arose because the initial hashing of the dynamic array assumed the memory layout contained zero bytes after the `leaves` data.

This isn't guaranteed, so we now instead copy the `transactionsHashes` into a structure beginning at the canonical zero pointer. Batching does not seem to be needed as we don't run into stack issues even when hashing 32 transactions.


daveroga (approved):
LGTM

druiz0992 (approved):

PR #768, opened 2022-06-21
feat: added ropsten config info

druiz0992:
This PR adds some config values for Ropsten PoS test network, and its used to test future Ethereum mainnet.

Deployment is tested on AWS infrastructure. This change here should not modify other deployments


signorecello (approved):
LGTM!

PR #780, opened 2022-06-29
Westlad/multisig

Westlad:
This fixes [issue D](https://github.com/EYBlockchain/nightfall_3_private/issues/126).

The admin application that runs in the administrator container has been heavily modified so that it now works with a multisig contract and `onlyOwner` actions now require a number of approvals before they can be executed.  Full details and information on testing are contained in the nightfall-administrator [README](https://github.com/EYBlockchain/nightfall_3/blob/westlad/multisig/nightfall-administrator/README.md).

Note that inclusion of the multisig slightly breaks contract upgradability because a multisig is not compatible with the truffle upgrade plugin. However a simple work around is to disable the multisig temporarily and to use an ephemeral key pair for the upgrade before switching back to the multisig.  This can be achieved by calling the relevant contract's `transferOwnership` function using the multisig.


IlyasRidhuan (approved):

Westlad:
@druiz0992 I've now added instructions to the README for removing the multisig .

PR #786, opened 2022-06-30
David/adversary test2

druiz0992:


druiz0992:
@ChaitanyaKonda, thanks for the feedback. We have currently prioritized the fact of having few actors and thus having an adversary + challenger. There is an additional functionality implemented in https://github.com/EYBlockchain/nightfall_3/commit/eb229bfbb9c24193585538c4a34fb8963618b0b4 that allows us to maintain  adversary producing blocks even in the event of a rollback to test deep rollbacks. Let me know if this satisfies your concerns.

ChaitanyaKonda (approved):

PR #788, opened 2022-07-01
Refactoring mocks

### LINKED ISSUE #746: Mock contracts should be outside nightfall deployer

signorecello:
Closes #746 

I did a bunch of things here:

- ~~I put ping-pong inside a mocha test. This makes it way easier for testing, as you only have to run ```npm run pin-pong``` instead of all those docker-compose scripts like ```./pong-nightfall```, ```pong-down```, ```ganache-standalone``` etc. This also removes a lot of duplicate code. Ping-pong now uses the same environment as all the other tests.~~
- Config files are now split in many files. I didn't use any criteria, simply wanted all the truffle configs to be in a different files. But this will allow us to break up config files in the future which is easier for development and to reduce merge conflicts
- Nightfall-deployer now will not deploy mocks. Future developments (like enterprise users) may not want mock contracts to be deployed in production, even if they were protected by an environment variable. Instead, ```start-nightfall``` will now run a script that deploys mocks if you use the `-dev` flag. When you run any of the tests (including ping-pong) the mocks will be there.


apps/proposer/src/proposer.mjs
```diff
@@ -14,13 +11,15 @@ export default async function startProposer(nf3, proposerBaseUrl) {
   // Mnemonic are only required for services connecting to a client that
   // can generate a compressed PKD.
 
-  await nf3.init(undefined, 'optimist');
+  await nf3.init(
```
druiz0992, line 14: propose doesnt need a mnemonic for now. And if it did, it may be better with a variable

signorecello:
Reverted the latest commit as I was heading a bit ahead of what's in the ticket. Ping-pong is not refactored, I'll cherry-pick it in some other branch later

signorecello:
Dropping this. The move to a multisig ownership makes impossible to approve the set restrictions as we "lose" the private key once the main deployment is done. We could do this by programatically interact with ```nightfall-administrator``` but that seems a bit of overhead we currently don't need.

We'll get back to this once we remove restrictions altogether.

Issue #574, opened 2022-03-12
Monitor delete and update operations in blocks collection 


druiz0992:
For prod and staging environments, we are using a publishing service that monitors modifications to `blocks collection` and updates those changes in a dynamoDb table. For this we are using a feature in momgoDb called `change streams` that return all updates to this collection.
The three operations that need to be monitored are insertions, updates and deletions.

Currently, only insertions are working.

This issue requests to include updates and deletions.

- Updates
Currently, updates are detected, but we need to make the updates to the dynamoDb table. 

- Deletions
Deletions are not supported by `change streams`, so we need to modify how ñ deletions are performed in `blocks collection`. One alternative may be to add a new field `delete` to the document, and set it to false by default during insertions and updates. When a deletion occurs, we could first update  this field to true, add a small delay, and then perform the deletion. 
With this I would expect that `change streams` will alert of the modification so that the publisher may remove the entry from the dynamoDb.
One concern I have is that the addition of this field may have an impact on other parts of the system. 

The second concern is that if the document is deleted, the publisher may not find this entry on the `blocks collection` because it has been already deleted.

PR #789, opened 2022-07-01
Delete and update operations in blocks collection #574

eugypalu:
Issue #574 

Added to the `deleteTreeByBlockNumberL2` function, the ability to update the block by setting `rollback = true`, so as to trigger the change stream, wait for a small delay and then delete the block.


druiz0992 (approved):

daveroga (approved):
LGTM

PR #790, opened 2022-07-01
David/adversary multiblock

### LINKED ISSUE #604: Fix adversary to work with proposing multiple L2 Blocks
`Block.build` works differently in Adversary's Optimist and Optimist and Optimist's logic is not replicable in. adversary
Hence proposing multiple blocks logic has to be written for adversary's optimist too.

druiz0992:
Fixes problem in leafCount computation when adversary generated multiple blocks

Fixes #604

### DIFF
--- a/test/adversary/transpile-adversary.mjs
+++ b/test/adversary/transpile-adversary.mjs
@@ -93,13 +93,13 @@ const transpileBlockBuilder = (_pathToSrc, _pathToInject) => {
   const reRoute = `const badBlock = createBadBlock({
       proposer,
       root: updatedTimber.root,
-      leafCount: updatedTimber.leafCount,
+      leafCount: timber.leafCount,
       nCommitments,
       blockNumberL2,
       previousBlockHash,
       frontier: updatedTimber.frontier,
     }, errorIndex);
-    this.localLeafCount = badBlock.leafCount;
+    this.localLeafCount = updatedTimber.leafCount;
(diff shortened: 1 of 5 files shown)


test/adversary/transpile-adversary.mjs
```diff
@@ -111,7 +111,7 @@ const transpileBlockBuilder = (_pathToSrc, _pathToInject) => {
       proposer,
       transactionHashes: transactions.map(t => t.transactionHash),
       transactionHashesRoot: this.calcTransactionHashesRoot(transactions),
-      leafCount: badBlock.leafCount,
+      leafCount: timber.leafCount,
```
IlyasRidhuan, line 114: I think this still needs to be `badBlock` - because that we may make a challenge with `incorrectLeafCount`?

druiz0992: i think you're right

druiz0992: @IlyasRidhuan should i remove the `getTreeByBlockNumberL2` update?

IlyasRidhuan: No I think it's better this way. `getTreeByBlockNumberL2` is idempotent so I feel better about it

daveroga (approved):
Looks good to me

IlyasRidhuan (approved):

PR #791, opened 2022-07-01
fix: Changing incorrect comment from '50 days' to '10 days'

aleksandar-veljkovic:
Changed time jump from 3600 * 24 * 10 seconds to 3600 * 24 * 50 seconds, as the comment suggests a time jump of 50 days.


druiz0992 (changes requested):
10 days was fine. Comment is what should be changed (from 50 to 10 days)

druiz0992 (approved):

PR #793, opened 2022-07-01
remove uneeded code in start-nightfall

druiz0992:
There are still remaining parts of code where old wallet persist. This PR removes those bits

- Removed old wallet code from start-nightfall
- Removed unused geth volumes from start-nightfall


daveroga (approved):
LGTM

PR #795, opened 2022-07-05
feat: adding ack to block produced websocket 🔌

### LINKED ISSUE #665: Make Proposer and Optimist sockets more aware of each other
Proposer and optimist currently operate on a "fire-and-forget" basis when blocks are created by optimist and sent to proposer.

Any blocks sent by optimist but missed by proposer requires a lengthy restart process in order to undo changes in the optimist database. This can be improved by a better "handshake" procedure between the two.

This could involve: waiting for acknowledgements, checking block numbers, checking liveness, etc.

signorecello:
It implements a basic ```ack``` response when nf3 receives and signs a block. This is not a very extensive implementation so I wouldn't really finish #665 for now. In order to close it, I would think of refactoring using rabbitmq or sockets.io or any other tools that give us better reliability, while providing us analytics and other data management tools.

The optimist will attempt to broadcast the block every 10 seconds, three times, if it doesn't get the ```ack``` from the proposer.


druiz0992 (changes requested):
We plan to use this in several places (proposer, challenger...). Can we include the functionality in a library in common-files maybe and just show in block proposer how to use it?

cli/lib/nf3.mjs
```diff
@@ -1008,22 +962,7 @@ class Nf3 {
...
-          // logger.debug('sent websocket ping');
-        }, WEBSOCKET_PING_TIME),
-      );
-      // and a listener for the pong
-      // connection._ws.on('pong', () => logger.debug('websocket received pong'));
-      logger.debug('websocket connection opened');
-      connection.send('challenge');
-    };
-    connection.onmessage = async message => {
+    this.rabbitmq.subscribe({ queue: 'block' }, message => {
```
daveroga, line 965: Here it seems it should subscribe to `challenge` instead `block`

daveroga: Good approach though with rabbitmq

signorecello:
I started looking into rabbitmq here, then after talking with @druiz0992 we decided to move that to another (draft) PR. I'll try to close this one asap

daveroga (approved):
Well organized now with some of the common functions in common-files :)

daveroga (approved):
Well organized now with some of the common functions in common-files :)

druiz0992 (approved):

PR #799, opened 2022-07-06
(wallet/nightfall bowser) refactor fetch circuits files

LijuJoseJJ:
this PR does a small refactor,  at time localrun out/program file and circuit abi.json file are now fetched locally instead frrom aws s3 bucket

### DIFF
--- a/wallet/package-lock.json
+++ b/wallet/package-lock.json
@@ -15844,7 +15844,7 @@
     "growly": {
       "version": "1.3.0",
       "resolved": "https://registry.npmjs.org/growly/-/growly-1.3.0.tgz",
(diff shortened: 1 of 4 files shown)


druiz0992:
@LijuJoseJJ do we need to do anything when we deploy new circuits so that wallet is updated?

LijuJoseJJ:
> @LijuJoseJJ do we need to do anything when we deploy new circuits so that wallet is updated?

@druiz0992 I did changes when run containers are run locally (i.e `hosted-utils-api-server` container exist).
I don't this deploy wallet have same behaviour.
@IlyasRidhuan please clarify

IlyasRidhuan:
We should look into modifying our deployment so that the compiled circuit files and abis from the `zokrates-worker` are saved into the S3 buckets directly - along with the proving keys (which are already part of it).

PR #802, opened 2022-07-07
Luizoamorim/add endpoint get commitments by compressedpkd

luizoamorim:
## Description
This pull request create a new endpoint to get all commitments by a compressed pkd

## Proposal 
I think that we want in the future move to use typescript and have more segregation in the responsibilities.

For this new endpoint I'm already start to change (if everybody agree) for a new architecture.

### Already implemented in this PR for this endpoint only
The route folder will continue with our endpoints 
The service folder gonna have the functions to make some logic for each use case
And (NEW) the repository folder is just responsible for make the communication with the database

image

### Future
When we start to use typescript. For the repositories, the idea is have an interface for each database model that will have all the data manipulation functions that this model should have.
And we'll have classes that implements these interfaces for each kind of database.

### Why?
With more segregation it will be easier to test our APIs.


nightfall-client/src/routes/commitment.mjs
```diff
@@ -82,6 +97,24 @@ router.get('/commitments', async (req, res, next) => {
...
+  logger.debug('commitment/all endpoint received GET');
+  const { compressedPkd } = req.query;
+  try {
+    const commitments = await getCommitmentsByCompressedPkd(compressedPkd);
+    res.json({ commitments });
+  } catch (err) {
+    logger.error(err);
+    next(err);
+  }
+});
```
imagobea, line 116: Since this is the beginning of a new era (I do not dislike the idea, let's see what the rest have to say), why don't we create a file `commitments` (note the `s` at the end). Then this route could simply be `/commitments`.

**What do you think?**

imagobea (changes requested):
Hi @luizoamorim nicely done!

As I've said in the comments I like the new approach. But I am also interested in hearing what the others have to say.

My only comment about this would be re the `repositories` dir. Why the name - did you research this?

For example when doing Python backend we used to call this layer simply `dbi` (database interface) - I've just rescued this from old projects! But perhaps "repositories" is +common in node + mongo? Let me know.

Thanks a lot :boom:

nightfall-client/src/repositories/commitment-repository.mjs
```diff
@@ -0,0 +1,48 @@
+import config from 'config';
```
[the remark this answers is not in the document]
luizoamorim: @daveroga Yes the idea is refactor:
Service folder will contain our controllers to do some logic (filter, verification ... )
And Repository is to have only communication with the database. 

At the beginning it will stay a little bit confuse. But in the final of the day the idea is have all the code refactored for the new structure.

nightfall-client/src/routes/commitment.mjs
```diff
@@ -12,10 +12,25 @@ import {
...
+    const commitment = await getCommitmentBySalt(salt);
+    if (commitment === null) logger.debug(`Found commitment ${commitment} for salt ${salt}`);
+    else logger.debug(`No commitment found for salt ${salt}`);
+    res.json({ commitment });
+  } catch (err) {
+    logger.error(err);
+    next(err);
+  }
+});
+
```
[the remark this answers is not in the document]
luizoamorim: I'll remove his. My bad.

nightfall-client/src/repositories/commitment-repository.mjs
```diff
@@ -0,0 +1,48 @@
...
+ * mnemonic coming from the SDK or Wallet.
+ * @returns all the commitments in the database.
+ * @author luizoamorim
+ */
+// export async function getAllCommitments() {
+//   initDb();
+
+//   const allCommitments = await db.collection(COMMITMENTS_COLLECTION).find({}).toArray();
+//   return allCommitments;
+// }
```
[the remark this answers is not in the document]
luizoamorim: @imagobea Totally. I forgot of remove.

luizoamorim:
> Hi @luizoamorim nicely done!
> 
> My only comment about this would be re the `repositories` dir. Why the name - did you research this?
> 
> For example when doing Python backend we used to call this layer simply `dbi` (database interface) - I've just rescued this from old projects! But perhaps "repositories" is +common in node + mongo? Let me know.
> 
> Thanks a lot boom

@imagobea 
So repositories is based in Service-Repository pattern. The service will act like a Façade between the endpoints and the data store. And repositories are based on the Models existent in our system. Like Commitment Model, so CommitmentRepository.
The idea to hava the repo is have the code more organized and segregated. The repository is responsible to organize the communication with the database table or collection based on the business rules with very clear name of functions (E.g getCommitmentsByCompressedPkd or listCommitmentsByCompressedPkd).
The Service will be a facade that receive some input from the route (_some projects use controller between the rountes and the service. Controller is to take care of the inputs coming from the api call_) and will send this input for the repo to do something in the database. If was some GET will send a return back to the route.
So the route doesn't need to know what is happening in the service. And the service just need to do calls for the repos to build some return to the route.
The names are just created from somebody to use this pattern.

nightfall-client/src/routes/commitment.mjs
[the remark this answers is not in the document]
luizoamorim: Removed in this [commit](https://github.com/EYBlockchain/nightfall_3/pull/802/commits/b37ed8578ab8e08593ddb96a260dc9cdc5b917d3)

nightfall-client/src/repositories/commitment-repository.mjs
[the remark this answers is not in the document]
luizoamorim: Resolved in this [commit](https://github.com/EYBlockchain/nightfall_3/pull/802/commits/e260f9e3889adafe04532ed0f08874dfe58f9e58)

nightfall-client/src/repositories/commitment-repository.mjs
```diff
@@ -0,0 +1,48 @@
...
+ */
+export default async function getAllCommitmentsByCompressedPkd(compressedPkd) {
+  initDb();
+  const allCommitmentsByCompressedPKD = await db
+    .collection(COMMITMENTS_COLLECTION)
+    .find({
+      'preimage.compressedPkd': compressedPkd.toString(),
+    })
+    .toArray();
+  return allCommitmentsByCompressedPKD;
```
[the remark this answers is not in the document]
luizoamorim: I've tried to pass
```
const query = {}
const options = {
  projection: {
  preimage: { compressedPkd },
  }
}
```
And return me only the objects with an _id and with preimage: { compressedPkd: number }.
I was losing time in this. So maybe we can refactor late.

nightfall-client/src/repositories/commitment-repository.mjs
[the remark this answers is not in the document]
daveroga: Ok

daveroga (changes requested):
I think the operations are ok but the refactor should be done in a future PR. Now  the functionality should be placed in the corresponding modules that exist now.

nightfall-client/src/repositories/commitment-repository.mjs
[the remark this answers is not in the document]
luizoamorim: @daveroga In my vision, really, we should start to break the responsibilities at least for the new code. And we can do the refactor progressively. For me is easier keep how as is now. But I really prefer to start the refactoring progressively.

luizoamorim:
@daveroga @imagobea @israelboudoux The propose of a new architecture will stay for the future. So:

- I cleaned up the new structure I was creating.
- Refactored the code based on the updates on master.
- For me the code is good to be merged. But maybe you can find something to change in Review.

luizoamorim:
# Export commitments

The endpoint now allow the user get commitments by an array of compressed zkp public keys. This way we can leave the code prepared for future necessities. 
So, at this moment the user should pass a array with just one compressed key.

luizoamorim:
# Import commitments
The last commitmet coverage add the endpoint for saveAllCommitments for the import commitments flow.
So the task #804 was done here.

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -688,3 +688,67 @@ export async function findUsableCommitmentsMutex(
...
+ * business logic and of a repository doing the communication with the database for this
+ * use case.
+ * @returns all the commitments existent in this database.
+ * @author luizoamorim
+ */
+export async function getAllCommitments() {
+  const connection = await mongo.connection(MONGO_URL);
+  const db = connection.db(COMMITMENTS_DB);
+  const allCommitments = await db.collection(COMMITMENTS_COLLECTION).find({}).toArray();
+  return allCommitments;
```
imagobea, line 724:
```suggestion
  return db.collection(COMMITMENTS_COLLECTION).find().toArray();
```
Haven't tested this, but pretty sure you don't need the curly braces `{}`

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -688,3 +688,67 @@ export async function findUsableCommitmentsMutex(
     findUsableCommitments(compressedZkpPublicKey, ercAddress, tokenId, _value, onlyOne),
   );
 }
+
+/**
+ * @function getCommitmentsByCompressedPkdList do the role of a service taking care of the
```
imagobea, line 693: I prefer `getCommitmentsByCompressedPkdList` instead of getAllCommitmentsByCompressedZkpPublicKeyList, L701 ;p

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -688,3 +688,67 @@ export async function findUsableCommitmentsMutex(
...
+ * use case.
+ * @param {string[]} compressedZkpPublicKey a list of compressedZkpPublicKey derivated from the user
+ * mnemonic coming from the SDK or Wallet.
+ * @returns all the commitments existent for this list of compressedZkpPublicKey.
+ * @author luizoamorim
+ */
+export async function getAllCommitmentsByCompressedZkpPublicKeyList(listOfCompressedZkpPublicKey) {
+  const connection = await mongo.connection(MONGO_URL);
+  const db = connection.db(COMMITMENTS_DB);
+  const allCommitmentsByListOfCompressedZkpPublicKey = await db
```
imagobea, line 704: You could return directly from 704, specially because the `const` name is quite long

imagobea (changes requested):
Hello @luizoamorim 

Thanks for this updated version of the code and sorry that we could not provide better requirements from the start.

A part from a few naming suggestions (basically I think we can drop the `all` bit from functions and consts), the code LGTM.

The reason why I am requesting changes is because I would rather move all the import functionality to a separate branch/PR, so that others more skilled can take a look.

Finally, I'm thinking we can also rm the changes to the package.json? If there are no tests yet maybe there's no need to add Jest for now.

Cheers :taco:

nightfall-client/src/routes/commitment.mjs
```diff
@@ -88,6 +91,61 @@ router.get('/commitments', async (req, res, next) => {
...
+      throw response;
+    }
+    res.json({ response });
+  } catch (err) {
+    logger.error(err);
+    next(err);
+  }
+});
+
 router.get('/withdraws', async (req, res, next) => {
```
israelboudoux, line 149: I think you meant `withdrawals`

nightfall-client/src/routes/commitment.mjs
```diff
@@ -88,6 +91,61 @@ router.get('/commitments', async (req, res, next) => {
...
+  const arrayOfCompressedZkpPublicKey = req.body;
+  try {
+    const allCommitmentsByListOfCompressedZkpPublicKey =
+      await getAllCommitmentsByCompressedZkpPublicKeyList(arrayOfCompressedZkpPublicKey);
+    res.json({ allCommitmentsByListOfCompressedZkpPublicKey });
+  } catch (err) {
+    logger.error(err);
+    next(err);
+  }
+});
```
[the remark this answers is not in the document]
israelboudoux: I would move to the Rest approach in terms of nomenclature, like the following:
- Root path: `/commitments`
  - GET / (You already have this one!)
  - POST / (Instead of saveAll - having the swagger will give enough information this operation can save more than one object!)
  - etc.

This is a little guide: https://restfulapi.net/resource-naming/

What do you think?

nightfall-client/src/routes/commitment.mjs
```diff
@@ -12,6 +12,9 @@ import {
   getWithdrawCommitments,
```
israelboudoux, line 12: A very critical thing I would do, but probably not now, it is to add an `API Key` so that the endpoint wouldn't be freely available to everyone inside the same network, unless some that are supposed to be (e.g. `/health-check`). Please, could you create a ticket for this? (add a configurable API Key for all the nightfall-client endpoints)

nightfall-client/src/services/commitment-storage.mjs
israelboudoux: > I prefer `getCommitmentsByCompressedPkdList` instead of getAllCommitmentsByCompressedZkpPublicKeyList, L701 ;p

Cool! I would do this as well!

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -688,3 +688,67 @@ export async function findUsableCommitmentsMutex(
...
+ * @returns
+ */
+export async function saveAllCommitments(listOfCommitments) {
+  const connection = await mongo.connection(MONGO_URL);
+  const db = connection.db(COMMITMENTS_DB);
+  let exist = false;
+  for (const commitment of listOfCommitments) {
+    if (
+      // eslint-disable-next-line no-await-in-loop
+      (await db.collection(COMMITMENTS_COLLECTION).find({ _id: commitment._id }).toArray()).length >
```
israelboudoux, line 740: I don't know what would be the impact of removing this verification, the `for` iteration. A `save` operation normally is meant to insert or update, then having a pre-existing object in the DB wouldn't be a problem.

israelboudoux (changes requested):

nightfall-client/src/routes/commitment.mjs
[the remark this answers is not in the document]
imagobea: @israelboudoux so you suggest that instead of having 2 routes

a) `/` for get all
b) `/compressedPublicKeys` for get commitments filtering by compressedPub..

just have `a)`, right?

Two thoughts:

1. This part of the API already has a dedicated endpoint to filter by `salt` - so having `b)` wouldn't be "too alien" :alien:

2. Now we know that this compressedPub.. is not "sensitive", however I think we kept the POST method to avoid exceeding max query string length (although feels unlikely)

What do you say?? :blush:

nightfall-client/src/services/commitment-storage.mjs
luizoamorim: I created this thinking in the follow case:
- The user send a list of commitments: 50% already exist in the database and the other 50% doesn't exist.
- The idea was not insert nothing if have some wrong commit in the list.

But yes, if we remove. The database will generate an exception showing that the _id already exists. I just don't know if will generate the exception and will insert the commitments that don't exists. Or will generate the exception and won't insert nothing.

nightfall-client/src/routes/commitment.mjs
luizoamorim: This endpoing was not added by me.

nightfall-client/jest.config.js
```diff
@@ -0,0 +1,186 @@
+/*
+ * For a detailed explanation regarding each configuration property, visit:
+ * https://jestjs.io/docs/configuration
```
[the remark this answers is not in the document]
luizoamorim: Removed jest.config. We won't use it now.

nightfall-client/src/routes/commitment.mjs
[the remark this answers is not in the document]
luizoamorim: I'll not handle the root path now. Because we'll need to change all the calls. Let's do this refactor when we start to create the swagger documentation. 

I agree to remove the all so when we already have a plural.

nightfall-client/src/services/commitment-storage.mjs
luizoamorim: But the name change, is not more compressedPkd.

nightfall-client/src/services/commitment-storage.mjs
luizoamorim: image
You're correct.

nightfall-client/src/services/commitment-storage.mjs
luizoamorim: Could be.

nightfall-client/src/services/commitment-storage.mjs
luizoamorim: So. I'll filter only the commitments that don't exist in the database yet and will insert all the others like we talked in the chat.

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -688,3 +688,67 @@ export async function findUsableCommitmentsMutex(
...
+    .collection(COMMITMENTS_COLLECTION)
+    .find({
+      'preimage.compressedZkpPublicKey': { $in: listOfCompressedZkpPublicKey },
+    })
+    .toArray();
+  return allCommitmentsByListOfCompressedZkpPublicKey;
+}
+
+/**
+ * @function getAllCommitments do the role of a service taking care of the
```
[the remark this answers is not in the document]
luizoamorim: I removed the All from this plurals.

nightfall-client/src/routes/commitment.mjs
luizoamorim: I think is better you open this ticket. And add what do you think in the description.

nightfall-client/src/routes/commitment.mjs
```diff
@@ -88,6 +90,41 @@ router.get('/commitments', async (req, res, next) => {
   }
 });
 
+/**
+ * @description the endpoint that will send a reponse with all the
+ * existent commitments for the list of compressedPkd received in the
+ * request body.
+ * @author luizoamorim
+ */
+router.post('/byCompressedZkpPublicKey', async (req, res, next) => {
```
imagobea, line 99:
```suggestion
router.post('/compressedZkpPublicKeys', async (req, res, next) => {
```

Please update the route :pray:

imagobea: Like Israel suggested, please add a comment explaining why this is POST instead of GET

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -688,3 +688,38 @@ export async function findUsableCommitmentsMutex(
     findUsableCommitments(compressedZkpPublicKey, ercAddress, tokenId, _value, onlyOne),
   );
 }
+
+/**
+ * @function getCommitmentsByCompressedZkpPublicKeyList do the role of a service taking care of the
+ * business logic and of a repository doing the communication with the database for this
+ * use case.
```
imagobea, line 695: Is this still accurate? Im not sure if the docs refer to the other approach we discussed previously :thinking:

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -688,3 +688,38 @@ export async function findUsableCommitmentsMutex(
...
+export async function getCommitmentsByCompressedZkpPublicKeyList(listOfCompressedZkpPublicKey) {
+  const connection = await mongo.connection(MONGO_URL);
+  const db = connection.db(COMMITMENTS_DB);
+  const commitmentsByListOfCompressedZkpPublicKey = await db
+    .collection(COMMITMENTS_COLLECTION)
+    .find({
+      'preimage.compressedZkpPublicKey': { $in: listOfCompressedZkpPublicKey },
+    })
+    .toArray();
+  return commitmentsByListOfCompressedZkpPublicKey;
```
imagobea, line 710: I will insist one more time about returning directly from L704 :upside_down_face: (in case you missed the comment!?)

imagobea (approved):
Hi @luizoamorim ,

Thank you for the changes and for taking out the import from this PR.

I'm approving, but please see my comments:

- I would really like to update the "all commitments" route as per the suggestion
- Perhaps there are some comments outdated from the first version where you changed the structure a bit
- Please remove spaces from `package.json`

Many thanks again!! :pray:

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -688,3 +688,38 @@ export async function findUsableCommitmentsMutex(
...
+ * business logic and of a repository doing the communication with the database for this
+ * use case.
+ * @returns all the commitments existent in this database.
+ * @author luizoamorim
+ */
+export async function getCommitments() {
+  const connection = await mongo.connection(MONGO_URL);
+  const db = connection.db(COMMITMENTS_DB);
+  const allCommitments = await db.collection(COMMITMENTS_COLLECTION).find().toArray();
+  return allCommitments;
```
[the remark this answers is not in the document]
imagobea: Docs are fine. Eslint complains about returning from that line.

nightfall-client/src/services/commitment-storage.mjs
imagobea: Eslint complains about returning from that line.

nightfall-client/src/services/commitment-storage.mjs
imagobea: Docs are fine.

daveroga (approved):
LGTM

PR #806, opened 2022-07-09
Studying rabbitmq implementation

signorecello:
Moving rabbitmq studies and possible implementation here


israelboudoux:
Hey @signorecello, could we discuss about that solution before implementing it? :D

PR #807, opened 2022-07-12
fix: adapt config for real deployment

druiz0992:
Added some configurability to MULTISIG variables for ease of deployment


daveroga (approved):
LGTM

PR #811, opened 2022-07-14
Issue o - Poseidon hash

Westlad:
This fixes [issue o](https://github.com/EYBlockchain/nightfall_3_private/issues/137).  It replaces the SHA256 hash, used for commitment and nullifier calculation, with a Poseidon hash.   As a result, there are significant circuit changes because most input variables change type to a field.  This results in considerable simplification overall however, and much smaller circuits.


Westlad:
Closing.  Will rebase onto #781 (KEM-DEM) as this will make the merge of the latter simpler.

Issue #813, opened 2022-07-14
Allow Optimist to store invalid blocks info


daveroga:
We will store information about invalid blocks in optimist and the type of challenge they generate.

- invalidCode
- invalidMessage
- block data

PR #814, opened 2022-07-14
Save invalid blocks in optimist database

daveroga:
Fixes: #813 

We will store information about invalid blocks in optimist and the type of challenge they generate:

- Created collection `invalid_blocks` in optimist.
- `invalid_blocks` will store this information when an invalid block is detected:
   - invalidCode
   - invalidMessage
   - block data
- Fixed adversary test check balances when calling `npm run adversary-test-all`

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -17,6 +17,7 @@ module.exports = {
   CHALLENGER_COLLECTION: 'challengers',
   TRANSACTIONS_COLLECTION: 'transactions',
   SUBMITTED_BLOCKS_COLLECTION: 'blocks',
+  INVALID_BLOCKS_COLLECTION: 'invalid_blocks',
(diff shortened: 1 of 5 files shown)


druiz0992:
Do we store address of challenger?

daveroga:
> Do we store address of challenger?
We store the invalid block once it is produced and detected by optimist. At this time we don't know the challenger that will pick the challenge.
I could investigate if I can get the challenger from the emit Rollback detected from the Challenges contract in the optimist, but the sender is not passed now in the event. We could pass it in the Rollback event to know the challenger and then store it in the invalid block stored previously in the optimist database.

israelboudoux (approved):
LGTM

druiz0992:
@daveroga it would be would if we could get the challenger address for statistics

druiz0992 (approved):

PR #823, opened 2022-07-20
Stuffing ping-pong in a mocha test

signorecello:
I'm cherry-picking some leftovers from the (then dropped) PR about mock contracts.

This PR proposes that ping-pong tests are made into a mocha test so it's easier and faster to run, develop, debug, etc


israelboudoux (approved):
LGTM, nice cleaning btw, this will require me to change my stuff related to the Ping-Pong test that I'm doing for Jenkins, but it is for the better.

Westlad (approved):

druiz0992 (changes requested):
Changes in this PR affect how the proposer is launched in private repo. Before merging this PR, private repo needs to work correctly

signorecello:
I finally was able to make the change on the deployment repo (uff), so I'm requesting re-review

druiz0992 (approved):

PR #825, opened 2022-07-22
fixed duplicate nullifier, part fixes of adversary, added duplicate commitment, lot of refactor for efficiency

ChaitanyaKonda:
What this PR does
- Added detection of duplicate nullifier in different transactions of the block or within same transaction during check block
- Corrects nullifier challenge such that it can also send a nullifier challenge when there is a duplicate within same transaction/block
- Removed `checkTransactionHash` because transactionHash is now calculated during `proposeBlock`
- Removed `challengeHistoricRoot` because `challengeProofVerification` can do this
- Refactored check block
- Efficient rewrite of duplicate nullifier by removing nullifier collection which was purely meant to retrieve all the nullifiers and cross check nullifiers of an L2 transaction against each nullifier in collection. The iteration was over all the nullifiers and not all the nullifiers of L2 transactions. Which is even worse for efficiency
- Remove the duplicate work that duplicate nullifier challenge does to fetch relevant parameters to be passed to nullifier challenge 
- Removed duplicate transaction and replaced with duplicate commitment
- duplicate commitment checks are done within a transaction, within a block and over all L2 transactions 
- Rollback checks not just the validity of the current L2 block transactions but all transactions in mempool, this has been removed because validity is checked now in transaction submitted event handler, block assembly and rollback event handler
- Added `removeCommitmentsFromMemPool` and `removeNullifiersFromMemPool` to identify other currently in mempool transactions that have the same commitments/nullifiers as an L2 transaction being processed in Block proposed event handler
- Fixed adversary test. proposeBlock now calculates transactionHash, so the transactionHash of a modified transaction should be updated too for duplicate nullifier, duplicate commitment and verify proof transaction creation in adversary 
- - If an unknown transaction with a duplicate commitment is received by an optimist during BPEH, then it has to save this despite duplicate commitment error because this transaction needs to be deleted during rollback. Added this
- Current tests check for expected balance which is the same if there is challenges are done or not done entirely. Hence these tests should move to also check for number of rollback events for completeness. 
- Challenge queue is paused at the beginning and unpaused only after transactions are submitted. Expected balance check after pausing queue won’t work if there are duplicate commitment or nullifier transactions in L2 block which go unchallenged because unused commitment/nullifier are not submitted by the adversary in L2 block and hence this will not be marked as on chain
- IncorrectProof is not tested because there is a bug where `ERCAddress` is bigger than 20 bytes. So `libCheckOverflows` throws `ERC address is out of range` EVM error and reverts.

Test:
- Github action tests adversary. Adversary test is currently broken because of issues outside the scope of the this issue. Which is why it will fail. But if duplicateCommitment or duplicateNullifier adversary tests are run individually. This will work as expected

To be done in a different issue
- `checkProofVerify` in case verification keys change before block assembly 
- `checkDuplicateCommitent` and `checkDuplicateNullifier` to be done at the time of block assembly to identify duplicate commitment/nullifier from the time after the transaction was checked for validity in transaction submitted to the time it is picked for assembly into an L2 Block


nightfall-optimist/src/services/database.mjs
```diff
@@ -132,23 +131,23 @@ export async function getBlockByTransactionHashL1(transactionHashL1) {
...
-function to get a block by blockHash, if you know the hash of the block. This
-is useful for rolling back Timber.
-*/
-export async function getBlockByBlockHash(blockHash) {
-  const connection = await mongo.connection(MONGO_URL);
-  const db = connection.db(OPTIMIST_DB);
-  const query = { blockHash };
-  return db.collection(SUBMITTED_BLOCKS_COLLECTION).findOne(query);
-}
+// export async function numberOfBlockWithTransactionHash(transactionHash) {
```
signorecello, line 134: Can we delete this instead of commenting out?

test/adversary/adversary-code/block.mjs
```diff
@@ -7,18 +7,16 @@ let error = process.env.BAD_BLOCK_SEQUENCE
       'ValidBlock',
       'ValidBlock',
       'ValidBlock',
-      'IncorrectTreeRoot', // Needs two prior blocks
-      'ValidBlock',
-      'IncorrectLeafCount', //  Needs one prior block
-      'ValidBlock',
-      'DuplicateTransaction', // needs atleast one transaction in a prior block
+      // 'IncorrectTreeRoot', // Needs two prior blocks
```
signorecello, line 10: Same here :)

nightfall-optimist/src/services/database.mjs
ChaitanyaKonda: I removed a lot of functions but left this behind for later use

nightfall-optimist/src/event-handlers/block-proposed.mjs
```diff
@@ -78,35 +80,23 @@ async function blockProposedEventHandler(data) {
...
-          ...tx.nullifiers.filter(nulls => nulls !== ZERO),
-          ...tx.nullifiersFee.filter(nulls => nulls !== ZERO),
-        ])
-        .flat(Infinity),
-      block.blockHash,
-    );
     // mark transactions so that they are out of the mempool,
     // so we don't try to use them in a block which we're proposing.
-    await removeTransactionsFromMemPool(block.transactionHashes, block.blockNumberL2, timeBlockL2); // TODO is await needed?
+    await removeTransactionsFromMemPool(block.transactionHashes, block.blockNumberL2); // TODO is await needed?
```
RogerTaule, line 93: Why are you removing timeBlockL2?

nightfall-optimist/src/event-handlers/block-proposed.mjs
```diff
@@ -78,35 +80,23 @@ async function blockProposedEventHandler(data) {
...
+    const blockCommitments = transactions.map(t => t.commitments.filter(c => c !== ZERO)).flat();
```
RogerTaule, line 94: You need to take into account commitmentFee too.
```suggestion
    const blockCommitments = transactions
      .map(t => [...t.commitments, ...t.commitmentFee].filter(c => c !== ZERO))
      .flat();
```

nightfall-optimist/src/event-handlers/block-proposed.mjs
```diff
@@ -78,35 +80,23 @@ async function blockProposedEventHandler(data) {
...
+    const blockNullifiers = transactions.map(t => t.nullifiers.filter(c => c !== ZERO)).flat();
```
RogerTaule, line 95: You need to take into account nullifiersFee too.
```suggestion
     const blockNullifiers = transactions
      .map(t => [...t.nullifiers, ...t.nullifiersFee].filter(c => c !== ZERO))
      .flat();
```

nightfall-optimist/src/event-handlers/block-proposed.mjs
ChaitanyaKonda: I must have missed this out during rebase

nightfall-optimist/src/event-handlers/block-proposed.mjs
ChaitanyaKonda: This PR has been pending for 26 days that I wanted to fix the merge conflicts and have a merged as soon as possible because it makes major changes to optimist and is already a very PR. Since the fees was merged before this, I have to first understand fee and then add further features such that code can e added to check all the fee commitments and fee nullifiers too for duplicates and pass in the correct isNullifierFee1 or isCommitmentFee 1 flag to challenges.sol for duplicate nullifier and nullifier commitment challenges. I will to raise this as a separate issue fix

nightfall-optimist/src/event-handlers/block-proposed.mjs
LijuJoseJJ: timeBlockL2 added back

nightfall-optimist/src/event-handlers/block-proposed.mjs
LijuJoseJJ: commitmentFee change added

nightfall-optimist/src/event-handlers/block-proposed.mjs
LijuJoseJJ: filter nullifiersFee added

israelboudoux:
Hey @ChaitanyaKonda / @LijuJoseJJ, could you solve the merge conflicts please, so we can merge this PR? 
Thanks!

nightfall-optimist/src/event-handlers/rollback.mjs
```diff
@@ -36,131 +33,52 @@ async function rollbackEventHandler(data) {
...
 
-  logger.debug(
-    `Nullifier of deleted transactions: ${
-      transactions.find(t => t.transactionHash === invalidTransactionHashesArr[0])?.nullifiers[0]
-    }`,
+  // We can now reset or delete the local database.
+  await Promise.all(
+    blocksToBeDeleted
+      .map(async block => [addTransactionsToMemPool(block), deleteBlock(block.blockHash)])
+      .flat(1),
```
pawelgrzybek, line 76: `1` is a default value for `flat()` so could easily be skipped.

ChaitanyaKonda: True. It is a minor change, so I don't want to kick start the github action tests again because it has been waiting to be merged for 1.5 months now

daveroga:
Don't we test deep rollbacks with the `NONSTOP_QUEUE_AFTER_INVALID_BLOCK: 'true'` in test?

daveroga (approved):
LGTM

Issue #828, opened 2022-07-26
Add timestamp to db collections


druiz0992:
Whenever a block / transaction is added to mongoDb, we need to include a timestamp (UTC) of when the transaction was mined/challenged. We will use that information in the block scanner

PR #829, opened 2022-07-27
Save block proposed time in transactions

daveroga:
Fixes: #828 

## Objective
Whenever a block / transaction is added to mongoDb, we need to include a timestamp (UTC) of when the transaction was mined/challenged. We will use that information in the block scanner

## Solution
- Add time of block proposed to transactions in optimist and client database (`timeBlockL2`)
- Add time of block proposed to blocks in optimist and client database (`timeBlockL2`).


druiz0992 (changes requested):
- Why can’t you add block-info.mjs to common files and avoid duplication?
- Can you add time stamp to block as well?

daveroga:
> * Why can’t you add block-info.mjs to common files and avoid duplication?
> * Can you add time stamp to block as well?

Yes, did it but didn't pushed it yet. Now both points are done.

druiz0992 (approved):

Westlad (approved):

PR #839, opened 2022-07-31
refactor done seprated constants from config

LijuJoseJJ:
this PR fixes  https://github.com/EYBlockchain/nightfall_3_private/issues/158


Westlad (changes requested):
I think we need to be careful about the definition of a `config` item and a `constant`.  A `constant` will never depend on the details of our deployment.  An example is `ZERO: 0x000....`.  A `config` item is something that is deployment dependent, for example the name of a Mongo collection or the height of a Merkle tree.  There are only a few grey areas, such as the definition of calldata types.  These could arguably be either but are best categorised as constants because they will actually never be dependent on the details of a deployment.

LijuJoseJJ:
> I think we need to be careful about the definition of a `config` item and a `constant`. A `constant` will never depend on the details of our deployment. An example is `ZERO: 0x000....`. A `config` item is something that is deployment dependent, for example the name of a Mongo collection or the height of a Merkle tree. There are only a few grey areas, such as the definition of calldata types. These could arguably be either but are best categorised as constants because they will actually never be dependent on the details of a deployment.

changes accommodated

common-files/constants/constants.json
```diff
@@ -0,0 +1,28 @@
...
+  "EXCLUDE_DIRS": "common",
+  "PROOF_QUEUE": "generate-proof",
+  "MAX_QUEUE": 5,
+  "ZKP_KEY_LENGTH": 32,
+  "CONFIRMATION_POLL_TIME": 1000,
+  "CONFIRMATIONS": 12,
+  "NODE_HASHLENGTH": 32,
+  "DEFAULT_ACCOUNT_NUM": 10,
+  "ZERO": "0x0000000000000000000000000000000000000000000000000000000000000000",
+  "HASH_TYPE": "poseidon",
```
Westlad, line 17: Note a constant - we change the hash type from time to time

common-files/constants/constants.json
```diff
@@ -0,0 +1,28 @@
...
+  "TXHASH_TREE_HASH_TYPE": "keccak256",
```
Westlad, line 18: Not a constant

common-files/constants/constants.json
```diff
@@ -0,0 +1,28 @@
+{
+  "CONTRACT_ARTIFACTS": "/app/build/contracts",
+  "PROPOSERS_CONTRACT_NAME": "Proposers",
+  "SHIELD_CONTRACT_NAME": "Shield",
+  "CHALLENGES_CONTRACT_NAME": "Challenges",
+  "STATE_CONTRACT_NAME": "State",
+  "BLOCK_PROPOSED_EVENT_NAME": "BlockProposed",
+  "EXCLUDE_DIRS": "common",
+  "PROOF_QUEUE": "generate-proof",
+  "MAX_QUEUE": 5,
```
Westlad, line 10: Not a constant

common-files/constants/constants.json
```diff
@@ -0,0 +1,28 @@
...
+  "ZKP_KEY_LENGTH": 32,
+  "CONFIRMATION_POLL_TIME": 1000,
```
Westlad, line 12: Not a constant

common-files/constants/constants.json
```diff
@@ -0,0 +1,28 @@
...
+  "CONFIRMATIONS": 12,
```
Westlad, line 13: Not a constant

common-files/constants/constants.json
```diff
@@ -0,0 +1,28 @@
+{
+  "CONTRACT_ARTIFACTS": "/app/build/contracts",
```
Westlad, line 2: not a constant

common-files/constants/constants.json
```diff
@@ -0,0 +1,28 @@
...
+  "PROPOSERS_CONTRACT_NAME": "Proposers",
+  "SHIELD_CONTRACT_NAME": "Shield",
+  "CHALLENGES_CONTRACT_NAME": "Challenges",
+  "STATE_CONTRACT_NAME": "State",
+  "BLOCK_PROPOSED_EVENT_NAME": "BlockProposed",
+  "EXCLUDE_DIRS": "common",
```
Westlad, line 8: not a constant

common-files/constants/constants.json
```diff
@@ -0,0 +1,28 @@
...
+  "PROOF_QUEUE": "generate-proof",
```
Westlad, line 9: not a constant

common-files/constants/constants.json
```diff
@@ -0,0 +1,28 @@
...
+  "MAX_QUEUE": 5,
+  "ZKP_KEY_LENGTH": 32,
+  "CONFIRMATION_POLL_TIME": 1000,
+  "CONFIRMATIONS": 12,
+  "NODE_HASHLENGTH": 32,
+  "DEFAULT_ACCOUNT_NUM": 10,
```
Westlad, line 15: Not a constant

common-files/constants/constants.json
LijuJoseJJ: moved to config

common-files/constants/constants.json
LijuJoseJJ: moved to config

common-files/constants/constants.json
LijuJoseJJ: moved to config

common-files/constants/constants.json
LijuJoseJJ: moved to config

common-files/constants/constants.json
LijuJoseJJ: moved to config

common-files/constants/constants.json
LijuJoseJJ: moved to config

Westlad (approved):

common-files/constants/constants.json
LijuJoseJJ: moved to config

common-files/constants/constants.json
LijuJoseJJ: moved to config

common-files/constants/constants.json
LijuJoseJJ: moved to config

druiz0992 (approved):

PR #841, opened 2022-08-01
remove escrow transfer function and admin bugfix

### LINKED ISSUE #835: remove escrow control
We do not require the ability to control escrow funds because the Pausable interface fulfils an equivalent but safer functionality

Westlad:
fixes #835 

This PR removes the ability to transfer funds from the Shield escrow pools. It's no longer required as withdrawals can be paused in the event of an emergency.

It also fixes a bug in the administrator container whereby actions that required more than one function call to the blockchain were not correctly handled.

There is no specific test for the removal of the transfer function (because it was not tested in any case).  The normal test suite suffices.  To test the administrator function, refer to the instructions [here](https://github.com/EYBlockchain/nightfall_3/blob/westlad/escrow-removal/nightfall-administrator/README.md)


druiz0992 (approved):

signorecello (approved):
LGTM

PR #842, opened 2022-08-01
Docker Improvements #2

israelboudoux:
Moving all docker-related files to the new `docker` folder. 
- All dependencies were amended;
- A docker-compose.yml still exists at the root of the project, since the `eslint` complains when it is removed.


signorecello (approved):
LGTM

druiz0992 (changes requested):
Can we not change container images names? There are several reasons:
- Its good idea to have nightfall as the preamble, so its easy to filter containers
- there are some scripts using existing names, and it will be costly to change them

israelboudoux:
> Can we not change container images names? There are several reasons:
> 
> * Its good idea to have nightfall as the preamble, so its easy to filter containers
> * there are some scripts using existing names, and it will be costly to change them

Hey @druiz0992, that change was needed since the image names are derived from the folder where the docker-compose files are living, (e.g. if they are under `nightfall_3` folder, the image name for the `administrator` service will be `nightfall_3_administrator`), then with this change they all have changed. I'll see how can I avoid these changes and I'll let you know.

israelboudoux:
> > Can we not change container images names? There are several reasons:
> > 
> > * Its good idea to have nightfall as the preamble, so its easy to filter containers
> > * there are some scripts using existing names, and it will be costly to change them
> 
> Hey @druiz0992, that change were needed since the image names are derived from the folder where the docker-compose files are living, (e.g. if they are under `nightfall_3` folder, the image name for the `administrator` service will be `nightfall_3_administrator`), then with this change they all have changed. I'll see how can I avoid these changes and I'll let you know.

@druiz0992, changes are done as you asked!

README.md
```diff
@@ -90,7 +90,7 @@ this point.
 
 ### To end the application
 
-To stop the application, you can run `docker-compose down` and it should exit cleanly.
+To stop the application, you can run `docker-compose -f docker/docker-compose.yml -f docker/docker-compose.dev.yml -f docker/docker-compose.ganache.yml -p 'nightfall_3' down -v` and it should exit cleanly.
```
pawelgrzybek, line 93: Can we abstract it out to npm script, like `npm run down` or something like that?

docker-compose.yml
```diff
@@ -1,143 +1 @@
...
-  build: null
-networks:
-  nightfall_network:
-    driver: bridge
-    ipam:
-      driver: default
-      config:
-        - subnet: 172.16.238.0/24
-          gateway: 172.16.238.1
+# LEFT EMPTY ON PURPOSE, OTHERWISE 'eslint' FAILS, I'M NOT SURE WHY!
```
pawelgrzybek, line 1: Which ESLint rule fails?

pawelgrzybek (changes requested):

README.md
israelboudoux: It is done @pawelgrzybek!

pawelgrzybek (approved):

druiz0992 (approved):

PR #843, opened 2022-08-01
Flexible Input/Output circuits

### LINKED ISSUE #262: Investigate requirement for the Invalid Transaction checks and challenges
As part of the `checkTransaction` we check the structure of our transactions - i.e. such that they conform to one of the transaction types.

Seemingly, we have been able to make do without actually throwing these challenges (see screenshot below). In fact our `invalid transaction` test seems to be caught as  an `Invalid Public input hash`.

We should check if it is necessary to check the transaction structure - or if any invalid structure is covered by other challenges. This would to save code in `optimist` but also in `challenges.sol` (where space is very much a luxury).

image

### LINKED ISSUE #818: Generalize transfer and withdrawal circuits
Instead of having single_transfer and double_transfer, create a general circuit using generics that can be used for multiple scenarios.
### Zokrates Circuits  
- [x] Generic Verify Transaction
- [x] Generic Circuit Template
- [x] Generate Deposit Circuits
- [x] Generate Transfer Circuits
- [x] Generate Withdraw Circuits

IlyasRidhuan:
This PR closes #818 and closes #262 and removes the need for specific circuits based on the commitments and nullifiers. Instead we have 3 circuits (deposit, transfer, withdraw) that can support variable number of commitment/nullifiers.  There are considerable changes to the way the circuits and commitment selection works in this PR.

### Circuits
1) The circuits are refactored to make liberal use of generics across the key `verify` functions (e.g. `verify_nullifier`, `verify_commitment` and `verify_structure`).
2) The public transaction struct has been included as part of the public input to the proof. This means the circuit also performs the structural check for the validity of the transaction struct itself. This heavily simplifies the challenges as we no longer need to have a specific transaction structure challenge.

Operation | Constraints Before | Constraints After | % Change
-- | -- | -- | --
Deposit | 619 | 921 | +48%
Single Transfer | 26,320 | 39,363 | +50%
Double Transfer | 39,295 |39,363 | +0.01%
Withdraw | 12,075 | 26,050 | +112%

**Note:** There is no longer a specific `single_transfer` circuit as the transfer circuit handles all scenarios (hence the constraint is as large as possible to also handle double_transfer). Similarly, withdraw must be able to accommodate the most complex scenario of 2 inputs with a single output, which increases the constraints significantly.

### Commitment Selection
We now have the flexibility to better select the type of transaction that will be performed based on the current user's UTXO (notes). Previously, commitment selection was based on:

1. Perform a single transfer when possible
2. Else perform a double transfer if the total inputs were greater than the required output
**Note: A double transfer would fail if the sum of inputs were equal to the required output (i.e. it had to be strictly greater than). Withdraw's needed to be of exact value.**

With the new flexible circuits we perform the following checks

1. Perform a single transfer if possible.
2. Perform a double transfer without change if we have a sum of inputs that match the exact value.
3. Perform a single transfer with change if we have a single input > output required
4. Perform a double transfer with change otherwise.

This is still subject to change, especially in the prioritising of (3) and (4). Note that a withdraw now attempts to do (1) to (3) instead of being restricted to an exactly matching input.

### Testing
This can be tested by running `npm run test-circuits` while nightfall is running


nightfall-deployer/circuits/common/generic_circuit/Verifiers/verify_commitments.zok
```diff
@@ -0,0 +1,34 @@
...
+    private field[NumCommitments] newCommitmentValues,\
+    private field[NumCommitments] newCommitmentSalts,\
+    private Point[NumCommitments] recipientPublicKey\
+) -> bool:
+
+    //Check that all the commitments are valid. If NumCommitments equals zero this loop will be ignored
+    for u32 i in 0..NumCommitments do
+        
+        //Calculate the commmitment hash from the newCommitment parameters
+        field commitment = if newCommitmentValues[i] == 0 then  0 else \
```
RogerTaule, line 20: ERC721 tokens have always value = 0. Therefore, the poseidon hash will never be calculated and the verification will fail. I think we also need to check if the commitmentHash is equal to zero. If ERC20 or ERC1155 and value is zero, since the minimum value for this tokens is 1, this will still hold. The equation will also hold for ERC721 since if the commitment is void the hash will be zero.
```suggestion
        field commitment = if newCommitmentValues[i] == 0 && commitmentHashes[i] == 0 then  0 else \
```

nightfall-deployer/circuits/common/generic_circuit/Verifiers/verify_nullifiers.zok
```diff
@@ -0,0 +1,61 @@
...
+
+        //Calculate the nullifier hash from the oldCommitment parameters
+        field calculatedOldCommitmentHash = poseidon([\
+            packedErcAddress,\
+            idRemainder,\
+            oldCommitmentValues[i],\
+            ...zkpPublicKeys,\
+            oldCommitmentSalts[i]\
+        ])
+        field nullifier = if(oldCommitmentValues[i] == 0) then 0 else poseidon([nullifierKeys, calculatedOldCommitmentHash]) fi
```
RogerTaule, line 48: Same change as in the verify_commitments
```suggestion
        field nullifier = if(oldCommitmentValues[i] == 0 && nullifierHashes[i] == 0) then 0 else poseidon([nullifierKeys, calculatedOldCommitmentHash]) fi
```

nightfall-deployer/circuits/common/generic_circuit/Verifiers/verify_nullifiers.zok
```diff
@@ -0,0 +1,61 @@
...
+
+        //Check that the calculated nullifier matches with the one contained in the transaction
+        assert(nullifier == nullifierHashes[i])
+
+        //Check that the nullifier is contained in the tree
+        bool pathValidity = if(oldCommitmentValues[i] == 0) then true else pathCheck([roots[i], ...paths[i]], orders[i], calculatedOldCommitmentHash) fi
```
RogerTaule, line 54: Same change 
```suggestion
        bool pathValidity = if(oldCommitmentValues[i] == 0 && nullifierHashes[i] == 0) then true else pathCheck([roots[i], ...paths[i]], orders[i], calculatedOldCommitmentHash) fi
```

RogerTaule (changes requested):

nightfall-deployer/circuits/common/generic_circuit/Verifiers/verify_nullifiers.zok
```diff
@@ -29,33 +29,39 @@ def main<NumNullifiers>(\
...
         //Check that the calculated nullifier matches with the one contained in the transaction
         assert(nullifier == nullifierHashes[i])
 
         //Check that the nullifier is contained in the tree
-        bool pathValidity = if(oldCommitmentValues[i] == 0) then true else pathCheck([roots[i], ...paths[i]], orders[i], calculatedOldCommitmentHash) fi
+        bool pathValidity = if(oldCommitmentValues[i] != 0 || (i+1) != MinNullifiers) then pathCheck([roots[i], ...paths[i]], orders[i], calculatedOldCommitmentHash) else true fi
         assert(pathValidity)
 
-        //Set the changeZkpPublicKeys if i = 0. Otherwise just set the same value
-        firstInputZkpPublicKeys = i == 0 ? zkpPublicKeys : firstInputZkpPublicKeys
```
RogerTaule: Why the firstInputZkpPublicKeys inicialization was removed?

nightfall-deployer/circuits/withdraw.zok
```diff
@@ -11,7 +11,7 @@ def main(\
     PublicTransaction tx,\
     field[2] roots,\
     private Nullifiers<2> nullifiers,\
-    private Commitments<2> commitments\
+    private Commitments<1> commitments\
```
RogerTaule, line 14: If we set Commitments<1> in the withdraw.zok, withdraw_stub.zok needs to also be updated

RogerTaule (dismissed):

RogerTaule (approved):
Looks good to me :)

druiz0992 (approved):

PR #845, opened 2022-08-01
Several fixes

druiz0992:
Several issues fixed:
- if number of unprocessed transactions is larger than `TRANSACTIONS_PER_BLOCK` , it could happen that transactions put in a block exceed this limit also.
- If blockProposed is reverted, transactions are sent back to mempool . **IMPORTANT** I am not sure changes applied are correct.
- Change ERC721 Token ID to start from 2^254 + 1
- Increased waiting time in ping pong test to adapt to testnet environments
- Modified `limit` function to be in line with the rest of the uses in the file

**NOTE** Changes to tokenId to start at 2^254+1 (use 255 bits) uncovered a bug in the way commitments are retrieved; ie they were shown as Numbers (which did not have enough dynamic range to show such large numbers). Now token Ids are returned as 32Byte hex string. I think this change does not affect anything else, but i would like confirmation (changes in nightfall-client/src/services/commitment-storage.mjs)


IlyasRidhuan (approved):

imagobea (approved):
To go with the flow I'm approving the changes - I'm afraid I can't evaluate much. But I would suggest to define testing scenarios for the changes. E.g.

1. When number of unprocessed transactions is bigger than TRANSACTIONS_PER_BLOCK, block is proposed correctly and contains the right number of transactions.
Define also test cases for expected behaviour regarding the transactions left - eg, do we keep creating blocks till unprocessed transactions is `lt` TRANSACTIONS_PER_BLOCK?

2. If blockProposed is reverted, transactions are sent back to mempool

3. Define a test for the change in ERC721 regarding `tokenId`

Thank you

druiz0992:
@imagobea 

 - `When number of unprocessed transactions is bigger than TRANSACTIONS_PER_BLOCK, block is proposed correctly and contains the right number of transactions`- private repo stress testing

 - `Define also test cases for expected behaviour regarding the transactions left - eg, do we keep creating blocks till unprocessed transactions is lt TRANSACTIONS_PER_BLOCK?` - tested in most token tests

- `If blockProposed is reverted, transactions are sent back to mempool` -> adversary test

- `Define a test for the change in ERC721 regarding tokenId` -> ERC721 test

PR #851, opened 2022-08-04
Enable pino logger

### LINKED ISSUE #846: Refactoring logs #1
- We will stick to the `pino` logging library (https://www.npmjs.com/package/pino), just to follow what **nightfall-sdk** is using, then it will become our default library. We need to configure it;
- The logging level and other possible definitions will be defined via environment variables (See https://github.com/maticnetwork/nightfall-sdk/blob/master/libs/utils/logger.ts);
- The default format for logs is JSON, then this should be configured accordingly in the library;
- The logs can use `pretty JSON` when using a `local` environment, just to facilitate reading the logs during development. We can do this via some environment variable;
- The output should be directed to **stdout/stderr** - I think we are already doing it!;
- Change all logs calls that aren't the default ones (`trace`, `debug`, `info`, `warn`, `error`) to `debug` (e.g. `silly`);
- Test to verify the correct behavior of the applications and to assert that the work is completed thoroughly.

In another moment we will refactor the log messages and add information that could be valuable for troubleshooting (See [#847](https://app.zenhub.com/workspaces/polygon-nightfall-6148eee81b1164001874730b/issues/eyblockchain/nightfall_3/847))

druiz0992:
Fixes #846 

Use pino logger so that its easier to export logs to data dog monitoring services


israelboudoux (approved):
LGTM

Westlad (approved):

israelboudoux:
I've added a DNM because I'll do a couple of tests before merging.

israelboudoux:
I've added a DNM because I'll do a couple of tests before merging.

PR #859, opened 2022-08-08
Roger taule/issue733

### LINKED ISSUE #733: Payment to multiple proposers
This issue is about enabling a user of Polygon-Nightfall to be able to pay a Proposer for including their transaction in a block.
A protocol which will enable the payment mechanism is described
[here](https://github.com/EYBlockchain/nightfall_3/blob/westlad/payements/doc/payments.md).  However the person addressing this issue should be free to vary the protocol as their thinking develops.
Note that this protocol also introduces `zero commitments` and `zero nullifiers`.  These are needed to reduce the number of circuits that would otherwise be required and therefore they have wider impact than just this issue.

RogerTaule:
This PR closes #733 and allows payments to multiple proposers to be done. Payments are done with ETH for deposits and MATIC for transfers / withdrawals.  The transaction structure has been considerably modified to support payments. Each transaction now has two extra nullifiers and one commitment (change) for the fees. In addition, the historical root for the fee nullifiers are added and the fee is also stored in the transaction structure. 

**Circuits**
Public transaction has been modified to accomodate the new public transaction structure. To make sure that the fee nullifiers / commitment are valid, we use the generic `verify` functions implemented in #843 

| **Operations**   | **Constraints before**  | **Constrains after**   |  **% Change**  |
|---|---|---|---|
|  Deposit  |  921  | 923  |  0.2% | 
|  Transfer   |  39,363   |  64,191  |  63 %  | 
|  Withdraw  |  26,050  |  50,244  | 92.8 %  |  

**Note:** Since deposit is directly paid in ETH, there is almost no change in the number of constraints of the circuit

**Smart Contracts**

Two changes has been done to the Smart Contract: add the matic address in the config and  allow proposers to collect the fees.

1. Instead of hardcoding the matic address in the circuits, which would imply that for each environment (mainnet, testnet, development...) the address would need to be changed, the matic address is initialized in the smart contracts. Then, if a transaction fee isn't paid in MATIC, that transaction will be challenged.

2. A feebook mapping  `(bytes32 -> uint256[2])` has being added in the State.sol to keep track of all the fees paid to the proposer. The input bytes32 corresponds to the keccak of the block proposer and the block L2 number. The first element of the output array corresponds to ETH payments, while the second one corresponds to MATIC payments. The feebook is set when proposing a new block using `proposeBlock`, and can be claimed using `requestBlockPayment`. 

** NF3 ** 

The `submitTransaction` function has been adjusted to perform correctly with the payments.


zokrates-worker/src/index.mjs
```diff
@@ -33,7 +33,13 @@ const checkCircuitsOutput = async () => {
       : `${DEFAULT_CIRCUIT_FILES_URL}/${env}`;
     const url = `${baseUrl}/proving_files/hash.txt`;
     const outputPath = `./output`;
-    const circuits = ['deposit', 'double_transfer', 'single_transfer', 'withdraw'];
+    const circuits = [
+      'deposit',
```
druiz0992, line 37: We need to change these transactions to `deposit, transfer and withdraw`.

common-files/utils/websocket.mjs
```diff
@@ -0,0 +1,100 @@
+/* ignore unused exports */
```
druiz0992, line 1: There has been a problem with the rebase.. this file should be deleted

nightfall-client/src/services/withdraw.mjs
```diff
@@ -8,73 +8,103 @@ It is agnostic to whether we are dealing with an ERC20 or ERC721 (or ERC1155).
...
-  const [oldCommitment] = (await findUsableCommitmentsMutex(
-    compressedZkpPublicKey,
+  const shieldContractInstance = await waitForContract(SHIELD_CONTRACT_NAME);
+
+  const maticAddress = await shieldContractInstance.methods.getMaticAddress().call();
+
+  const withdrawValue = value.bigInt > MAX_WITHDRAW ? MAX_WITHDRAW : value;
+  const commitmentsInfo = await getCommitmentsValues(
+    withdrawValue.bigInt,
+    [],
```
IlyasRidhuan, line 38: These are really hard to parse with all the empty arrays, can we convert this into a named object so we don't forget what all these parameters are meant to represent?

nightfall-client/src/utils/getCommitmentValues.mjs
```diff
@@ -0,0 +1,114 @@
...
+  let salts;
+
+  let oldCommitments = [];
+  let newCommitments = [];
+  let leafIndices = [];
+  let localSiblingPaths = [];
+  let blockNumberL2s = [];
+  let roots = [];
+  let nullifiers = [];
+  if (!isFee || totalValueToSend > 0n) {
```
IlyasRidhuan, line 37: This function a bit unclear with all the `lets` and this `if`.

Instead of wrapping all the actual work inside this `if` can we just check the opposite condition and returns the empty array early, i.e. `if(isFee && totalValueToSend <= 0n)`.

Then we won't need all these `lets`.

IlyasRidhuan (changes requested):

nightfall-optimist/src/services/challenges.mjs
```diff
@@ -218,31 +241,45 @@ export async function createChallenge(block, transactions, err) {
...
               Block.buildSolidityStruct(block),
               transactions.map(t => Transaction.buildSolidityStruct(t)),
               currentTxIdx,
               currentNullifierIdx,
+              currentIsNullifierFee,
               Block.buildSolidityStruct(oldBlock),
               oldBlockTransactions.map(t => Transaction.buildSolidityStruct(t)),
               oldTxIdx,
               oldNullifierIdx,
-              salt,
```
IlyasRidhuan, line 245: I believe we still need salt here for the challenge

nightfall-optimist/src/services/challenges.mjs
```diff
@@ -260,7 +297,6 @@ export async function createChallenge(block, transactions, err) {
             priorBlockTransactions.map(t => Transaction.buildSolidityStruct(t)), // the transactions in the prior block
             Block.buildSolidityStruct(block),
             transactions.map(t => Transaction.buildSolidityStruct(t)),
-            salt,
```
IlyasRidhuan, line 263: salt needed here

wallet/src/common-files/classes/transaction.js
```diff
@@ -154,7 +189,9 @@ class Transaction {
       commitments,
       nullifiers,
       compressedSecrets,
-      proof: arrayEquality(proof, [0, 0, 0, 0, 0, 0, 0, 0]) ? [0, 0, 0, 0] : compressProof(proof),
+      commitmentFee,
+      nullifiersFee,
+      proof: compressProof(proof),
```
IlyasRidhuan, line 194: This proof field needs to look like the one in `master`

IlyasRidhuan (changes requested):

druiz0992 (approved):

cli/lib/nf3.mjs
```diff
@@ -439,7 +438,7 @@ class Nf3 {
     value,
     tokenId,
     compressedZkpPublicKey,
-    fee = this.defaultFee,
+    fee = this.defaultFeeMatic,
```
druiz0992, line 441: Where is this fee used?

cli/lib/nf3.mjs
```diff
@@ -501,7 +500,7 @@ class Nf3 {
     value,
     tokenId,
     recipientAddress,
-    fee = this.defaultFee,
+    fee = this.defaultFeeMatic,
```
druiz0992, line 503: where is this fee used?

IlyasRidhuan (approved):

druiz0992 (approved):

nightfall-client/src/utils/getCommitmentInfo.mjs
```diff
@@ -0,0 +1,139 @@
...
+    // Commitment Tree Information
+    const commitmentTreeInfo = await Promise.all(oldCommitments.map(c => getSiblingInfo(c)));
+    const localSiblingPaths = commitmentTreeInfo.map(l => {
+      const path = l.siblingPath.path.map(p => p.value);
+      return generalise([l.root].concat(path.reverse()));
+    });
+    const leafIndices = commitmentTreeInfo.map(l => l.leafIndex);
+    const blockNumberL2s = commitmentTreeInfo.map(l => l.isOnChain);
+    const roots = commitmentTreeInfo.map(l => l.root);
+    logger.info(
```
druiz0992, line 116: Construction doesnt work. Change to -> 
    logger.info(
      `Constructing transfer transaction with blockNumberL2s ${blockNumberL2s} and roots ${roots}`);

PR #865, opened 2022-08-10
fix: solve edge case in which user is doing a single transfer of mati…

RogerTaule:
…c and paying the fee with matic

### DIFF
--- a/nightfall-client/src/services/transfer.mjs
+++ b/nightfall-client/src/services/transfer.mjs
@@ -52,9 +52,12 @@ async function transfer(transferParams) {
 
   const maticAddress = await shieldContractInstance.methods.getMaticAddress().call();
 
+  const addedFee = maticAddress === ercAddress ? fee.bigInt : 0;
+
   const totalValueToSend = values.reduce((acc, value) => acc + value.bigInt, 0n);
   const commitmentsInfo = await getCommitmentInfo({
     transferValue: totalValueToSend,
+    addedFee,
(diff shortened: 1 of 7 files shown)


nightfall-client/src/utils/getCommitmentInfo.mjs
```diff
@@ -19,23 +19,42 @@ const { BN128_GROUP_ORDER } = config;
...
     tokenId,
-    generalise(transferValue),
+    generalise(transferValue + addedFee),
   );
-  if (oldCommitments) logger.debug(`Found commitments ${JSON.stringify(oldCommitments, null, 2)}`);
-  else {
+  if (oldCommitments) {
+    if (addedFee > 0) feeIncluded = true;
+    logger.debug(`Found commitments ${JSON.stringify(oldCommitments, null, 2)}`);
+  } else if (addedFee > 0) {
```
IlyasRidhuan, line 43: This is a duplicate of the above conditional

nightfall-client/src/services/transfer.mjs
```diff
@@ -63,7 +66,7 @@ async function transfer(transferParams) {
   });
 
   const commitmentsInfoFee =
-    fee === 0
+    fee === 0 || (maticAddress === ercAddress && commitmentsInfo.feeIncluded)
```
IlyasRidhuan, line 69: I think `commitmentsInfo.feeIncluded` is the only condition needed

PR #867, opened 2022-08-10
Luizoamorim/feat save commitments

luizoamorim:
This PR add an endpoint to saveCommitments functionality and also a function in the service to do the data manupulation.

### DIFF
--- a/nightfall-client/src/routes/commitment.mjs
+++ b/nightfall-client/src/routes/commitment.mjs
@@ -14,6 +14,7 @@ import {
   getWalletPendingSpentBalance,
   getCommitments,
   getCommitmentsByCompressedZkpPublicKeyList,
+  saveCommitments,
 } from '../services/commitment-storage.mjs';
 
 const router = express.Router();
--- a/nightfall-client/src/routes/commitment.mjs
+++ b/nightfall-client/src/routes/commitment.mjs
@@ -91,6 +91,25 @@ router.get('/commitments', async (req, res, next) => {
 });
 
 /**
+ * @description the endpoint that will save a list of commitments
+ * @author luizoamorim
+ */
+router.post('/saveAll', async (req, res, next) => {
+  logger.debug('commitment/saveAll endpoint received POST');
+  const listOfCommitments = req.body;
+  try {
+    const response = await saveCommitments(listOfCommitments);
+    if (response instanceof Error) {
+      throw response;
+    }
+    res.json({ response });
+  } catch (err) {
+    logger.error(err);
+    next(err);
+  }
+});
+
+/**
  * @description the endpoint that will send a reponse with all the
  * existent commitments for the list of compressedPkd received in the
  * request body. We're using POST for this endpoint, because if the
--- a/nightfall-client/src/services/commitment-storage.mjs
+++ b/nightfall-client/src/services/commitment-storage.mjs
@@ -706,6 +706,35 @@ export async function findUsableCommitmentsMutex(
 }
 
 /**
+ *
+ * @function saveCommitments save a list of commitments in the database
+ * @param {[]} listOfCommitments a list of commitments to be saved in the database
+ * @returns
+ */
+export async function saveCommitments(listOfCommitments) {
+  const connection = await mongo.connection(MONGO_URL);
+  const db = connection.db(COMMITMENTS_DB);
+  let exist = false;
+  for (const commitment of listOfCommitments) {
+    if (
+      // eslint-disable-next-line no-await-in-loop
+      (await db.collection(COMMITMENTS_COLLECTION).find({ _id: commitment._id }).toArray()).length >
+      0
+    ) {
+      exist = true;
+      break;
+    }
+  }
+
+  if (exist) {
+    return new Error('Some of these commitments already existis in the database!');
+  }
+
+  const response = await db.collection(COMMITMENTS_COLLECTION).insertMany(listOfCommitments);
+  return response;
+}
+
+/**
  * @function getCommitmentsByCompressedZkpPublicKeyList do the role of a service taking care of the
  * business logic and of a repository doing the communication with the database for this
  * use case.


nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -705,6 +705,35 @@ export async function findUsableCommitmentsMutex(
...
+      // eslint-disable-next-line no-await-in-loop
+      (await db.collection(COMMITMENTS_COLLECTION).find({ _id: commitment._id }).toArray()).length >
+      0
+    ) {
+      exist = true;
+      break;
+    }
+  }
+
+  if (exist) {
```
druiz0992, line 729: Why do we raise error if duplicated commitment? Can’t you just not copy that one?

druiz0992 (changes requested):

imagobea (changes requested):
Hello @luizoamorim

Apologies for the late feedback. As discussed:

- We have implemented a number of validations in the SDK, but perhaps they belong in the API in case other requests are sent from different services

- In standup today we confirmed that an invalid commitment can't be spent, this means we might want to bulk insert all commitments (after proving that the "format" of the commitment is correct). At the same time, I'm not familiar with MongoDB, so might be good to check what would happen in the event of trying to insert an existing id.

- Once we have addressed the above and the logic is clear, we should double-check the route (we prob don't need lines 103 - 105)

Many thanks :peace_symbol:

nightfall-client/src/services/commitment-storage.mjs
luizoamorim: I'm thinking to run a map in the list of commitments and just get those that doesn't exists in the database yet.
This way we just save new commitments.
Because the insertMany operation save all until the find the first that already exists.

luizoamorim:
@daveroga @imagobea @israelboudoux 

I decided to follow the logic of remove the existent commitments found in database. The other using upsert I didn't had success.
So I'm following the next steps:

```
/**
   * 1. listOfCommitments => get only the ids
   */
  const commitmentsIds = listOfCommitments.map(commitment => commitment._id);
  /**
   * 2. Find commitments that already exists in DB
   */
  const commitmentsFound = await db
    .collection(COMMITMENTS_COLLECTION)
    .find({ _id: { $in: commitmentsIds } })
    .toArray();

  /**
   * 3. remove the commitments found in the database from the list
   */
  const onlyNewCommitments = listOfCommitments.filter(
    commitments =>
      commitmentsFound.find(commitmentFound => commitmentFound.id === commitments.id) === undefined,
  );

  if (onlyNewCommitments.length > 0) {
    /**
     * 4. Insert all
     */
    await db.collection(COMMITMENTS_COLLECTION).insertMany(onlyNewCommitments);

    /**
     * 5. Sycronize from beggining
     */
    await syncState();

    return { successMessage: 'Commitments have been saved successfully!' };

File: nightfall-client/src/event-handlers/block-proposed.mjs [hunk 2]
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-cycle */
```
imagobea, line 1: :eyes: I can see this rule has been added to several files in the PR, is this needed? What was the error??

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -753,14 +755,61 @@ export async function findUsableCommitmentsMutex(
   );
 }
 
+/**
+ *
+ * @function saveCommitments save a list of commitments in the database
```
imagobea, line 760: I suggest to rename the function as e.g. `insertCommitmentsAndResync` (?)

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -753,14 +755,61 @@ export async function findUsableCommitmentsMutex(
...
+ * @param {[]} listOfCommitments a list of commitments to be saved in the database
+ * @returns if all the commitments in the list already exists in the database
+ * throw an error, else return a success message.
```
imagobea, line 763: You can use the `@throws {Error}` to explain the "throw error" bit

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -753,14 +755,61 @@ export async function findUsableCommitmentsMutex(
...
+ */
+export async function saveCommitments(listOfCommitments) {
+  const connection = await mongo.connection(MONGO_URL);
+  const db = connection.db(COMMITMENTS_DB);
+
+  /**
+   * 1. listOfCommitments => get only the ids
+   */
```
imagobea, line 771: Make the comments less verbose

```suggestion
// listOfCommitments => get only the ids
```

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -753,14 +755,61 @@ export async function findUsableCommitmentsMutex(
...
+  const commitmentsIds = listOfCommitments.map(commitment => commitment._id);
```
imagobea, line 772:
```suggestion
  const commitmentIds = listOfCommitments.map(commitment => commitment._id);
```

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -753,14 +755,61 @@ export async function findUsableCommitmentsMutex(
...
+  /**
+   * 2. Find commitments that already exists in DB
+   */
+  const commitmentsFound = await db
```
imagobea, line 776:
```suggestion
  const commitmentsFromDb = await db
```

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -753,14 +755,61 @@ export async function findUsableCommitmentsMutex(
...
+
+  /**
+   * 3. remove the commitments found in the database from the list
+   */
+  const onlyNewCommitments = listOfCommitments.filter(
+    commitments =>
+      commitmentsFound.find(commitmentFound => commitmentFound.id === commitments.id) === undefined,
+  );
+
+  if (onlyNewCommitments.length > 0) {
```
imagobea, line 789: Length `0` is falsy already

```suggestion
  if (onlyNewCommitments.length) {
```

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -753,14 +755,61 @@ export async function findUsableCommitmentsMutex(
...
+    .find({ _id: { $in: commitmentsIds } })
+    .toArray();
+
+  /**
+   * 3. remove the commitments found in the database from the list
+   */
+  const onlyNewCommitments = listOfCommitments.filter(
+    commitments =>
+      commitmentsFound.find(commitmentFound => commitmentFound.id === commitments.id) === undefined,
+  );
```
imagobea, line 787: Singular!
Very nice chain of array methods :blush: :clap: 

```suggestion
  const onlyNewCommitments = listOfCommitments.filter(
    commitment =>
      commitmentsFound.find(commitmentFound => commitmentFound.id === commitment.id) === undefined,
  );
```

nightfall-client/src/routes/commitment.mjs
```diff
@@ -90,6 +91,21 @@ router.get('/commitments', async (req, res, next) => {
   }
 });
 
+/**
+ * @description the endpoint that will save a list of commitments
+ */
+router.post('/saveAll', async (req, res, next) => {
+  logger.debug('commitment/saveAll endpoint received POST');
```
imagobea, line 98: If we had a standard API, POST `/commitments` would already signal insertion.

I assume the above is not possible, therefore perhaps a "one-word" route should suffice, e.g. `/save`. This could also be `/insert` or `/create`.

imagobea (approved):
Very good @luizoamorim :woman_dancing: 
I have added some thoughts and recommended some changes, but to me this is exactly what we agreed. Well played!! :long_drum:

daveroga (approved):
LGTM

nightfall-client/src/routes/commitment.mjs
luizoamorim: Changed to save

nightfall-client/src/services/commitment-storage.mjs
luizoamorim: Changed!!

nightfall-client/src/services/commitment-storage.mjs
luizoamorim: Can be! Changed!

nightfall-client/src/services/commitment-storage.mjs
luizoamorim: Make sense! Changed.

nightfall-client/src/services/commitment-storage.mjs
luizoamorim: In this case I prefer leave in plural. Because a list of ids for several commitments not just one.

nightfall-client/src/services/commitment-storage.mjs
luizoamorim: Changed for just 1 line.

nightfall-client/src/services/commitment-storage.mjs
luizoamorim: Can be. Changed.

nightfall-client/src/event-handlers/block-proposed.mjs
luizoamorim: This problem started to appear. I don't know exactly why.

nightfall-client/src/services/commitment-storage.mjs
luizoamorim: Added.

PR #872, opened 2022-08-12
Payments refactor

RogerTaule:
This PR does a refactor of #733. When payments were first implemented, the transaction structure were modified by adding a couple of bytes32 arrays containing the fee nullifiers and the commitments nullifiers. Right now, transactions are limited to a maximum of two nullifiers per transfer / withdrawal, but the transaction contains 4 slots (two for nullifiers itself and two for the nullifiers fee). The idea is to also use the fee slots so that users can create transfers that use up to 4 commitments. For this purpose, the fee arrays are merged into the initial nullifier array, so the transaction will have a 4 slot nullifier array. Same idea applies for the commitment.

This PR also modifies the commitment selection algorithm, so that transfers and withdrawals will be able to use up to 4 slots if possible. A more detailed description about how the algorithm works can be found in **commitment-selection.md**

Circuits are also modified. In order to add this extra flexibility, some constraints needs to be added. The circuits will have now the following constraints:

|  *Operations*  | *Constraints before*  | *Constraints after*   | *Change (%)*   | 
|---|---|---|---|
| Deposit  | 923  | 1,277  | 38.8%  |
| Transfer  |  64,191 |  67,694  | 5.4%  |
| Withdraw | 50,244  | 53,348  | 6.1%  |


ChaitanyaKonda (approved):

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -609,147 +670,349 @@ async function findUsableCommitments(compressedZkpPublicKey, ercAddress, tokenId
...
+  const rankedSubsetCommitmentsArray = possibleSubsetsCommitments
+    .filter(subset => subset.length > 0)
+    .sort((a, b) => {
+      const changeA = a.reduce((acc, com) => acc + com.preimage.value.bigInt, 0n) - value.bigInt;
+      const changeB = b.reduce((acc, com) => acc + com.preimage.value.bigInt, 0n) - value.bigInt;
+      if (changeA - changeB === 0n) {
+        return b.length - a.length;
+      }
+
+      return Number(changeA - changeB);
```
IlyasRidhuan, line 958: I don't think this is safe for `BigInt` due to overflows, (although changes aren't likely to overflow).

This is probably safer `return changeA > change B ? 0 : -1`

nightfall-client/src/services/commitment-storage.mjs
```diff
@@ -609,147 +670,349 @@ async function findUsableCommitments(compressedZkpPublicKey, ercAddress, tokenId
...
+  const rankedSubsetCommitmentsFeeArray = possibleSubsetsCommitmentsFee
+    .filter(subset => subset.length > 0)
+    .sort((a, b) => {
+      const changeA = a.reduce((acc, com) => acc + com.preimage.value.bigInt, 0n) - value.bigInt;
+      const changeB = b.reduce((acc, com) => acc + com.preimage.value.bigInt, 0n) - value.bigInt;
+      if (changeA - changeB === 0n) {
+        return b.length - a.length;
+      }
+
+      return Number(changeA - changeB);
```
IlyasRidhuan, line 989: same as above `return changeA > changeB ? 0 : -1`

nightfall-client/src/utils/computeCircuitInputs.mjs
```diff
@@ -80,7 +68,7 @@ const computePrivateInputsCommitments = (
   recipientPublicKeys,
   isTransfer,
 ) => {
-  const padLength = isTransfer ? 2 : 1;
+  const padLength = isTransfer ? 3 : 2;
```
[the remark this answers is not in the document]
RogerTaule: The reason is that transfers can have up to 3 commitments while withdraws can have only 2

IlyasRidhuan (approved):

PR #873, opened 2022-08-12
ErcAddress and fee address check 

RogerTaule:
This PR addresses a small fix found in the SDK related to ercAddress comparison. To avoid mistakes, both maticAddress and ercAddress are converted to lower case


imagobea (approved):
:boom: Works like a charm @RogerTaule many thanks again for the fix and the help!!

israelboudoux (approved):

Issue #803, opened 2022-07-07
Use makeBlockNow in testing


signorecello:
TL;DR at the end

I'll sing you a song of my people. Once upon a time, nightfall would only create new blocks once the mempool was full. That was enough for the people of old, but was not suitable for production usage. So, great human effort was required in order to produce a function that was given the glorious name of ```makeBlockNow```, which means, in ancient English, "make block now". It's function is to make a block even if the mempool is not full.

For a time, this function was not used in testing, so humanity couldn't use it to make tests run faster. This issue proposes that this function is thereby used for testing.

TL;DR: Start using ```makeBlockNow``` in tests

Westlad:
Fixed by #875

PR #875, opened 2022-08-16
refactor: updating tests to use make block now

RogerTaule:
This PR solves #803 and implements `makeBlockNow` function for the circuits tests. 
The ganache test run time is reduced by approximately a 40%. 
In addition, tests are can be run now independently (so we can use .skip to only test specific tests). However, I'd strongly suggest modifying even more the tests so that the commitments are deleted afterEach test. This would completely decouple each test and will help to test new circuit features more easily.


signorecello (approved):
Love this. Thank you so much!

druiz0992 (approved):

PR #878, opened 2022-08-16
Update node

### LINKED ISSUE #876: upgrade Node to latest lts
We should be using node v16 and npm 8 to get the latest features.  The upgrade needs to happen in the containers.  The `package-lock.json` files should be upgraded to the latest version via `npm upgrade` and dependencies resolved.

Westlad:
fixes #876 
Updates node to v16.16 (latest lts) and npm to 8.17.

Use of node 8 enables vulnerable dependencies to be overridden with the `overrides` keyword in `package.json`.  This is required to address a vulnerability in `got`; `web3js` depends on a vulnerable version.


daveroga:
Having problems with swarm-js dependency in web3. Also referenced in other projects like this https://github.com/ChainSafe/web3.js/issues/5315

israelboudoux:
We are still having issues regarding dependency vulnerabilities because a transient dependency `web3 -> web3-bzz -> swarm-js` that isn't maintained anymore and has a dep on a **got** vulnarable version. There is a pending [PR](https://github.com/VictorTaelin/swarm-js/pull/42) in the `swarm-js` repo which would solve this, I've added a comment referring to the maintainers to see if they can merge it.

pawelgrzybek:
Looks like we have two very similar PRs.
https://github.com/EYBlockchain/nightfall_3/pull/908

Westlad:
Oh blast, I forgot I opened this before I went on leave.  I'll close it as #908 is more current and has @pawelgrzybek 's suggestions incorporated.

PR #881, opened 2022-08-18
Improving GitHub Actions tests

israelboudoux:
This PR adds missing tests to Github Actions pipeline:
- Tests under common-files are now part of the pipeline;
- Missing tests under `/test` like `kem-dem.test.mjs` and `timber.test.mjs` were added too;
- A new test @RogerTaule related to circuits was created was included in the pipeline.


signorecello (approved):
LGTM but man we really need some solid testing strategy. I hate all these tests on package.json. Maybe it'd be worth to have a repo just for QA stuff with a bunch of tests in there, and keep only the unit tests here (when we have those). But let's take this offline

israelboudoux:
I'm sorry for the "test" commit messages, I was testing if the pipeline was working. I'll squash the commit accordingly.

PR #891, opened 2022-08-23
Fix restriction withdrawal test

RogerTaule:
Adding some logs and fixing a wrong comparison between numbers and bigInt to fix the restriction withdrawal test


israelboudoux (approved):
LGTM

IlyasRidhuan (approved):

PR #892, opened 2022-08-24
test: replace nf3 with new SDK

imagobea:
FYI the pipelines fail because the dependency exists just locally with `npm link`.
This is because the SDK is written in TS, therefore you need to go through a `build` process before you can use it. The NPM package will be based on this build.

**Testing SDK locally**

1. Clone `nightfall-sdk` and checkout branch `bea/31-guarantee-sdk-usability` which contains the necessary changes for a successful build and import
2. On `nightfall-sdk` run `npm run build` to generate the `dist` folder
3. On `nightfall-sdk` run `npm link` to create a symlink (you can see it if you run `npm list -g`)
4. On `nightfall_3` (NF) run `npm link nightfall-sdk` to start using the SDK (no need to install anything after)

You can check NF `node_modules` to see if the process completed successfully.

**About replacing nf3 in test/e2e/protocol/health-and-contracts.test.mjs**

Some of the functionality does not exist in new SDK:

- `getContractAddress` is not available at `User` level, however, the `Shield` test is possible
- `incomingViewingKeys` not available ar `User` level

Both methods exist but these are "internal calls" while building an instance of `User`. IMO the "non-replaceable" tests could be removed, and we could revamp the suite to test e.g `User` class creation via `UserFactory`.

### DIFF
--- a/test/e2e/protocol/health-and-contracts.test.mjs
+++ b/test/e2e/protocol/health-and-contracts.test.mjs
@@ -3,6 +3,7 @@ import chai from 'chai';
 import config from 'config';
 import chaiHttp from 'chai-http';
 import chaiAsPromised from 'chai-as-promised';
+import { UserFactory } from 'nightfall-sdk';
 import Nf3 from '../../../cli/lib/nf3.mjs';
 
 // so we can use require with mjs file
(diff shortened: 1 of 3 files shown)


test/e2e/protocol/health-and-contracts.test.mjs
```diff
@@ -15,19 +16,29 @@ const { mnemonics, signingKeys } = config.TEST_OPTIONS;
...
 const nf3User1 = new Nf3(signingKeys.user1, environment);
 
+let user;
 before(async () => {
   await nf3User1.init(mnemonics.user1);
+  user = await UserFactory.create({
+    blockchainWsUrl: 'ws://localhost:8546',
+    clientApiUrl: 'http://localhost:8080',
+    ethereumPrivateKey: signingKeys.user1,
+  });
```
imagobea, line 26: Improve (check if the values are available somewhere else).

imagobea:
Continues in #924

PR #894, opened 2022-08-25
Incorrect right-sibling hash

IlyasRidhuan:
When the node index is even, we need to hash the frontier slot with `nodeValue` not with 0 (that is only when the node index is odd).


RogerTaule (commented):
LGTM

RogerTaule (approved):

druiz0992 (approved):

PR #895, opened 2022-08-25
Supporting stMATIC

signorecello:
Quick changes, we need to support another token on Goerli, stMATIC, so LIDO can proceed with their integration

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -390,6 +390,11 @@ module.exports = {
           address: '0x07865c6E87B9F70255377e024ace6630C1Eaa37F',
           amount: process.env.USDC_TESTNET_RESTRICT || '1000000000',
         },
+        {
+          name: 'stMATIC',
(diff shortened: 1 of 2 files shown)


config/default.js
```diff
@@ -390,6 +390,11 @@ module.exports = {
           address: '0x07865c6E87B9F70255377e024ace6630C1Eaa37F',
           amount: process.env.USDC_TESTNET_RESTRICT || '1000000000',
         },
+        {
+          name: 'stMATIC',
+          address: '0x9A7c69A167160C507602ecB3Df4911e8E98e1279',
+          amount: process.env.USDC_TESTNET_RESTRICT || '1000000000',
```
daveroga, line 396: Here the environment variable should be changed for `process.env.stMATIC_TESTNET_RESTRICT`

signorecello: Thanks, you're right, I'll fix it :)

daveroga (approved):
LGTM

RogerTaule (approved):

Issue #877, opened 2022-08-16
Feature Block Explorer: Add endpoint in the nightfall client to return the transactions in the mempool


daveroga:
The Blockchain Explorer needs to know the transactions in the mempool for the current proposer.
We need to create an endpoint in the proposer and in the optimist to know these transactions.

PR #898, opened 2022-08-30
Block Explorer: add mempool endpoint in the proposer

daveroga:
Fixes: #877 

We need to add mempool endpoint in the proposer for the block explorer to know information about pending transactions.

- `proposer/mempool` endpoint created in proposer that calls existing `proposer/mempool` from optimist.


druiz0992 (approved):

RogerTaule (approved):
LGTM

PR #902, opened 2022-09-02
Updating memory arrays to calldata whenever possible

### LINKED ISSUE #819: Change data location of contract input parameters from "memory" to "calldata" if possible
When passing an input parameter to a contract function, we need to define if we want to use "memory" or "calldata" as a data location, being the latest cheaper. This is due to calldata being a non-modifiable storage, which means that once parameters are passed cannot be modified inside the function. In other words, calldata storage can only be used for read only parameters.

Currently, all input parameters are set to "memory". Changing parameters from "memory" to "calldata" whenever possible will result in a reduction of the gas costs of the smart contract actions.

RogerTaule:
This PR closes #819 and uses calldata to store input parameters of solidity calls whenever possible.
It also removes MiMc.sol since it is never used and is deprecated.


daveroga (approved):
LGTM

PR #905, opened 2022-09-06
constants and config refactor

### LINKED ISSUE #890: Constants.json in common-files holds some variables that are not constant in the event of a contract update
https://github.com/EYBlockchain/nightfall_3/blob/chait/dupNullifierError/common-files/constants/constants.json

- PROPOSE_BLOCK_TYPES
- SUBMIT_TRANSACTION_TYPES
- BLOCK_TYPES (might require a contract migrate potentially but it is best to have this configureable too)
- TRANSACTION_TYPES (might require a contract migrate potentially but it is best to have this configureable too)

LijuJoseJJ:
this PR fixes #890


config/default.js
```diff
@@ -429,4 +429,13 @@ module.exports = {
   // assumption is if LOCAL_PROPOSER is true, wallet UI app
   // is running in local machine
   isLocalRun: process.env.LOCAL_PROPOSER === 'true',
+  BLOCK_TYPES: '(uint48,address,bytes32,uint256,bytes32,bytes32)',
```
signorecello, line 432: for organizational sake, can we make these types into a nested object? like:

```
{
...,
"TYPES": {
    "BLOCK": '(uint48,address,bytes32,uint256,bytes32,bytes32)',
    etc
},
...
}
```

signorecello (changes requested):

config/default.js
ChaitanyaKonda: To be specific to their exact function of function/structure signatures, `SIGNATURES` would be very apt

ChaitanyaKonda (approved):

config/default.js
LijuJoseJJ: done

signorecello (approved):
LGTM, thanks for the changes

RogerTaule (changes requested):
The following constants are unused: BLOCK_PROPOSED_EVENT_NAME, ZKP_KEY_LENGTH and NODE_HASHLENGTH.

The following config variables are unused: NULLIFIER_COLLECTION, WALLETS_COLLECTION, PEERS_COLLECTION, PROOF_QUEUE and ELLIGATOR2. 

Could you please remove them? Thanks!

RogerTaule (approved):
LGTM

nightfall-optimist/src/services/block-utils.mjs
```diff
@@ -5,7 +5,7 @@ import Web3 from 'web3';
 // of the block object isn't.  They can thus be called directly when instantiating the Block class
 // would be problematic because of its reliance on the Optimist database.
 
-const { BLOCK_TYPES } = constants;
+const { SIGNATURES } = constants;
```
RogerTaule, line 8: Signatures variable is stored in config and not in constants
```suggestion
const { SIGNATURES } = config;
```

RogerTaule (changes requested):

nightfall-optimist/src/services/block-utils.mjs
LijuJoseJJ: done

RogerTaule (approved):

PR #914, opened 2022-09-12
checkTransaction at the time of block assembly

LijuJoseJJ:
this PR fixes #https://github.com/EYBlockchain/nightfall_3_private/issues/209


IlyasRidhuan (changes requested):

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -62,6 +63,12 @@ async function makeBlock(proposer, number = TRANSACTIONS_PER_BLOCK) {
   // we retrieve un-processed transactions from our local database, relying on
   // the transaction service to keep the database current
   const transactions = await getMostProfitableTransactions(number);
+
+  // check transction before building block
+  transactions.forEach(async transaction => {
+    await checkTransaction(transaction, true);
```
RogerTaule, line 69: Why are you setting checkTransaction parameter inL2AndNotInL2 to true?

RogerTaule, line 69: Why are you setting checkTransaction parameter inL2AndNotInL2 to true?

RogerTaule (changes requested):

nightfall-optimist/src/services/block-assembler.mjs
RogerTaule: In fact, couldn't we remove this parameter? Since duplicate commitments nullifiers is already checked in the circuit, this is no longer needed. If a transaction have a duplicate nullifier / commitment in the same tx the proof verification will fail

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -62,6 +63,12 @@ async function makeBlock(proposer, number = TRANSACTIONS_PER_BLOCK) {
   // we retrieve un-processed transactions from our local database, relying on
   // the transaction service to keep the database current
   const transactions = await getMostProfitableTransactions(number);
+
+  // check transction before building block
+  transactions.forEach(async transaction => {
```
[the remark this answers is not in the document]
LijuJoseJJ: done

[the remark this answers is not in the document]
LijuJoseJJ: done

nightfall-optimist/src/services/block-assembler.mjs
ChaitanyaKonda: > Why are you setting checkTransaction parameter inL2AndNotInL2 to true?
It needs to be `false`.

`In fact, couldn't we remove this parameter? Since duplicate commitments nullifiers is already checked in the circuit, this is no longer needed. If a transaction have a duplicate nullifier / commitment in the same tx the proof verification will fail`
The logic with duplicate commitment or nullifier will get more complicated than it is today. 
For example, during transactionSubmittedEventHandler, if there is a duplicate commitment from a new transaction, it can still be added into the mempool as long as the fee paid for this transaction is higher and the commitment already in DB is not part of L2 block yet. Also, if the commitment already in DB is part of an L2 block but it is part of an invalid transaction, then it this new transaction if it is valid, should be added to DB because this transaction will be proposed after rollback removes the commitment in DB. This logic will be handled in a different issue but since `checkTransaction` will be used at transactionSubmittedEventHandler, blockAssembler, rollbackEventHandler, blockProposedEventHandler, it needs to handle error handling correctly based on flags such as - `isAlreadyInL2` and `isAlreadyInMempool`

Similar complex handling will exist at blockProposed handler stage too. If the commitment from a transaction from block proposed stage is a duplicate of another commitment already in L2 block, then raise a challenge. Instead if it is a duplicate of another commitment already in mempool and not in L2 block, then mark this mempool commitment which is part of a different transaction to `false` too.

I will create a separate issue for this.

ChaitanyaKonda: Meanwhile @LijuJoseJJ please make the following changes.

Replacing `inL2AndNotInL2` with more apt and clear flags of  `isAlreadyInL2` and `isAlreadyInMempool`. For the moment, if a new commitment or nullifier received already exists in L2  or mempool during `TransactionSubmittedEventHandler`, we want to detect and not save this new transaction received. For BlockProposedEventHandler, RollbackEventHandler, BlockAssembly, we only care if the new commitment or nullifier received is in L2 and handle according to the context. The following changes should help

- Replace, `inL2AndNotInL2` (terrible and confusing name) with `isAlreadyInL2` and `isAlreadyInMempool` in `checkTranasction` 
- Flag values
  - BlockProposedEventHandler
    - isAlreadyInL2: true
    - isAlreadyInMempool: false
  - RollbackEventHandler
    - isAlreadyInL2: true
    - isAlreadyInMempool: false
  - TransactionSubmittedEventHandler
    - isAlreadyInL2: true
    - isAlreadyInMempool: true
  - BlockAssembly
    - isAlreadyInL2: true
    - isAlreadyInMempool: false
 - Replace  ` const query = inL2AndNotInL2
    ? { nullifiers: nullifierHash }
    : {
        nullifiers: nullifierHash,
        blockNumberL2: { $gte: -1, $ne: blockNumberL2OfTx },
      };` in `getL2TransactionByNullifier` with  `const query = (isAlreadyInL2 && isAlreadyInMempool)
    ? { nullifiers: nullifierHash }
    : {
        nullifiers: nullifierHash,
        blockNumberL2: { $gte: -1, $ne: blockNumberL2OfTx },
      };`
 - Replace   `const query = inL2AndNotInL2
    ? { commitments: commitmentHash }
    : {
        commitments: commitmentHash,
        blockNumberL2: { $gte: -1, $ne: blockNumberL2OfTx },
      };` in `getL2TransactionByCommitment` with `const query = (isAlreadyInL2 && isAlreadyInMempool)
    ? { commitments: commitmentHash }
    : {
        commitments: commitmentHash,
        blockNumberL2: { $gte: -1, $ne: blockNumberL2OfTx },
      };`
 - replace `inL2AndNotInL2` in `TransactionError` in `checkDuplicateCommitment` and `checkDuplicateNullifier` with `isAlreadyInL2`

ChaitanyaKonda (changes requested):
As commented above

nightfall-optimist/src/services/block-assembler.mjs
LijuJoseJJ: done

israelboudoux (dismissed):
Hello @LijuJoseJJ , would you mind add some unit tests re this change please?
Thank you!

nightfall-optimist/src/services/database.mjs
```diff
@@ -419,35 +419,37 @@ export async function clearBlockNumberL1ForTransaction(transactionHashL1) {
...
-        commitments: commitmentHash,
-        blockNumberL2: { $gte: -1, $ne: blockNumberL2OfTx },
-      };
+  const query =
+    isAlreadyInL2 && isAlreadyInMempool
+      ? { commitments: commitmentHash }
+      : {
+          commitments: commitmentHash,
+          blockNumberL2: { $gte: -1, $ne: blockNumberL2OfTx },
+        };
```
RogerTaule, line 433: I still find that isAlreadyInL2 and isAlreadyMempool names are confusing. What the function is doing is checking if the transaction can be found in L2 or in the mempool. Therefore, I would suggest changing the names for more clear ones such as "checkL2" and "checkMempool". 

On the other hand, I don't fully understand how the query is built. From my understanding, there's no scenario in which the function will be called with both flags set to false. However, by default both parameters are set to false. Moreover, right now it is always checking if it is in L2 and optionally if it is also in mempool. Therefore, the else condition is not only handling the case true / false but also two additional undesired cases.

I think it would probably be easier to remove "isAlreadyInL2" flag. If so, then probably a more appropiate name for the other flag would be "alsoCheckMempool". Then, if in a future PR we want to just check if a transaction is in the mempool and not in L2 we can create a different function that only checks for a duplicate tx in the mempool.

RogerTaule (changes requested):

nightfall-optimist/src/services/database.mjs
LijuJoseJJ: hi @RogerTaule 
so flag "isAlreadyInL2" and "isAlreadyInMempool" are only used in two functions `getL2TransactionByCommitment` and `getL2TransactionByNullifier` and the purpose of both of these function is to query transaction collection by commitment and nullifier respectively.
And for now, functions `getL2TransactionByCommitment` and `getL2TransactionByNullifier` are only called in checkTransaction (functions checkDuplicateCommitment and checkDuplicateNullifier) specifically to fetch tx if there any which holds same commitment or nullifier

@ChaitanyaKonda  please correct if I am wrong
so understanding is user can submit multiple tx for same commitment (for better gas so that current proposer can pick and submit, but this not possible for now by checkTransaction call at time of transaction submit). so any given time same commitment or nullifier in multiple tx can present in mempool and one such tx in L2.

Note: `removeTransactionsFromMemPool` is called twice (in block assembly and in block proposed

checkTransaction are called in four case
1. Transaction submit
        isAlreadyInL2 = true & isAlreadyInMempool = true    (regardless where tx is fetch if it exist)
2. block assembly
        isAlreadyInL2 = true & isAlreadyInMempool = false  (look tx only in L2)
3. block proposed
        isAlreadyInL2 = true & isAlreadyInMempool = false  (look tx only in L2)
4. rollback
         isAlreadyInL2 = true & isAlreadyInMempool = false (look tx only in L2)

RogerTaule: So, I understand the logic, but since you are always checking in L2 and the only that changes is if you are checking also the mempool, why do we need the first flag? I mean, it would be enough with a flag that specifies if you want to check just in L2 or also check the mempool. I think it makes more sense done this way because the case false false will never exist and currently there's no case in which you are checking for a tx in the mempool but not in L2. If in a future this case arises, we could simply create another function.

I still think those names are very confusing. isAlreadyInL2 doesn't mean that the tx is in L2 but that you are looking IF the transaction is already in L2. Therefore, as suggested in my above comment, I would change the names with others such as checkL2 or checkMempool

LijuJoseJJ: I kind of feel `check` prefix to variable name is usually should go for function name which does the check job.
And I have used `is` prefix for the boolean flags
@ChaitanyaKonda can we have your take on it.

regarding case [false  & false]  i also believe such case never will be handled.
but for [false & true] case when we want all/one tx in mempool of particular (commitment or nullifier)
apologies @ChaitanyaKonda  since you have created this issue and suggested these changes
can we have your thought on this as well whenever you are free

nightfall-optimist/src/event-handlers/transaction-submitted.mjs
```diff
@@ -71,7 +71,7 @@ async function transactionSubmittedEventHandler(eventParams) {
       saveTransaction({ ...transaction });
     }
 
-    await checkTransaction(transaction, true);
+    await checkTransaction(transaction, { isAlreadyInL2: true, isAlreadyInMempool: true });
```
druiz0992, line 74: Transactions are now checked at reception and during block assembly. I understand why it’s necessary to do as late as possible. But then, can we just remove the check when transactions are received?

nightfall-optimist/src/services/check-block.mjs
```diff
@@ -218,7 +218,12 @@ export async function checkBlock(block, transactions) {
   try {
     for (let i = 0; i < transactions.length; i++) {
       transaction = transactions[i];
-      await checkTransaction(transaction, false, { blockNumberL2: block.blockNumberL2 }); // eslint-disable-line no-await-in-loop
+      // eslint-disable-next-line no-await-in-loop
+      await checkTransaction(
```
druiz0992, line 222: Why do we need to check transactions from blocks that we have proposed? If we have done checks during block assembly, why do we need to do them again?

nightfall-optimist/src/event-handlers/transaction-submitted.mjs
LijuJoseJJ: @druiz0992 
I think intention is to keep check at the time receive is because, consider a case where proposer is a bad proposer who create a bad-block(overriding the code of check of block assembly - for example) we still need transaction check at receive event-handler in optimist. also if you notice once check fails only we trigger create challenge (is challenger exist in optimist)

nightfall-optimist/src/services/check-block.mjs
LijuJoseJJ: @druiz0992 
same as above comment - case when proposer is bad-proposer

> I think intention is to keep check at the time receive is because, consider a case where proposer is a bad proposer who create a bad-block(overriding the code of check of block assembly - for example) we still need transaction check at receive event-handler in optimist. also if you notice once check fails only we trigger create challenge (is challenger exist in optimist)

RogerTaule (approved):

Westlad (approved):

PR #917, opened 2022-09-12
Optimizing challenges

RogerTaule:
This PR adds a series of optimizations for the current challenges. It does the following changes:

- leafCount is no longer the amount of commitments in the merkle tree before insertion. With this change, leafCount is now the amount of commitments **after** insertion.
- The siblingPath info is stored for the transactions merkle tree.
- The whole set of transactions contained in a block is no longer sent as part of the calldata in challengeDuplicateNullifier and challengeDuplicateCommitment. Instead, we verify that the transaction is real by verifying its sibling path.
- For the challengeProofVerification, there's no need of sending the whole transactions contained in a block. We just need to check that the historicBlock is real. A new function that only validates the block is created in State.sol to do that. 
- The way check-block was detecting duplicated commitments / nullifiers was outdated since it was relying in the fact that each transaction has two commits and two nullifiers, which is no longer true. The corresponding functions has been fixed.
- Transaction checker does no longer check for duplicate nullifiers or commitments in the same transaction, since this is no possible because the circuits will not allow it. Hence, it makes no sense checking it.

### DIFF
--- a/nightfall-deployer/contracts/Challenges.sol
+++ b/nightfall-deployer/contracts/Challenges.sol
@@ -248,23 +248,29 @@ contract Challenges is Stateful, Key_Registry, Config {
         challengeAccepted(transactionBlock.blockL2);
     }
 
-    function challengeHistoricRoot(
-        Block calldata blockL2,
-        Transaction[] calldata transactions,
-        uint256 transactionIndex
-    ) external onlyBootChallenger {
+    function challengeHistoricRoot(TransactionBlock calldata transactionBlock)
+        external
+        onlyBootChallenger
+    {
         checkCommit(msg.data);
-        state.areBlockAndTransactionsReal(blockL2, transactions);
-        require(
-          (
-            transactions[transactionIndex].historicRootBlockNumberL2[0] >= state.getNumberOfL2Blocks() ||
-            transactions[transactionIndex].historicRootBlockNumberL2[1] >= state.getNumberOfL2Blocks() ||
-            transactions[transactionIndex].historicRootBlockNumberL2[2] >= state.getNumberOfL2Blocks() ||
-            transactions[transactionIndex].historicRootBlockNumberL2[3] >= state.getNumberOfL2Blocks()
-          ),
-          'Historic roots are not greater than L2BlockNumber on chain'
+        state.areBlockAndTransactionReal(
+            transactionBlock.blockL2,
+            transactionBlock.transaction,
+            transactionBlock.transactionIndex,
+            transactionBlock.transactionSiblingPath
         );
-        challengeAccepted(blockL2);
+        require(
+            (transactionBlock.transaction.historicRootBlockNumberL2[0] >=
+                state.getNumberOfL2Blocks() ||
+                transactionBlock.transaction.historicRootBlockNumberL2[1] >=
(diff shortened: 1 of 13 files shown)


israelboudoux (changes requested):
Hello @RogerTaule, would you mind add some unit tests re this change please?
Thank you!

nightfall-deployer/contracts/Challenges.sol
```diff
@@ -85,34 +89,45 @@ contract Challenges is Stateful, Key_Registry, Config {
...
+        );
+        state.areBlockAndTransactionReal(
+            transaction2.blockL2,
+            transaction2.transaction,
+            transaction2.transactionIndex,
+            transaction2.transactionSiblingPath
+        );
+
+        require(
+            transaction1.blockL2.blockNumberL2 != transaction2.blockL2.blockNumberL2,
```
IlyasRidhuan, line 115: I don't think this is correct - this would make duplicate commitments in different transactions but in the same block un-challengeable.

IlyasRidhuan: The requirement if both transactions are in the same block is that they don't point to the same transaction (i.e. txIndex1 != txIndex2)

nightfall-deployer/contracts/Challenges.sol
```diff
@@ -122,100 +137,101 @@ contract Challenges is Stateful, Key_Registry, Config {
...
```
IlyasRidhuan, line 163: Same here as for commitment

nightfall-deployer/contracts/Challenges.sol
```diff
@@ -122,100 +137,101 @@ contract Challenges is Stateful, Key_Registry, Config {
...
             [uint256(0), 0, 0, 0],
             super.getMaticAddress()
         );
 
-        if (uint256(transactions[transactionIndex].nullifiers[0]) != 0) {
-            state.areBlockAndTransactionsReal(
-                blockL2ContainingHistoricRoot[0],
-                transactionsOfblockL2ContainingHistoricRoot[0]
-            );
+        if (uint256(transaction.transaction.nullifiers[0]) != 0) {
```
IlyasRidhuan, line 201: Can this sequence of checks just be compressed into a `for loop` that goes to `nullifiers.length`

nightfall-deployer/contracts/Challenges.sol
```diff
@@ -224,31 +240,36 @@ contract Challenges is Stateful, Key_Registry, Config {
...
-        challengeAccepted(blockL2);
+        challengeAccepted(transaction.blockL2);
     }
 
-    function challengeHistoricRoot(
-        Block calldata blockL2,
-        Transaction[] calldata transactions,
-        uint256 transactionIndex
-    ) external onlyBootChallenger {
+    function challengeHistoricRoot(TransactionInfoBlock calldata transaction)
```
IlyasRidhuan, line 251: I know this wasn't part of the PR, but can we change this to `challengeHistoricRootBlockNumber` - there  might a reference that has to be updated in `optimist`

IlyasRidhuan (changes requested):

IlyasRidhuan (approved):

israelboudoux (approved):

PR #923, opened 2022-09-13
David/ping pong update

druiz0992:
Changes required to make #823 work in deployment:
- Applied missing changes to `apps/challenger` to be consistent with `app/proposer`
- Remove `registerChallenger` from `apps/challenger/src/challenger.mjs` as its not needed anymore
- Remove `log(environment)` from apps as it reveals private information
- Remove unnecessary nf3 log
- Fix typo `MULTISIG_APPROVERS` in config
- Add correct initial values to AWS deployment in config
- Fix logs in ping-pong test
- Raise error during proof generation if witness is not found

### DIFF
--- a/apps/challenger/src/app.mjs
+++ b/apps/challenger/src/app.mjs
@@ -1,40 +1,20 @@
 /* ignore unused exports */
+/* eslint-disable import/no-unresolved */
+
 import express from 'express';
 import bodyParser from 'body-parser';
 import cors from 'cors';
 import fileUpload from 'express-fileupload';
 import config from 'config';
-import Nf3 from '../../../cli/lib/nf3.mjs';
 import startChallenger from './challenger.mjs';
+import Nf3 from '../cli/lib/nf3.mjs';
 
-const { SIGNING_KEY, CHALLENGER_PORT } = config;
-const {
-  CLIENT_URL = '',
-  OPTIMIST_HTTP_URL = '',
-  OPTIMIST_WS_URL = '',
-  BLOCKCHAIN_URL = '',
-  PROPOSER_URL = '',
-} = process.env;
-const environment = {
-  clientApiUrl:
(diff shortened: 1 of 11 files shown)


signorecello (approved):
LGTM, thanks for all the help man :)

nightfall-optimist/src/routes/proposer.mjs
```diff
@@ -67,14 +69,22 @@ router.post('/register', async (req, res, next) => {
         logger.warn(
           'Proposer was already registered on the blockchain but not with this Optimist instance - registering locally',
         );
-        const stateContractInstance = await waitForContract(STATE_CONTRACT_NAME);
-        const currentProposer = await stateContractInstance.methods.getCurrentProposer().call();
+        stateContractInstance = await waitForContract(STATE_CONTRACT_NAME);
+        currentProposer = await stateContractInstance.methods.getCurrentProposer().call();
```
israelboudoux, line 73: Hey @druiz0992, why do not reuse the same instances from the lines 60-61?

israelboudoux (approved):
LGTM, just a minor thing to see.

nightfall-optimist/src/routes/proposer.mjs
druiz0992: You’re right, thanks

apps/challenger/src/app.mjs
```diff
@@ -53,6 +33,8 @@ app.use(
 );
 
 app.get('/healthcheck', (req, res) => res.sendStatus(200));
+if (!CHALLENGER_PORT) throw new Error('Please specify a challenger port');
```
pawelgrzybek, line 36: Will this exception ever happen? On line `12` we specify a fallback value of `8192` in case the port is not defined. Either we should get rid of the default value from line 12, or remove this condition.

druiz0992: you're right. Thanks

PR #929, opened 2022-09-13
Containerizing proposer

signorecello:
This PR proposes that the proposer (pun intended) is containerized so all its paths match with our deployment repo. A new script is born that simply runs this new `docker-compose` file

### DIFF
--- a/docker-compose.apps.yml
+++ b/docker-compose.apps.yml
@@ -1,0 +1,38 @@
+version: '3.5'
+# Use this script for making nightfall_3 use stubs.
+services:
+
+  proposer:
+    build:
+      dockerfile: proposer.Dockerfile
+      context: .
+    ports:
+      # to use with postman and etc
+      - 8083:8080
+    environment:
+      BLOCKCHAIN_WS_HOST: blockchain
+      BLOCKCHAIN_PORT: 8546
+      ENABLE_QUEUE: 1
+      OPTIMIST_HOST: optimist
+      OPTIMIST_PORT: 80
+      OPTIMIST_WS_PORT: 8080
+      CLIENT_HOST: client
+      CLIENT_PORT: 8080
+    networks:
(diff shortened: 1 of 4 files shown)


druiz0992 (changes requested):
proposer.Dockerfile is based on node:14. Please set it to node:16

start-apps
```diff
@@ -0,0 +1,2 @@
+#! /bin/bash
+docker-compose -f docker-compose.apps.yml up proposer
```
israelboudoux, line 2: Hey @signorecello, what about the Challenger?

druiz0992: i think we can add a separate PR for challenger. Proposer is urgent

druiz0992 (approved):

israelboudoux (approved):

docker-compose.apps.yml
```diff
@@ -0,0 +1,38 @@
...
+# Use this script for making nightfall_3 use stubs.
+services:
+
+  proposer:
+    build:
+      dockerfile: proposer.Dockerfile
+      context: .
+    ports:
+      # to use with postman and etc
+      - 8083:8080
```
druiz0992, line 11: There seems to be a port inconsistency with `app.mjs` where port is 8092

druiz0992 (changes requested):
I launched proposer and was not able to do `curl http://localhost:8092/healthcheck`

druiz0992:
it would be quite useful to add a proposer test that launches containers checks healthcheck, and exits. This way, we always guarantee proposer is working

signorecello:
> 

Hey I couldn't agree more, but I think it should be part of another ticket, as some team members need this container to work ASAP

signorecello:
Added #934 and will work on that ASAP

druiz0992 (approved):

Issue #714, opened 2022-06-02
Fast wallet synchronization


druiz0992:
Current, wallet requires the complete block information sent by lambda function `syncBlockProd`. It would be enough to simple obtain timber data stored in DynamoDb Table

Issue #523, opened 2022-02-23
Add Rollback handler to browser events


IlyasRidhuan:
The current infrastructure can only stream blockProposed events back to the browser.

This needs to be expanded to include Rollback (once we can trigger challenges in the testnet environment).

PR #930, opened 2022-09-13
wallet - fast wallet synchronization

eugypalu:
Resolves https://github.com/EYBlockchain/nightfall_3/issues/523
Resolves https://github.com/EYBlockchain/nightfall_3/issues/714
Resolves https://github.com/EYBlockchain/nightfall_3/issues/723

Issue 523 Add Rollback handler to browser events

To receive rollback and blockproposed events from the publisher, the wallet subscribes to the topics rollback and blockProposed, both with QOS = 2.
When a rollback type message is received on the topic rollback, the rollbackeventhandler is invoked.

Issue 723 Block proposed and Rollback removal event is not being handled in syncState

The blockproposed and rollback events have been handled in such a way that they do not interfere with lambda synchronization. When the wallet is opened, it connects to the broker and establishes a clean connection to receive only the new blockproposed events, while restoring a persistent connection on the rollback topic to receive previous events as well.
In the case of blockproposed, new events are handled if and only if the wallet has completed synchronization via the lambda, preventing you from saving the same block twice.
Rollback events are handled only if the database contains a block greater than or equal to the one communicated in the rollback.

Issue 714 Fast wallet synchronization

Wallet synchronization is divided into two parts: 
(1) that which occurs through the publisher and notifies the wallet of all rollback or blockproposed events received 
(2) that which occurs through the lambda.

The previous solution consisted of a lambda that communicated all blocks to the wallet block by block via ws. The wallet then extracted timber from these blocks. To work properly, the wallet had to wait for all of the blocks to be received.

In the new solution, the lambda is invoked from the wallet via a Rest API of type GET. Using the AWS API Gateway GET method increases the size of the message (from 256kb of ws to 6mb). The wallet then calls the GET, passing as a parameter: 

The environment in which it is working (preprod, prod, etc.) 
The final synchronized timber block
The last synchronized L2Block (initially coincides with the timber block).
The goal is to have the wallet ready as soon as possible, so the plan is to communicate all timber blocks first, then synchronize all other blocks once the timber synchronization is complete.

Edge Case: 
If the wallet is brand new, the call returns only the most recent timber block.

If the size of the timber + L2Blocks message exceeds the maximum size of the message that can be sent via lambda, the call will return the timber blocks first, then the missing timber blocks (if not fully synchronized) and the L2 blocks when called.

Previously, the publisher would send rollback/blockproposed events to all connected wallets. This caused a bottleneck if multiple wallets were connected.
Instead of a direct wallet-publisher connection, the new solution employs the pubsub paradigm. As a result, the bottleneck shifts from the publisher to the broker. When the publisher receives a rollback or blockproposed notification, it communicates it to the broker, who then distributes it to all wallets.

To receive messages, each wallet establishes two separate connections to the broker, one for each subscribed topic. In the case of the topic rollback, the wallet establishes a persistent connection, allowing it to receive rollback events even when offline.
In the case of the topic BlockProposed, the wallet maintains a non-persistent connection over time, preventing it from receiving events while offline. This is because block synchronization occurs via lambda, whereas rollback comes only from the publisher.


config/default.js
```diff
@@ -5,6 +5,20 @@ function configureAWSBucket() {
   return `${bucket}-${mode}`;
 }
 
+function parseCircuitFilesPath() {
+  let circuits = ['deposit', 'withdraw', 'single_transfer', 'double_transfer'];
```
pawelgrzybek, line 9: More like a personal preference, but I would avoid conditional overriding here, like so…

```js
const useStubs = process.env.USE_STUBS === "true";
const circuits = [
  "deposit",
  "withdraw",
  "single_transfer",
  "double_transfer",
].map((circuit) => (useStubs ? `${circuit}_stub` : circuit));
```

wallet/src/nightfall-browser/event-handlers/block-proposed.js
```diff
@@ -154,37 +160,38 @@ async function blockProposedEventHandler(data, zkpPrivateKeys, nullifierKeys) {
     }),
   );
 
+  let height = 1;
```
pawelgrzybek, line 163: We reuse this multiplee times across the codebase, maybe it is a good candidate to pt it to common-files at this point?

PR #955, opened 2022-09-16
use HardHat as a test runner

siosw:
# This PR...
* adds hardhat as a test runner while leaving the rest of the current testing infrastructure untouched. 
* prepares migration from web3.js -> ethers  and truffle -> hardhat over time

## Implementation
Currently hardhat@latest doesn't natively support running ESM tests. This PR is using hardhat@esm which is not stable yet

## Usage
tests can be run with `npx hardhat test-esm <PATH> --no-compile`

## Possible future direction
* replace the deployer container with hardhat tasks, this would simplify the docker setup / hopefully improve test speed / remove truffle dependency
* start writing unit tests with ethers / hardhat


package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Optimistic Nightfall",
   "main": "./nightfall-client/src/index.mjs",
   "scripts": {
+    "preinstall": "npx npm-force-resolutions",
```
pawelgrzybek, line 7: Did you try to update the top level `colors` package before doing this?

siosw: you're right, this shouldn't be necessary.
I settled for pinning the dependency version using overrides.
Unfortunately, the GH dependency check still fails, even if a non vulnerable version is pinned.

fvictorio:
Hi, we just published a version of Hardhat that supports esm under the `esm` npm tag. You can try it with `npm i hardhat@esm`. We are not quite ready to publish it under `latest` yet, but we want people to start trying it out so we can understand better what's missing and if it works fine. Please let me know if you find any issues!

package.json
```diff
@@ -4,25 +4,24 @@
   "description": "Optimistic Nightfall",
   "main": "./nightfall-client/src/index.mjs",
   "scripts": {
-    "ping-pong": "mocha --timeout 0 --bail --exit test/ping-pong/ping-pong.test.mjs",
     "lint": "eslint . --ext js,mjs,jsx,ts,tsx && find-unused-exports",
     "prepare": "husky install",
     "doc:build:sdk": "jsdoc -c jsdoc.json cli/lib/nf3.mjs",
     "build-adversary": "node test/adversary/transpile-adversary.mjs",
-    "nightfall-down": "docker-compose -f docker/docker-compose.yml -f docker/docker-compose.dev.yml -f docker/docker-compose.ganache.yml -f docker/docker-compose.apps.yml -p 'nightfall_3' down -v",
```
israelboudoux, line 12: Why did you remove this script? It is being used in the pipeline.

israelboudoux (changes requested):
Just a minor thing to verify.

package.json
siosw: good catch! This should not have been removed 😅

signorecello (approved):
Happy to merge this! Thanks!

PR #958, opened 2022-09-16
KYC backend functionality

Westlad:
This PR implements KYC controls for Nightfall. It is the first of two linked PRs.  This one implements the backend code.  The next PR will make the necessary changes to the `administrator` container to enable on-the-fly creation and removal of whitelist managers.  With this PR, they can only be created at deployment.

This PR introduces a new actor, the Whitelist Manager (WM). Each WM is assigned a group ID, which can be any number > 0. They are then able to manage users with the same group ID, creating them and removing them.  Note that from the user's point of view, the group ID has no effect; they are either whitelisted and able to transact, or they aren't.  It's simply an administrative convenience so that a WM controls a specific group of users.

A user who is not whitelisted is able neither to deposit funds to Layer 2 nor to withdraw them back to Layer 1 (Ethereum).  They may still move funds within Layer 2.

All the endpoints for a WM are implemented in `client` and are accessible via the `Nf3`, 'user' interface. It will be beneficial to implement similar in Wallet but this PR does not do that.

The functionality can be switched off by a call to the `setWhitelisting` function in `KYC.sol`.  This needs to be from the multisig contract.


israelboudoux (changes requested):
Hey @Westlad, please, add some unit tests.

Thanks!

signorecello:
The integration tests are great but I also agree unit tests should be provided. I think a script can be added to the client `package.json` and run with `docker-compose run`

Westlad:
> Hey, can the nf3 changes be reflected in the SDK repo instead, and imported here as an NPM package? I know the new SDK is a new codebase but @imagobea, @velkoskis, @luizoamorim or @daveroga can probably make this change.

> Otherwise we'd be making the change in both places

Very happy for them to be added and that makes sense to me but I'm not going to do it as part of this PR because the new SDK package isn't generally available yet and so that gives me a testing problem (I won't know if I've done it correctly).

nightfall-deployer/contracts/KYC.sol
```diff
@@ -0,0 +1,49 @@
...
+
+    function addUserToWhitelist(address _user) external {
+        // if a non-manager calls this, they will just assign someone to the zero group, which is the null value and has no effect
+        users[_user] = managers[msg.sender];
+    }
+
+    function removeUserFromWhitelist(address _user) external {
+        require(users[_user] != 0, 'This user is not whitelisted, so cannot be delisted');
+        require (managers[msg.sender] == users[_user], 'You are not the manager of this group' );
+        delete users[_user];
```
imagobea, line 29: From my little knowledge of Solidity: I see your comment but I wonder: wouldn't it be worth to check if the caller is manager? Or to have a modifier

nightfall-deployer/contracts/KYC.sol
```diff
@@ -0,0 +1,49 @@
...
+        managers[_manager] = _groupId;
+    }
+
+    function removeWhitelistManager(address _manager) external onlyOwner {
+        delete managers[_manager];
+    }
+
+    function setWhitelisting(bool _whitelisting) external onlyOwner {
+        whitelisting = _whitelisting;
+    }
```
imagobea, line 42:
```suggestion
    function enableWhitelisting(bool _whitelisting) external onlyOwner {
        whitelisting = _whitelisting;
    }
```

nightfall-deployer/contracts/KYC.sol
```diff
@@ -0,0 +1,49 @@
...
+
+    function isWhitelisted(address _user) public returns (bool) {
+        if (whitelisting == false ) return true; // whitelisting is turned off
+        if (users[_user] != 0) return true;
+        return false;
+    }
```
imagobea, line 48: Can/will managers be users at the sametime and therefore belong to the whitelist?

Do we need something like "isManager"?

nightfall-client/src/routes/kyc.mjs
```diff
@@ -0,0 +1,54 @@
...
+  try {
+    const response = await removeUserFromWhitelist(address);
+    res.json(response);
+  } catch (err) {
+    logger.error(err);
+    next(err);
+  }
+});
+
+export default router;
```
imagobea, line 54: So these routes are for managing users, will we have something similar to manage "managers"?

nightfall-deployer/migrations/2_deploy_upgradeable.js
```diff
@@ -71,10 +72,50 @@ module.exports = async function (deployer) {
...
+  console.log('****');
+  console.log('****');
+  console.log('****');
+  console.log('****');
+  console.log('****');
+  console.log('****');
+  console.log('****');
+  console.log('****');
+  console.log('****');
+  console.log('****');
```
imagobea, line 119: :eyes: I'm not very familiar with the deployer but I wonder if we need all that or we could do with some regular logging

imagobea (approved):
Hello @Westlad 

Sorry for the late comments, I wanted to plan stories for enabling this on the SDK and some questions came to mind

Many thanks!!

nightfall-deployer/contracts/KYC.sol
Westlad: Hi @imagobea, I don't think these checks are needed.  For the `addUserToWhistlist` function, if the caller is not  a manager then the mapping `managers` will return a value of zero (the 'undefined' value for an integer type).  This will then add the user to the 'zero' group, which is the same as not being in a group.  This means that if a non-manager calls this function, it has no effect.
For the removeUserFromWhitelist function, if the caller is not a manager then the second `require` will revert because the managers mapping will return a value of zero but the users mapping cannot be zero because of the first `require`.

nightfall-deployer/contracts/KYC.sol
Westlad: I think this is not required, as described above.

nightfall-deployer/contracts/KYC.sol
Westlad: Users can be managers, but there is no harm in having an `isManager` call.  I can add that.

nightfall-client/src/routes/kyc.mjs
Westlad: These will be added to the Administrator container, because they need to be called from the multisig.  So, they won't be in `Nf3`.

nightfall-deployer/migrations/2_deploy_upgradeable.js
Westlad: Good point, I'll remove it.

signorecello (approved):

test/unit/kyc.unit.test.mjs
```diff
@@ -0,0 +1,49 @@
...
+  });
+  it('Turning on whitelisting and creating a whitelisted user by a whitelist manager should succeed', async function () {
+    const MANAGER_GROUP = 42;
+    await ShieldInstance.enableWhitelisting(true);
+    const whitelisting = await ShieldInstance.whitelisting();
+    expect(whitelisting).to.be.a('boolean');
+    expect(whitelisting).to.equal(true);
+    const owners = await ethers.getSigners();
+    await ShieldInstance.createWhitelistManager(MANAGER_GROUP, owners[0].address);
+    expect((await ShieldInstance.isWhitelistManager(owners[0].address)).toNumber()).to.equal(
```
israelboudoux, line 43: Amazing to see these unit tests, but I'm intrigued, how do we know who is the _sender_?

israelboudoux (changes requested):
Some quick changes to double check.

docker/docker-compose.yml
```diff
@@ -68,6 +68,7 @@ services:
       BLOCKCHAIN_PORT: 8546
       ZOKRATES_WORKER_HOST: worker
       USE_STUBS: 'false'
+      WHITELISTING: ${WHITELISTING}
```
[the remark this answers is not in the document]
Westlad: Hi @israelboudoux, no it's an environment variable and thus a string.  It's set to 'enable' if you want whitelisting.

.github/workflows/check-PRs.yml
```diff
@@ -43,6 +43,21 @@ jobs:
           npm ci
           npm run test
 
+  unit-tests:
```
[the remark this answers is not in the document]
Westlad: No,  'test-general-stuff' doesn't make it clear that it runs unit tests.  That's why I set up a specific unit test folder (if it had been clear then I would have known to put this test there). I suggest therefore that we move 'general-tests' into the unit test folder if they are indeed unit tests, but that's not in scope for this PR.

config/default.js
```diff
@@ -49,6 +49,12 @@ module.exports = {
           '0x4789FD18D5d71982045d85d5218493fD69F55AC4',
         ],
   },
+  WHITELIST_MANAGERS: [
```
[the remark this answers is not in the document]
Westlad: Oh yes, good spot.  I'll add a README.

hardhat.config.ts
```diff
@@ -1,9 +1,18 @@
 import { HardhatUserConfig } from 'hardhat/config';
 import '@nomiclabs/hardhat-truffle5';
 import '@nomicfoundation/hardhat-toolbox';
+import 'hardhat-contract-sizer';
 
 const config: HardhatUserConfig = {
-  solidity: '0.8.9',
+  solidity: {
+    version: '0.8.2',
```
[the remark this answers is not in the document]
Westlad: Truffle actually uses 0.8.3 so it's probably best to use that for consistency.  I'll make that change.  0.8.9 is just the hardhat default.

nightfall-client/src/routes/kyc.mjs
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+import express from 'express';
+import logger from 'common-files/utils/logger.mjs';
+import { addUserToWhitelist, removeUserFromWhitelist, isWhitelisted } from '../services/kyc.mjs';
+
+const router = express.Router();
+
+router.get('/check', async (req, res, next) => {
+  logger.debug('kyc/check endpoint received GET');
```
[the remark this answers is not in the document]
Westlad: That's good.  I'll change it.

test/unit/kyc.unit.test.mjs
Westlad: Hardhat defaults to owners[0] so that will be the sending account unless you change it with a specific hardhat command in the test.

.github/workflows/check-PRs.yml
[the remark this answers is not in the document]
israelboudoux: Sure, it makes sense!

PR #979, opened 2022-09-21
Modifying Nullifier Struct in the circuits

### LINKED ISSUE #978: Removing rootKey from the nullifiers struct in circuits
Right now, the nullifier structure for the circuits is the following: 
`struct Nullifiers<N> {
	CommitmentPreimage<N> oldCommitments;
    field[N] rootKey;
    field[N][32] paths;
	field[N] orders;
}`

Each nullifier contains a separate root key. Then, for each nullifier verification we are calculating the zkpPublicKey and nullifiersKey from that one. Calculating the keys have a cost of around 4.000 constraints, so if the circuits have four nullifiers this means a total extra cost of 16.000 constraints.

As it is today, the client do not support spending nullifiers from different root keys, and this logic doesn't seem that will be changed in the next months. Therefore, I suggest removing the rootKey from the nullifier structure so that the keys calculation is only performed once. This change will have the following impact:
- Transfer circuit will go from 73K constraints to 62K constraints 
- Withdraw circuit will go from 55.5K constraints to 43.5K constraints

RogerTaule:
This PR modifies the nullifier struct in the circuits so that rootKey is only sent once and therefore the nullifierKey and zkpPublicKey are calculated only once. With this change, the transfer and withdrawal circuits have a constraint reduction of around 12K, which is roughly a 20% reduction

Yes, it closes #978 

No
## Any other comments? 
No

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -31,7 +31,7 @@ module.exports = {
   TXHASH_TREE_HASH_TYPE: 'keccak256',
   STATE_GENESIS_BLOCK: process.env.STATE_GENESIS_BLOCK,
   CIRCUITS_HOME: process.env.CIRCUITS_HOME || '/app/circuits/',
-  ALWAYS_DO_TRUSTED_SETUP: process.env.ALWAYS_DO_TRUSTED_SETUP || true,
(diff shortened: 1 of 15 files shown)


signorecello (approved):
Looks great man, thanks!

IlyasRidhuan (approved):

PR #896, opened 2022-08-26
Implement PoS for proposers

daveroga:
Feat #726
Fixes #724

Implement proposer selection and staking based on the PR #506 that was developed in a feature branch following the proposal #283.

## Features
- Add operations in the contracts for staking ETH. In the future decide if it's ETH or MATIC.
- Add/Update endpoints for this proposers staking in optimist.
- Add/Update operations to SDK for this proposers staking.
- This staking will be the base of the staking power when running spans of the weighted round robin algorithm for proposers selection in next task
- Implement weighted round robin algorithm based on the proposer staking for proposer selection

## Actions
- registerProposer now pass the amount to stake instead of bond. 
- updateProposer could update the URL and increment the stake .
- No REGISTRATION_BOND now but a MINIMUM_STAKE (100 wei for testing now).
- COOLING_OFF_PERIOD moved to CHALLENGE_PERIOD (1 week) for clarity.
- Changed messages from require statements in contracts to best practices used by Openzeppelin. <contract name>: <message>. So you know the contract where the message has been thrown. Useful now when debugging.
`withdrawBond` moved to `withdrawStake` and used in the same way but with the stake of the proposer instead of the bond, so the proposer can do the withdraw of their stake. Previously to that, the proposer should call `deregisterProposer` to stop being proposer and wait this CHALLENGE_PERIOD to be allowed to do the withdraw.
- `TimeLockedBond` moved to `TimeLockedStake` and added `challengeLocked` to store the amount locked by blocks proposed still in CHALLENGE_PERIOD and not claimed yet.
```
    struct TimeLockedStake {
        uint256 amount; // The amount held
        uint256 challengeLocked; // The amount locked by block proposed still in CHALLENGE_PERIOD
        uint256 time; // The time the funds were locked from
    }
```
- `changeCurrentProposer` will be based on this new proposer selection.
- Update test `e2e/protocol/proposer.test.mjs` for testing PoS features. `npm run test-e2e-protocol`.

Regarding issue of testing https://github.com/EYBlockchain/nightfall_3_private/issues/205 for testing the proposer, they were missing tests for:
- [x] Block Propose
- [x] Change proposer
- [x] Request block payment
- [x] Restrictions to add more than boot proposer

## Discuss
- Define first approach with a set of parameter numbers for testnet (minimum stake, proposer count, number of springs in a span, ...) that we should analyze the suitable numbers for testnet/mainnet.

Issue #283, opened 2021-11-12
Spike: integrating Polygon's POS into Nightfall for block proposers


ChaitanyaKonda:
Research the effort and method required to integrate POS into Nightfall for block proposers

daveroga:
This is the staking contract in which Polygon base their SDK: 
https://github.com/0xPolygon/staking-contracts/blob/main/contracts/staking.sol
I don't know for sure if we want to implement a PoS protocol that could be based on that contract and implement the operations for the proposers to stake/unstake and be the current proposer based on some rules from a list of staking proposers and a kind of epoch.
Or on the other hand we want to base the proposers list and current proposer based on the MATIC staked in the Polygon network discussed with @Westlad. Basically being a Polygon Staker should entitle you to be a Nightfall Proposer, and that will be done mainly through an integration at smart-contract level.

Another question is what we do with Challengers and Liquidity providers - do we leave these open for anyone or do we constrain them to be Polygon Stakers too? Something to be debated.

daveroga:
Hari found some other repo of Polygon/MATIC network: https://github.com/maticnetwork/contracts. The one I found is only for Polygon SDK as I see.
Here is also some documentation about staking:
https://github.com/maticnetwork/matic-docs/blob/master/docs/home/architecture/contracts/staking.md

ChaitanyaKonda:
- Contrast the round robin proposer selection approach with Polygon’s validator selection
- If it is a good solution, steps to replace the round robin proposer selection approach with Polygon’s validator selection 
- Staking using MATIC 
- contrast Polygon's stake, bond, slashing mechanism and incentive mechanism against Nightfall's
- If these are a good solution, steps to replace Nightfall's with Polygon's

As for consensus mechanism, we’re relying on Ethereum’s consensus PoW and soon PoS currently because our transactions are submitted to Ethereum mainnet not Polygon sidechain. As a result, our consensus will definitely have to be Ethereum’s

dwebchapey:
[NF_POLY_POS.pptx](https://github.com/EYBlockchain/nightfall_3/files/7854933/NF_POLY_POS.pptx)

Sharing deck following my study on Polygon POS integration and useful links for reference

ChaitanyaKonda:
How does Polygon's POS deal with idle proposers?

daveroga:
Shared hackmd document  for the proposal of the different parts of the Nightfall PoS based on the Polygon PoS information:
[...]

PR #980, opened 2022-09-21
docs: what is a proposer

### LINKED ISSUE #944: Define Proposer requirements
We need to understand exactly what does this SDK has to do, what functions must it take care of, etc. These can be understood by inspecting the old SDK and understanding what new functions need to be written as part of the multiple proposers EPIC

imagobea:
Code changes mostly remove unused code from old sdk and optimist/proposer API. Also some name updates in optimist/proposer API.

### What is a proposer

**Introduction**

Nightfall is currently an Optimistic rollup, i.e. an L2 protocol designed to extend the throughput of Ethereum's base layer. They reduce computation on the main Ethereum chain by processing transactions off-chain. Optimistic rollups can offer up to 10-100x improvements in scalability. Optimistic rollups also write transactions to Ethereum as calldata, reducing gas costs for users. [1]

Nightfall Proposer is responsible for transaction aggregation and publishing blocks on Ethereum.

**Nightfall current implementation vs PoS**

We want to transition from a centralised proposer to a multi-proposer solution.

Current implementation for multi-proposer relies on "round-robin selection". New proposers are stored in a linked list (we store the address of the last proposer added too). From this list, someone will act as `currentProposer`, being the only proposer able to produce new blocks. After 4 blocks in L1, proposer should be rotated by calling [changeCurrentProposer](https://github.com/EYBlockchain/nightfall_3/blob/master/nightfall-deployer/contracts/Proposers.sol#L26). How this works is not fully defined but the idea would be for each proposer to monitor `currentProposer` and proactively call the function whenever they go next. Cryptoeconomic incentives are in place to ensure that proposers act honestly. They must deposit a bond upon registration and they will also provide some stake each time they propose a block.

With @daveroga 's [PoS PR](https://github.com/EYBlockchain/nightfall_3/pull/896) we will transition into a weighted round-robin, where proposers are selected randomly but can improve their odds by having larger stakes. In this implementation, `changeCurrentProposer` defines a `Proposer set`. This set are the selected proposers with calculated weights that will be proposers during the sprints of next span (see image). So calling `changeCurrentProposer` will update `currentProposer` to the next one in the set, or will produce a new Proposer set by the end of the span/last sprint. FYI this implementation also drops bonds in favour of stake.

I highly recommend to [visit Hackmd](https://hackmd.io/m_85iGlMTzGkc4hZSElTKQ?view) to learn how Producer set is generated. You can also deep-dive into the original conversations and see some code on this repo, issue #283 .

image

**As a proposer...**

- **I can register to join the proposers network**. Requirements: a) Optimist running b) need to provide url c) with PoS, also stake - what is the minimum stake? At the very least it has to match `SLOT` d) is KYC needed? Etc.

- **I need to open a web socket to listen to Optimist events**. These are: a) "block" - I am `currentProposer` and I need to propose the next block, b) "blockProposed" - new block added to the chain, c) "rollback" - challenge was successful and block(s) need to be removed. The last 2 are to keep Optimist in sync with the network.

- **I can update the URL I set upon registration**. With PoS, the ability to add staking has been "tight" to this same function `updateProposer` in Proposer.sol.

- **I can de-register anytime** (even if I am `currentProposer`). After some time, I can also recover my bond/stake. 

- **I should be "motivated" to monitor/call `changeCurrentProposer`**, so that proposers rotate once the defined period has elapsed

- **I can receive off-chain transfer and withdrawal transactions for a `fee`**, which will be added to Optimist mempool. FYI Optimist already prioritises most profitable transactions when assembling a block (`getMostProfitableTransactions`).

- **I can set/update a minimum value for `fee`**, so that I can refuse transfer and withdrawal transactions that I don't consider profitable. Perhaps there is also a MAX_FEE?

- **I can earn rewards for each block-proposed which is legit**. After some time, I can also claim these rewards.
 
- **I will lose stake for every successful challenge** (that is, Challenger identifies malicious behaviour)

**As a user...**

- I can see who is current proposer
- I can get a list of proposers and their fees
- I can select to whom I send my off-chain transaction
- I can monitor both on-chain and off-chain transactions (Block Explorer will only display mempool for current proposer, but we might consider providing ability of displaying other mempools)
- I should be able to cancel my transaction if this is being ignored

Closes #944

N/a

## Any other comments?

**References**
[1] https://ethereum.org/en/developers/docs/scaling/optimistic-rollups/#transaction-execution-and-aggregation


signorecello (approved):
Amazing work. This is a great guideline for what comes next in terms of an informed proposer architecture. Thank you @imagobea

imagobea:
More questions:

- What happens to idle proposers

luizoamorim:
## Proposer Brainstorming

### I'm adding some thoughts about proposer fees.

First we need to establish the L2 transaction price in terms of bytes.

We also need to establish a minimum stake to be a proposer.

Then we can start the fee calculation from this minimum stake to a maximum fee. How?

- I.e: a minimum stake to be a proposer is 10 MATIC. For this minimum the fee is 1 * (transaction bytes * transaction price by bytes)
- Then if the proposer increase the stake to 20 MATIC, now the fee go to 1.1 * (transaction bytes * transaction price by bytes)

Now we can think in a way of the user choose with what proposer in the proposer set it want to do a transaction.

Let’s say that the proposer set is

- Alice: 50 Matic
- Bob: 20 Matic
- Clara: 20 Matic

And the proposer set is [ Alice, Alice, Alice, Bob, Clara ] - I think here we can order by the big stake to the less.

So the user can choose one of the prices in the proposer set to pay:

1.1 * (transaction bytes * transaction price by bytes) to Bob or Clara (The user will wait more to have the transaction added to the block)

1.4 * (transaction bytes * transaction price by bytes) to Alice (The user will have the transaction added first)

druiz0992 (approved):

PR #981, opened 2022-09-21
Edgefall integration

### LINKED ISSUE #927: Integrating Edgefall into Nightfall repo

signorecello:
This implements a way to deploy on Polygon Edge networks. It assumes an existing Polygon Edge network running on your localhost, or otherwise in a set `BLOCKCHAIN_WS_HOST:BLOCKCHAIN_PORT/BLOCKCHAIN_PATH`

Yes, it closes #927 

No

## Any other comments?
No


cli/lib/nf3.mjs
```diff
@@ -222,14 +224,20 @@ class Nf3 {
   async estimateGas(contractAddress, unsignedTransaction) {
     let gasLimit;
     try {
-      // eslint-disable-next-line no-await-in-loop
-      gasLimit = await this.web3.eth.estimateGas({
-        from: this.ethereumAddress,
-        to: contractAddress,
-        data: unsignedTransaction,
+      const res = await axios.post(this.web3WsUrl, {
```
druiz0992, line 227: This looks like a weird construct. Can you add some comments in the code so that this code is not modified later

druiz0992, line 227: Can you add some comments as to why you’ve done this? I want to prevent somebody reverting this in the future

nightfall-client/src/services/state-sync.mjs
```diff
@@ -36,6 +36,20 @@ export const syncState = async (
   console.log('From block', fromBlock);
   const stateContractInstance = await waitForContract(STATE_CONTRACT_NAME); // Rollback, BlockProposed
 
+  // console.log(
```
druiz0992, line 39: Remove this code

nightfall-deployer/entrypoint.sh
```diff
@@ -2,10 +2,11 @@
 set -o errexit
 set -o pipefail
 
-if [ -z "${ETH_PRIVATE_KEY}" ]; then
-  # wait until there's a blockchain client up
-  while ! nc -z ${BLOCKCHAIN_WS_HOST} ${BLOCKCHAIN_PORT}; do sleep 3; done
-fi
+echo "${ETH_NETWORK}"
```
druiz0992, line 5: Why did you comment this code? If correct, you should remove commented code

nightfall-deployer/src/circuit-setup.mjs
```diff
@@ -124,21 +127,27 @@ async function setupCircuits() {
...
 
-      // when deploying on infura - do serial tx execution to avoid nonce issue
       // when using a private key, we shouldn't assume an unlocked account and we sign the transaction directly
       if (config.ETH_PRIVATE_KEY) {
-        await Web3.submitRawTransaction(await tx.encodeABI(), keyRegistryAddress);
-      } else await tx.send();
+        const tx = {
+          from: process.env.FROM_ADDRESS,
+          to: keyRegistry.options.address,
+          data: call.encodeABI(),
```
druiz0992, line 140: Explain why you do this in the code

package.json
```diff
@@ -12,9 +12,9 @@
     "nightfall-down": "docker-compose -f docker/docker-compose.yml -f docker/docker-compose.dev.yml -f docker/docker-compose.ganache.yml -f docker/docker-compose.apps.yml -p 'nightfall_3' down -v",
     "test": "LOG_LEVEL=error mocha --timeout 0 --bail --exit test/e2e/protocol/*.test.mjs test/e2e/tokens/*.test.mjs",
     "test-e2e-protocol": "LOG_LEVEL=error mocha --timeout 0 --bail --exit test/e2e/protocol/*.test.mjs ",
-    "test-gas": "LOG_LEVEL=error mocha --timeout 0 --bail --exit test/e2e/gas.test.mjs ",
-    "test-circuits": "LOG_LEVEL=debug mocha --timeout 0 --bail --exit test/e2e/circuits.test.mjs ",
-    "test-e2e-tokens": "LOG_LEVEL=error mocha --timeout 0 --bail --exit test/e2e/tokens/*.test.mjs ",
+    "test-gas": "mocha --timeout 0 --bail --exit test/e2e/gas.test.mjs ",
```
druiz0992, line 15: Why this?

test/e2e/tokens/erc20.test.mjs
```diff
@@ -400,7 +400,7 @@ describe('ERC20 tests', () => {
       }
     });
 
-    it('should restrict withdrawals', async () => {
+    it('should restrict withdrawals', async function () {
```
druiz0992, line 403: Why this change?

druiz0992 (changes requested):
Some clarifications requested

nightfall-deployer/src/circuit-setup.mjs
signorecello: There's no account management on edge, you can't use send(). You need to encodeABI() and then sign it with a supplied private key

cli/lib/nf3.mjs
signorecello: I can, but I don't see why it's weird. `estimateGas()` doesn't work properly on Edge, but it works fine if you call the RPC node directly. This is being addressed, but in the meantime I'm using this workaround

test/e2e/tokens/erc20.test.mjs
signorecello: [Because `this` is bound to the wrong scope when you use arrow functions](https://stackoverflow.com/questions/51491553/mocha-using-this-skip-to-skip-tests-dynamically-doesnt-work)

nightfall-deployer/src/contract-setup.mjs
```diff
@@ -4,96 +4,62 @@ address of the contract that holds global state (State.sol)
...
+    from: process.env.FROM_ADDRESS,
+    // value: fee,
+    gas: config.WEB3_OPTIONS.gas,
+    gasPrice: config.WEB3_OPTIONS.gasPrice,
+  };
 
-  logger.debug(
-    `transfering ownership of contracts to simpleMultiSigAddress ${simpleMultiSigAddress}`,
-  );
+  const proposers = await waitForContract('Proposers');
```
daveroga, line 24: I would put the names of the variables with `Instance` as the first one above `stateInstance` to clarify they are instance of the contract.
So `proposersInstance`, `shieldInstance`, `challengesInstance`.

nightfall-deployer/src/contract-setup.mjs
```diff
@@ -4,96 +4,62 @@ address of the contract that holds global state (State.sol)
...
   const stateInstance = await waitForContract('State');
   logger.debug(`address of State contract is ${stateInstance.options.address}`);
 
-  const simpleMultiSigAddress = (await waitForContract('SimpleMultiSig')).options.address;
-  const shieldContractInstance = await waitForContract('Shield');
-  const stateContractInstance = await waitForContract('State');
-  const proposerContractInstance = await waitForContract('Proposers');
-  const challengesContractInstance = await waitForContract('Challenges');
+  const tx = {
+    from: process.env.FROM_ADDRESS,
```
daveroga, line 18: Why are we mixing `process.env` variables with `config` ones here? It would be better to put all in `config` variables that check this `process.env.FROM_ADDRESS`.

nightfall-deployer/src/contract-setup.mjs
```diff
@@ -4,96 +4,62 @@ address of the contract that holds global state (State.sol)
...
-    return Web3.submitRawTransaction(
-      challengesContractInstance.methods.transferOwnership(simpleMultiSigAddress).encodeABI(),
-      challengesContractInstance.options.address,
+  const contracts = {
+    proposers,
+    shield,
+    challenges,
+  };
+
+  for await (const contractName of ['proposers', 'shield', 'challenges']) {
```
daveroga, line 36: Here better to get all the contract names by the keys of the object to avoid errors.
```
for await (const contractName of Object.keys(contracts)) {
```

nightfall-deployer/src/contract-setup.mjs
```diff
@@ -4,96 +4,62 @@ address of the contract that holds global state (State.sol)
...
-      logger.warn(
-        'Writing contract addresses to the State contract failed. This is probably because they are aready set. Did you already run deployer?',
+    const transferOwnership =
+      contracts[contractName].methods.transferOwnership(simpleMultiSigAddress);
+
+    if (!config.ETH_PRIVATE_KEY) {
+      setStateContract.send();
+      transferOwnership.send();
+    } else {
+      tx.data = setStateContract.encodeABI();
```
daveroga, line 48: What if instead of this code
```
      tx.data = setStateContract.encodeABI();
      let signed = await web3.eth.accounts.signTransaction(
        { ...tx, to: contracts[contractName].options.address },
        config.ETH_PRIVATE_KEY,
      );
      await web3.eth.sendSignedTransaction(signed.rawTransaction);
```
we use the previous function that do the same code? 

```
      await Web3.submitRawTransaction(
        setStateContract.encodeABI(),
        contracts[contractName].options.address,
      );
```

daveroga (changes requested):
Some clarifications and changes requested

nightfall-deployer/src/contract-setup.mjs
signorecello: This `FROM_ADDRESS` is needed as environment variable, actually we do the same for `geth` deployments. We could do something like this in `config`:

`
FROM_ADDRESS: process.env.FROM_ADDRESS || 0x.....
`

And then use `config.FROM_ADDRESS` here. But I don't see the benefit, and that's why I just left it as it is.

I agree our codebase is very dependent on environment variables, we should think of a better way

nightfall-deployer/src/contract-setup.mjs
signorecello: indeed this seems to do the same thing, I don't know how I arrived on the top code, I'll try to refactor and see if there was any side effect I'm not remembering now

nightfall-deployer/src/contract-setup.mjs
```diff
@@ -14,26 +14,19 @@ async function setupContracts() {
...
-    proposers,
-    shield,
-    challenges,
+    proposersInstance,
+    shieldInstance,
+    challengesInstance,
   };
 
-  for await (const contractName of ['proposers', 'shield', 'challenges']) {
+  for await (const contractName of Object.keys(contracts)) {
```
daveroga, line 27: I don't know if here we could parallelized with `Promise.all` instead of `for await` like this and also avoid then `Object.keys`
```
  await Promise.all(
    contracts.map(async contract => {
      const setStateContract = contract.methods.setStateContract(stateInstance.options.address);
      const transferOwnership = contract.methods.transferOwnership(simpleMultiSigAddress);
      if (!config.ETH_PRIVATE_KEY) {
        setStateContract.send();
        transferOwnership.send();
      } else {
        await Web3.submitRawTransaction(setStateContract.encodeABI(), contract.options.address);
        await Web3.submitRawTransaction(transferOwnership.encodeABI(), contract.options.address);
      }
    }),
  );
```

daveroga (changes requested):

nightfall-deployer/src/contract-setup.mjs
signorecello: I tried that way previously, and it works. But after some research I found the modern `for await` and I believe it makes the code easier to read than `Promise.all`. It's a styling matter, I can make the change back to `Promise.all` if you prefer

nightfall-deployer/src/contract-setup.mjs
```diff
@@ -4,96 +4,49 @@ address of the contract that holds global state (State.sol)
...
-import logger from 'common-files/utils/logger.mjs';
 import Web3 from 'common-files/utils/web3.mjs';
 import { waitForContract } from 'common-files/utils/contract.mjs';
+import logger from 'common-files/utils/logger.mjs';
 
 async function setupContracts() {
   const stateInstance = await waitForContract('State');
   logger.debug(`address of State contract is ${stateInstance.options.address}`);
 
+  const proposersInstance = await waitForContract('Proposers');
```
pawelgrzybek, line 15:
```js
await Promise.all[waitForContract("Proposers"), waitForContract("Shield"), waitForContract("Challenges")]
];
```

?

pawelgrzybek (changes requested):
Also a little general note. Just a reminder for consistency. In some instances w use if statement with block brackets, in some without. Also, in some instances we destructure things from `config` and we dont in other cases. Can we strive for more consistency pleasE?

nightfall-deployer/src/circuit-setup.mjs
druiz0992: Again, the idea is that you highlight this in the code so that future readers understand. We don't want somebody removing this code in the future, and its good idea to have this explained in comments in the code.

nightfall-deployer/src/contract-setup.mjs
```diff
@@ -4,96 +4,64 @@ address of the contract that holds global state (State.sol)
...
+
   const stateInstance = await waitForContract('State');
-  logger.debug(`address of State contract is ${stateInstance.options.address}`);
+  console.log(`address of State contract is ${stateInstance.options.address}`);
+
+  const tx = {
+    from: process.env.FROM_ADDRESS,
+    // value: fee,
+    gas: config.WEB3_OPTIONS.gas,
+    gasPrice: 20000000000,
```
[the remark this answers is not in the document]
druiz0992: is it a good idea to have `gasPrice` hardcoded?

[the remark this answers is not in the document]
signorecello: no, you're right and I made the change (that's why it's marked outdated)

cli/lib/nf3.mjs
signorecello: hey as suggested I added a comment explaining it:
https://github.com/EYBlockchain/nightfall_3/blob/zepedro/edgefall/cli/lib/nf3.mjs#L225

that's why I resolved this comment. I'll add another comment to the `encodeABI` call, as suggested

nightfall-client/src/services/state-sync.mjs
signorecello: same, this is marked outdated by github, as I removed this code after your suggestion

nightfall-deployer/src/circuit-setup.mjs
signorecello: absolutely, I'll add another commit with a comment

nightfall-deployer/src/contract-setup.mjs
daveroga: What I found is this

"Use for await of if the order in which promises are resolved is important to you. Use Promise. all() if the order isn't important and you need all calls to succeed."

signorecello: Oh I thought `for await` would also parallelize it just the same. And actually after refactoring to your code it looks more concise and understandable. I'll push a new commit with this change, thanks man

signorecello:
> 

Hey Pawel I agree we should thrive for more consistency, I guess we should have a style guide and abide for that

druiz0992 (approved):

daveroga (approved):
LGTM

pawelgrzybek (approved):

PR #982, opened 2022-09-21
Optimizing Root Challenge

### LINKED ISSUE #773: Gas optimization of challengeNewRootCorrect in Challenges.sol
The challenge `challengeNewRootCorrect` is high gas consumption function and fails if we don't configure 30M block gas limit in the node and miner.
We should find a way to optimize the gas consumption of this function to be able to create this challenge with a lower block gas limit (15M).

RogerTaule:
This PR optimizes the gas used for challenging a new root.
To a better understanding of the changes done in this PR we need to take into account the following:
- All challenges have a first part that verifies that the challenger info is valid (so he isn't "cheating") and then the actual challenge. In this case, what the test was doing was inserting all the leaves to calculate the root of the previous block to make sure that the info provided is valid and then do the same with the current block to perform the challenge. 
- Timber merkle trees have the following property: if you have the updated frontier and the leafIndex of the last leaf, you are able to calculate the root. This means that the process can be split in two: adding all the leaves (which updates the frontier) and calculate the root.

In order to split the challenge, a new property called **frontierHash** is added to the block. This is the hash of the frontier after all the transactions has been processed. The idea is that when the challenger sends a frontier to perform a challenge, the challenger will check its validity by calculating the hash and checking if it matches with the one stored in the block.

To challenge the root, we just need the frontier (provided by the challenger and validated with the frontier hash) and the leafIndex (stored in the block itself. Keep in mind that PR #917 changed leafIndex so now means the last updated index). We can calculate the root using the property mentioned above.

It is also possible that the frontier hash was wrong. To challenge it, the challenger would send the previous frontier and all the transactions from the new block. Then, the challenge will first check the validity of the frontier (same as above) and then will perform all the leaves insertions. Finally it will perform a hash and check it with the one stored in the block.

With all this changes, each test is now approximately 25% the cost of the original challenge.

In addition to this logic change, the leaf insertion has been also converted into assembly to get an extra gas reduction. 

Yes, it closes #773 

No

## Any other comments? 
This PR must be merged after #917

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -457,11 +457,11 @@ module.exports = {
   // is running in local machine
   isLocalRun: process.env.LOCAL_PROPOSER === 'true',
   SIGNATURES: {
-    BLOCK: '(uint48,address,bytes32,uint256,bytes32,bytes32)',
+    BLOCK: '(uint48,address,bytes32,uint256,bytes32,bytes32, bytes32)',
     TRANSACTION:
       '(uint112,uint112,uint8,uint8,uint64[4],bytes32,bytes32,bytes32,bytes32[3],bytes32[4],bytes32[2],uint256[4])',
     PROPOSE_BLOCK: [
-      '(uint48,address,bytes32,uint256,bytes32,bytes32)',
+      '(uint48,address,bytes32,uint256,bytes32,bytes32,bytes32)',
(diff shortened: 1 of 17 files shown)


IlyasRidhuan (approved):

daveroga (approved):
LGTM

Issue #990, opened 2022-09-23
Create endpoint in client/optimist for contract-abi


daveroga:
We have face some problems getting the updated `abi` of the contracts. The same we have a `contract-address` operation in the client and optimist service, we will add a `contract-abi` in order to get this abi.

PR #992, opened 2022-09-23
Add contract-abi endpoints to client/optimist

daveroga:
Fix: #990
We have some issues when trying to get the version of the ABI for the current deployment in the Block Explorer on in some tests.
We will add enpoints to get ABI of the contracts in the `nightfall-client` and `nightfall-optimist` the same we have now with `contract-address`.
The endpoint will be `contract-abi`.

Example:
```
curl http://localhost:8080/contract-abi/State 
```

Issue #990


RogerTaule (approved):
LGTM

PR #1002, opened 2022-09-27
fix: updated filter conditions in client balances to return erc721 data

imagobea:

When we store ERC721 tokens we set the value to `0` (since we care about `tokenId`). Filtering commitments above 0 value is preventing `/pending-deposit` and `/pending-spent` from returning info re ERC721.

n/a

n/a

## Any other comments?

n/a


RogerTaule (approved):
LGTM

daveroga (approved):
LGTM

PR #1003, opened 2022-09-27
Using non-JSON logging only for local envs

israelboudoux:
Basically, when in local environments the logger will issue logs in a non-JSON format and, when in other envs, will use JSON format.

Nope

Nope

## Any other comments?
Nope


luizoamorim (approved):

daveroga (approved):
LGTM

PR #1004, opened 2022-09-27
fix setup

Westlad:
Fixes a bug in the setup script which prevents it running on an M1 Mac.

No.

No

## Any other comments?
This is a simple typo, that doesn't manifest itself unless the NO_CACHE flag is set.


Westlad:
Seems to have made many unintentional updates.

PR #1006, opened 2022-09-27
fix: fixing lower and upper case wrong comparison with ercAddress

RogerTaule:
## What does this implement/fix? 
Ethereum addresses uses capital and lower-case letters, and this was causing some issues in the SDK because commitments weren't fully consistent. All ercAddress are now converted to lower case

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -81,7 +81,7 @@ module.exports = {
   CURVE: process.env.CURVE || 'bn128',
 
   TRANSACTIONS_PER_BLOCK: Number(process.env.TRANSACTIONS_PER_BLOCK) || 2,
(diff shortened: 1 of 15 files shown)


siosw:
Should we consider consistently using the checksum addresses instead of all lower case? Then web3 will throw if the checksum is invalid which might be a nice safeguard.

Ofc there might be totally valid reasons not to do this

daveroga (approved):
LGTM

imagobea (approved):
Thank you @RogerTaule LGTM!

velkoskis (approved):
LGTM. Transfers and withdrawals work good now with fees!

PR #1010, opened 2022-09-27
Log Adjustments

israelboudoux:
- Removes the fields `pid` and `hostname` from JSON logs;
- Logs are now obfuscated only for non-local environments;
- Removing the `JSON.stringfy` so that the logs will be cleaner;
- Added env variable `LOG_HTTP_PAYLOAD_ENABLED` for allowing disabling the automatic logs of HTTP payloads;
- Removing unnecessary log statements from routes, since they are automatically performed when `LOG_LEVEL == 'debug'` & `LOG_HTTP_PAYLOAD_ENABLED == true`;
- Removing `logger.error(err)` from the `try...catch` in the routes since they are handled in the `errorHandler` middleware;
- Add more unit tests regarding to obfuscation.

No

## Any other comments?
No


Westlad:
Would it be better to move the unit test runner to the top level of the repo, with the test in `test/unit`?  I don't really mind if we locate them at top level or in each folder but it's best to be consistent.

israelboudoux:
> Would it be better to move the unit test runner to the top level of the repo, with the test in `test/unit`? I don't really mind if we locate them at top level or in each folder but it's best to be consistent.

Hello @Westlad, the thing is that we have different projects under the same repo and these unit tests should stay under the umbrella of the project where its units reside.

common-files/test/utils.test.mjs
```diff
@@ -119,8 +120,24 @@ describe('Utils tests', function () {
     assert.equal(result, '');
   });
 
-  it("Shouldn't obfuscate for the Dev environment", () => {
-    process.env.NODE_ENV = 'developer';
+  it("Shouldn't obfuscate for the local environment", () => {
+    process.env.ENVIRONMENT = 'local';
```
signorecello, line 124: I don't think there's an environment called `local`, they should match the ones you find in `config.ENVIRONMENTS`. If needed, add one, but we should be consistent

common-files/utils/logger.mjs
```diff
@@ -11,7 +11,7 @@ const LOGGER_TIME_STRING = 'yyyy-mm-dd HH:MM:ss.l';
 const getInstance = () => {
   const pinoOptions = {
     level: config.LOG_LEVEL || 'info',
-    messageKey: 'message',
+    base: undefined,
```
signorecello, line 14: Wouldn't it be `undefined` by default if we don't set it?

nightfall-optimist/src/routes/proposer.mjs
```diff
@@ -397,14 +338,8 @@ router.post('/encode', async (req, res, next) => {
       )
       .encodeABI();
 
-    logger.debug({
```
signorecello, line 400: I see a lot of these debug messages (raw transaction) being removed, but they're quite useful sometimes. Are they being logged somewhere else? If not, I'd leave them here

signorecello (changes requested):
Looks great! But specially the environment should be consistent with config. Can you have a look at it please?

common-files/test/utils.test.mjs
israelboudoux: Hey @signorecello, the thing with this is that for being considered local, it should be different from 'aws', then I could use whatever value since it isn't equal to 'aws'.

common-files/utils/logger.mjs
israelboudoux: Nope, it has some values and are printing the `host` and `pid` which are unnecessary for us.

nightfall-optimist/src/routes/proposer.mjs
israelboudoux: These logs in the routes are being carried out automatically when we set the `LOG_LEVEL=debug`. We log the payload and the response ;)

imagobea (approved):
Hey @israelboudoux , thank you for this. LGTM, I just saw a few typos and commented on the env var.

I must say I had a bit of a hard time following all the obfuscation functionality. The test suite is helpful as it serves as documentation. Because of that, perhaps tests could be organised around different "describes". E.g.

```
describe("Utils", () => {
  describe("Obfuscation", () => {
    describre("Obfuscate objects");
    describre("Obfuscate query string");
  })
})
```

A suggestion for future tests!! Cheers :100:

common-files/test/utils.test.mjs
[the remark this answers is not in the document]
israelboudoux: This test was removed in this [PR](https://github.com/EYBlockchain/nightfall_3/pull/1088) since all sensitive data will be obfuscated whatever the environment.

PR #1013, opened 2022-09-28
proposeBlock prevents to add a deposit that has not previously escrowed its funds using submitTransaction

### LINKED ISSUE #1001: Make sure L2 deposits are only made on-chain
Right now, nothing is preventing users to send a deposit transaction off chain and then this transaction being added to an L2 Block. Therefore, we can end up in a situation in which a user would have a valid L2 commitment but would have not escrowed the funds.

To solve it, we must add a `mapping(uint256 => bool) isEscrowed` which maps a transaction hash to a boolean which tells if the user has escrowed the funds or not. This mapping will be only set to true inside `submitTransaction` function.
Then, when a proposer calls `proposeBlock`, if the transaction is a deposit it will require that the mapping `isEscrowed` is set to true. Therefore, if a deposit transaction is added off chain the function will fail because the mapping will be set to false.

RogerTaule:
Currently, a user could potentially send a deposit off-chain to a proposer and this deposit could be added to a valid block without having escrowed the funds first. To solve this problem, a new mapping called `isEscrowed(uint256 => bool)` is added. This mapping checks if the funds for a transaction hash has been deposited. It is only set to true when submitting a deposit transaction using `submitTransaction`. Then, when a proposer calls `proposeBlock` it will check that all transactions with transaction type 0 (deposits) has been escrowed by checking the mapping.

This PR also contains a major refactor of the assembly code contained in `proposeBlock` to make it more understandable and organized. 

It also adds two new const variables in Config.sol that are the number of slots used in the transaction structure and the block structure respectively. If any of those structures are modified this constants will need to be updated. The reason of adding this constants is to not have to directly modify the assembly code every time that the structure is updated.

This PR also adds a couple of new modifiers inside State.sol
Yes, it closes #1001 

Tested locally - CI tests require adversary
In order to test that a deposit that has not previously escrowed will fail, one can simply comment the line in `submitTransaction` that updates the mapping `isEscrowed` and check that when trying to run the tests locally it fails because there's a revert in the Smart Contract call.
## Any other comments?
No

### DIFF
--- a/nightfall-deployer/contracts/Config.sol
+++ b/nightfall-deployer/contracts/Config.sol
@@ -12,6 +12,8 @@ contract Config is Ownable, Structures {
     uint256 constant COOLING_OFF_PERIOD = 1 weeks;
     bytes32 constant ZERO = bytes32(0);
     uint256 constant TRANSACTIONS_PER_BLOCK = 32;
+    uint256 constant BLOCK_STRUCTURE_SLOTS = 7;
+    uint256 constant TRANSACTION_STRUCTURE_SLOTS = 24;
 
     address bootProposer;
     address bootChallenger;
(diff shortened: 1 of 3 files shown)


daveroga (approved):
LGTM

IlyasRidhuan (approved):

Issue #885, opened 2022-08-19
Test creation of other bad L2 blocks & bad transactions


hnhngn:
Test creation of other bad L2 blocks & bad transactions such as:

- Duplicate Commitment in all transfer types and deposit
- Duplicate Nullifier in all transfer types and withdraw
- Incorrect Proof from deposit, all transfer types and withdraw wrong proof, wrong public inputs (check by making each public inputs incorrect)

PR #1028, opened 2022-09-29
optimist as a challenger instead of adversary for adversary.test.mjs

LijuJoseJJ:
This PR make adversary.test.mjs to make websocket connection to optimist instead of adversary container which run a challenger

https://github.com/EYBlockchain/nightfall_3/issues/885

## Any other comments?
evidence: temporary successful github workflow run created on this branch https://github.com/EYBlockchain/nightfall_3/actions/runs/3150503026/jobs/5123363365

### DIFF
--- a/docker/docker-compose.adversary.yml
+++ b/docker/docker-compose.adversary.yml
@@ -37,6 +37,9 @@ services:
       # right now it doesn't matter what is IS_CHALLENGER flag's value is because
       # optimist container state-sync code start challenges at time of startup
       # by calling startMakingChallenges() function
+      # Also we don't want adversary who create bad block to be challenger as well
+      # this bit of code logic is updated in test/adversary.test.mjs by making challenger
+      # wbsockeet client to connect to optimist container instead of adversary
       IS_CHALLENGER: 'false'
       NONSTOP_QUEUE_AFTER_INVALID_BLOCK: 'false'
       TRANSACTIONS_PER_BLOCK: ${TRANSACTIONS_PER_BLOCK:-2}
--- a/docker/docker-compose.adversary.yml
+++ b/docker/docker-compose.adversary.yml
@@ -34,7 +34,10 @@ services:
       BLOCKCHAIN_PORT: 8546
       HASH_TYPE: poseidon
       LOG_LEVEL: debug
-      IS_CHALLENGER: 'true'
+      # right now it doesn't matter what is IS_CHALLENGER flag's value is because
+      # optimist container state-sync code start challenges at time of startup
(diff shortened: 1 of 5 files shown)


nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -46,15 +46,15 @@ export async function signalRollbackCompleted(data) {
   // check that the websocket exists (it should) and its readyState is OPEN
   // before sending. If not wait until the challenger reconnects
   let tryCount = 0;
-  while (!ws || ws.readyState !== WebSocket.OPEN) {
+  while (ws && ws.readyState !== WebSocket.OPEN) {
```
ChaitanyaKonda, line 49: Can you explain this change please?

docker/docker-compose.yml
```diff
@@ -98,6 +98,9 @@ services:
       BLOCKCHAIN_PORT: 8546
       HASH_TYPE: poseidon
       LOG_LEVEL: debug
+      # right now it doesn't matter what is IS_CHALLENGER flag's value is because
```
ChaitanyaKonda, line 101: Then `IS_CHALLENGER` is not really used. It is not clear if this flag is needed anymore

docker/docker-compose.adversary.yml
```diff
@@ -34,7 +34,10 @@ services:
       BLOCKCHAIN_PORT: 8546
       HASH_TYPE: poseidon
       LOG_LEVEL: debug
-      IS_CHALLENGER: 'true'
+      # right now it doesn't matter what is IS_CHALLENGER flag's value is because
+      # optimist container state-sync code start challenges at time of startup
```
ChaitanyaKonda, line 38: User's who read this might not be aware, which optimist we are referring to. 
Because during ping pong tests we have adversary optimist and normal optimist.

But this docker compose can be used anytime we want adversary too and not just ping pong.
Can we just specify that we shouldn't expect a bad proposer to be a good challenger.
Hence while using adversary, one needs to up a different container for good optimist for challenging.

ChaitanyaKonda (commented):
Posted comments

nightfall-optimist/src/services/block-assembler.mjs
LijuJoseJJ: reverted as discussed

docker/docker-compose.adversary.yml
LijuJoseJJ: explanation added as asked for

ChaitanyaKonda (approved):

PR #1029, opened 2022-09-29
Force trusted setup if circuits have been modified

### LINKED ISSUE #989: Force trusted setup is circuits have been modified
Currently, trusted setup is only done if either keys are not found or it its forced. However, it circuits have been modified and no trusted setup is forced the transaction will fail the verification.  This issues proposes a mechanism to overcome this. It will provide a better usability so that circuits that are modified do always go through a trusted setup.

The mechanism could be something like:
- In worker we have circuits/deposit.zok and we have output/deposit/deposit_out.
- In the output/  we should also store a hash.txt 
- Then we can always hash(deposit.zok)  and see if it matches the value in hash.txt if not (or hash.txt doesnt exisit) then we must re-do the setup
```
for (let i = 0; i < vks.length; i++) {
    const circuit = circuitsToSetup[i];
    ----> const circuitHash = generateHash(circuit); // computes md5 hash of .zok circuit source file
    ----> const prestoredCircuitHash = loadCircuitHash(i); // loads prestored hash of circuit i from file in output/
    if (!vks[i] || config.ALWAYS_DO_TRUSTED_SETUP || circuitHash !== prestoredCircuitHash ) {
      // we don't have an existing vk so let's generate one
      try {
        logger.info(
          `no existing verification key. Fear not, I will make a new one: calling generate keys on ${circuit}`,
        );
       
        const res2 = await axios.post(
          `${config.PROTOCOL}${config.ZOKRATES_WORKER_HOST}/generate-keys`,
          {
            filepath: circuit,
            curve: config.CURVE,
            provingScheme: config.PROVING_SCHEME,
            backend: config.BACKEND,
          },
        );
        vks[i] = res2.data.vk;
        -----> storeCircuitHash(i, circuitHash);   ------> update circuit hash in output/
      } catch (err) {
        logger.error(err);
      }
    } else logger.info(`${circuit} verification key exists: trusted setup skipped`);
  }
```

eugypalu:
The goal of this pr is to check if there have been any changes to the circuits and if so, go to force a trusted setup. 
To do this is kept within the `/output/circuithash.txt` (in the worker), the list of all circuits with their hash. 
So it checks if the hash of the current circuit differs, or is not in the file, and in case a trusted setup is forced

Resolves #989

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -31,7 +31,6 @@ module.exports = {
   TXHASH_TREE_HASH_TYPE: 'keccak256',
   STATE_GENESIS_BLOCK: process.env.STATE_GENESIS_BLOCK,
   CIRCUITS_HOME: process.env.CIRCUITS_HOME || '/app/circuits/',
-  ALWAYS_DO_TRUSTED_SETUP: process.env.ALWAYS_DO_TRUSTED_SETUP || false,
   LOG_LEVEL: process.env.LOG_LEVEL || 'debug',
   MONGO_URL: process.env.MONGO_URL || 'mongodb://localhost:27017/',
   PROTOCOL: 'http://', // connect to zokrates microservice like this
--- a/config/default.js
+++ b/config/default.js
@@ -31,6 +31,7 @@ module.exports = {
   TXHASH_TREE_HASH_TYPE: 'keccak256',
   STATE_GENESIS_BLOCK: process.env.STATE_GENESIS_BLOCK,
   CIRCUITS_HOME: process.env.CIRCUITS_HOME || '/app/circuits/',
+  ALWAYS_DO_TRUSTED_SETUP: process.env.ALWAYS_DO_TRUSTED_SETUP || false,
   LOG_LEVEL: process.env.LOG_LEVEL || 'debug',
   MONGO_URL: process.env.MONGO_URL || 'mongodb://localhost:27017/',
   PROTOCOL: 'http://', // connect to zokrates microservice like this
--- a/nightfall-deployer/src/circuit-setup.mjs
+++ b/nightfall-deployer/src/circuit-setup.mjs
@@ -104,7 +104,7 @@ async function setupCircuits() {
       },
     );
 
-    if (checkHash.data || config.ALWAYS_DO_TRUSTED_SETUP) {
+    if (checkHash.data || !vks[i] || config.ALWAYS_DO_TRUSTED_SETUP) {
       try {
         logger.info({
           msg: 'No existing verification key. Fear not, I will make a new one: calling generate keys',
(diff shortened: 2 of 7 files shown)


config/default.js
```diff
@@ -31,7 +31,6 @@ module.exports = {
   TXHASH_TREE_HASH_TYPE: 'keccak256',
   STATE_GENESIS_BLOCK: process.env.STATE_GENESIS_BLOCK,
   CIRCUITS_HOME: process.env.CIRCUITS_HOME || '/app/circuits/',
-  ALWAYS_DO_TRUSTED_SETUP: process.env.ALWAYS_DO_TRUSTED_SETUP || false,
```
druiz0992, line 34: We still want to preserve the ability to force a trusted setup

nightfall-deployer/src/circuit-setup.mjs
```diff
@@ -92,8 +93,18 @@ async function setupCircuits() {
   for (let i = 0; i < vks.length; i++) {
     const circuit = circuitsToSetup[i];
 
-    if (!vks[i] || config.ALWAYS_DO_TRUSTED_SETUP) {
-      // we don't have an existing vk so let's generate one
+    const fileBuffer = fs.readFileSync(`./circuits/${circuit}`);
+    const hcircuit = crypto.createHash('md5').update(fileBuffer).digest('hex');
+
+    const checkHash = await axios.post(
```
druiz0992, line 99: There needs to be a way for force a trusted setup. Also, what happens if ther eis not vk[i] defined?

druiz0992 (changes requested):
Some additional questions:
- What happens if there are new circuits where hash doesnt match?

config/default.js
eugypalu: Done, reverted this change

nightfall-deployer/src/circuit-setup.mjs
eugypalu: I restored the ability to force trusted setup. 
from the tests I did in the case vk[i] is undefined there was always also a difference on the hash, but for safety I added the vk[i] check, if undefined trusted setup is forced

eugypalu:
> 

in case new circuits were added: 

-if the hash is present in the file but does not match the one just calculated, the hash contained in the file is updated and the trusted setup is generated.

-if, on the other hand, the hash is not present in the file, it is added to the file and the trusted setup is generated

nightfall-deployer/src/circuit-setup.mjs
druiz0992: since you already have tests, can you also include those? or were they manual?

RogerTaule (approved):

nightfall-deployer/src/circuit-setup.mjs
eugypalu: no, sorry but they were all manual tests

RogerTaule (commented):
Thanks for adding the getCircuitHash enpdoint :)

PR #1033, opened 2022-09-30
Dynamic commits

RogerTaule:


pawelgrzybek:
Looks like you added `paddArray` helper in multiple places, also it exists on others. Maybe move this one to common files and keep it only there once?

israelboudoux (changes requested):
Please, add tests related to that change!

IlyasRidhuan (approved):

Issue #1039, opened 2022-10-03
Use @polygon-nightfall/common-files 


pawelgrzybek:
**Describe the solution you'd like**
Instead of referencing `common-files` using relative path, we can now use a module published to npm registry. This MR replaces this across all test files in root level of a repo. Example:

```js
// before

import logger from '../../../common-files/utils/logger.mjs';
```

```js
// after

import logger from '@polygon-nightfall/common-files/utils/logger.mjs';
```

**How can this be tested**
Run tests and expect them to pass. This change should produce no visible difference to the functionality of nightfall.

PR #1040, opened 2022-10-03
#1039 -  Tests to use @polygon-nightfall/common-files

pawelgrzybek:

It replaces all relative paths to common files with npm module. Please

#1039

Any top level test script.

## Any other comments?

Please keep in mind that this PR adds `common-files` workspace. After running `npm i` this will create a symlink to `common-files` that is linked to scoped `@polygon-nightfall/common-files` module. Please run this one on your local env after this one is merged:

```
nvm use && npm i
```


apps/challenger/package.json
```diff
@@ -10,20 +10,20 @@
   "author": "daveroga",
   "license": "CC0-1.0",
   "dependencies": {
+    "@polygon-nightfall/common-files": "^0.0.1",
     "async-mutex": "^0.3.2",
     "body-parser": "^1.19.2",
-    "common-files": "file:../common-files",
+    "cls-hooked": "^4.2.2",
```
israelboudoux, line 16: Hey @pawelgrzybek, I have a question: the deps `express-fileupload`, `cls--hooked`, `uuid`, `pino`and `pino-pretty` are needed everywhere because the old way we used the `common-files` package. Now that we are moving to this new way, isn't it the time for removing all that possibility unnecessary deps?

pawelgrzybek: Let's have a quick call about it, because I am not sure If I understand. But trust me friend, if I can remove something, I will. I love removing code!

pawelgrzybek: I see your point now. You are absolutely right, we don't need these here anymore. I would prefer to fix this as part of the other issues and other ticket. I would prefer to keep this PR scoped to replacing common-files, and I will open another one to identify unused packages. Does it sound like a good plan to you my friend?

israelboudoux: > I see your point now. You are absolutely right, we don't need these here anymore. I would prefer to fix this as part of the other issues and other ticket. I would prefer to keep this PR scoped to replacing common-files, and I will open another one to identify unused packages. Does it sound like a good plan to you my friend?

Amazing plan! 😊

signorecello (approved):
Way to go, happy for this change. Thanks!

apps/challenger/package.json
signorecello: I also love removing code. I'd remove it all, but then stuff wouldn't work

daveroga (approved):
LGTM

Westlad (changes requested):
Just holding it while I understand something.

imagobea (approved):
Hello @pawelgrzybek, many thanks for dealing with this! LGTM. I was just wondering if you could briefly explain the changes re docker - deleting `pre-start` script in client and some other apps. For the sake of documenting the change and for docker-newbies like myself :nerd_face: Thanks again :boom:

pawelgrzybek:
Hi @Westlad.

As per your request I added back the necessary steps to Dockerfiles, to make sure we always run on the most recent version of common-files (and potentially other npm modules in the future). Even tho we can't skip the part of copy/pasting common files to a container, there is still an improvement in the way how I have done it. We no longer need to rely on `pre-install.sh` scripts that manually adds stuff to `node_modules` (this step messed up mongodb version on client and optymist, which is also fixed as part of my PR).

```
# copy common-files to a container, and create a symlink for it in global node_modules directory
WORKDIR /
COPY common-files common-files

WORKDIR /common-files
RUN npm ci
RUN npm link

# tell the app to fetch package from global node_modules, instead of from npm registry
WORKDIR /app
RUN npm link @polygon-nightfall/common-files
RUN npm ci
```

At some point we can run `npm version patch && npm publish` as part of the pipeline, and then we will be able to skip this linking step in the pipelines triggered by the merge to master. This should save us some time with containers creation.

Hopefully this solution is sufficient for now.
Thank you for your contribution.

pawelgrzybek:
Hi @imagobea.

So after all the changes to Dockerfiles are very minor. The only difference is the way how we inject `@polygon-nightfall/common-files` into the node_modules of particular app. Instead of manually copy/pasting common-files directory into node_module, we use `npm link` to create a symlink between modules. by doing so I was able to delete these `pre-install.sh` scripts.

I am very happy to schedule a very quick demo of this `npm link` feature and explain it to everyone not familiar with it. I am also happy to answer all your questions re this PR.

Thanks
Have a nice day 🥑🥖

Westlad (approved):

PR #1043, opened 2022-10-05
Fees upgrade

RogerTaule:
This PR should be merged after #1033.
Transaction structure contains a fee field which right now can be either matic (L2 token) or ether value, which makes it confusing for users. This PR fixes this behaviour by only storing the fees paid in L2. If the fees are paid in L1 (aka when deposit) the fee in the transaction is set to 0 and fee is stored in a mapping in the Smart Contract.

### DIFF
--- a/nightfall-deployer/contracts/State.sol
+++ b/nightfall-deployer/contracts/State.sol
@@ -23,7 +23,7 @@ contract State is ReentrancyGuardUpgradeable, Pausable, Config {
     mapping(address => FeeTokens) public pendingWithdrawalsFees;
     mapping(address => LinkedAddress) public proposers;
     mapping(address => TimeLockedStake) public stakeAccounts;
-    mapping(bytes32 => FeeTokens) public feeBook;
+    mapping(bytes32 => FeeTokens) public feeBookBlocks;
     mapping(bytes32 => bool) public claimedBlockStakes;
     LinkedAddress public currentProposer; // who can propose a new shield state
     uint256 public proposerStartBlock; // L1 block where currentProposer became current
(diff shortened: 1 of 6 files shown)


imagobea (commented):
Can't comment much on the contract changes but the rest looks okay. I think with this changes to the deposit transaction `DEFAULT_FEE_ETH` can be deprecated. Thanks for explaining this to me @RogerTaule !!

nightfall-deployer/contracts/State.sol
```diff
@@ -223,6 +215,54 @@ contract State is ReentrancyGuardUpgradeable, Pausable, Config {
...
+                            add(0x04, 0x20), //Inputs are 36 bytes long
+                            x, //Store output over input (saves space)
+                            0x20 //Outputs are 32 bytes long
+                        )
+                    )
+
+                    mstore(feePayments, add(mload(x), mload(feePayments)))
+                }
+
+                if eq(
```
IlyasRidhuan, line 242: why is this `eq( isZero (x),0)`  can we just test `isZero`?

RogerTaule: In this case we want to check that is not zero. So, fees will be paid on chain if the transaction is not a deposit, which means transactionType will be not 0. isZero function is 1 if what is evaluated is zero and 0 otherwise. Since we want to get inside the if we need to check that isZero(x) is equal to 0

IlyasRidhuan: ahh ok I get it the `isZero` is true (i.e. 1) so the test for not zero is to compare it with 0

Issue #1037, opened 2022-10-03
Add administrator multisig test to GHA


daveroga:
Add test for the multisig and administrator functionality in Github Actions.

PR #1045, opened 2022-10-05
KYC admin functions

### LINKED ISSUE #742: Implementing white lists
Provide a means to prevent anyone but an authorised user from moving funds between Layer 1 and Layer 2.
There should be a specific group of Ethereum addresses with the power to authorise/de-authorise individuals (whitelist-managers) and these should each be able to manage a specific group of users.  Whitelist managers can be added/removed via the Nightfall Multisig contract using the administrator container.  Additionally, a preliminary set of whitelist managers can be specified at deployment.
Whitelist-managers can also be a smart contract to enable later expansion to multisigs, DAOs and automated KYC.

Westlad:

This provides the `Administrator` container's `./admin` app with extra menu items that enable admin control of the whitelisting process, i.e.:
- Adding and removing whitelist managers
- Enabling/disabling whitelisting
- Checking if an address is a whitelist manager

Before this PR, these items could only be set at the point of contract deployment.
Note that these functions are the ones that must only be called by the contract multisig.  Whitelist managers do not have this level of access.  They access their whitelisting API via the client container (this is already merged).

fixes #742 

Currently it is not possible to test the administrator container automatically.  This will be fixed by the upcoming issue #1037 and so is not addressed in this PR.
To test manually, read the instructions in the `nightfall-administrator` [README.md](https://github.com/EYBlockchain/nightfall_3/blob/westlad/kyc-admin/nightfall-administrator/README.md)

### DIFF
--- a/nightfall-administrator/src/index.mjs
+++ b/nightfall-administrator/src/index.mjs
@@ -5,7 +5,7 @@
  */
 import yargs from 'yargs';
 import { hideBin } from 'yargs/helpers';
-import logger from 'common-files/utils/logger.mjs';
+import logger from '../../common-files/utils/logger.mjs';
 import { initUI } from './ui/menu.mjs';
 import start from './ui/get-info.mjs';
 import { initMultiSig } from './services/helpers.mjs';
@@ -20,8 +20,8 @@ async function main() {
   initUI();
   // start getting transaction information
   const signed = await start();
-  logger.info('******* Signed Transaction *******');
-  logger.info(signed);
+  console.log('******* Signed Transaction *******');
+  console.log(signed);
   process.kill(process.pid, 'SIGTERM');
 }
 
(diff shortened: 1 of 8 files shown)


nightfall-administrator/src/index.mjs
```diff
@@ -20,8 +19,8 @@ async function main() {
   initUI();
   // start getting transaction information
   const signed = await start();
-  logger.info('******* Signed Transaction *******');
-  logger.info(signed);
+  console.log('******* Signed Transaction *******');
```
pawelgrzybek, line 22: Any reason you replaced our `logger` with `console.log` ?

Westlad: Yes, this is actually output that you always need (because you cut and paste it into an email etc to send to the person running the multisig).  If we try to use the logger to print it then that stops working once someone sets LOG_LEVEL>info.  Think of it as the programme's actual desired output (write to console), rather than information about how the programme is doing (log to logger).

signorecello (approved):

daveroga (approved):

PR #1047, opened 2022-10-05
do not register second proposer in ping pong test in AWS

druiz0992:
This PR fixes bug in ping-pong-test, only visible in AWS where there is a proposer running. This test starts a second proposer with the same optimist, and the first one loses connection to optimist and its not able to receive blocks.

No

run ping pong test


daveroga (approved):
LGTM

signorecello (approved):
LGTM 😊

Issue #936, opened 2022-09-14
Define proposer fees functionality


daveroga:
The proposers will define a minimum fee for the user transactions in a block to be proposed when they are the current proposer.
We should define how these fees will work and here are some points for the proposal.
- Create an endpoint `/fee` as a service in the proposer for externally knowing ths fee. This endpoint will return the value of the fee expressed in the token to be paid as fee for the proposers in Polygon Nightfall. In the current moment is MATIC.
- Decide the need to have this fee in the blockchain in the proposers structure as now it's the url. Then probably we won't need to have the endpoint for the fee in the proposer and anyone could access the information through the proposers data in the `State` contract. This will cause gas fees for the proposers every time they need to update the fees using the `updateProposer` function passing this fee to be updated.

    ```
    struct LinkedAddress {
        address thisAddress;
        address previousAddress;
        address nextAddress;
        string url;
        uint256 fee;
    }
    ```

- Decide a maximum fee allowed for a proposer to be defined in order to guarantee users will pay for this fee for their transactions to be included in a L2 block.
- In case a proposer receive an `off-chain` transaction with less than the minimum fee established will throw an error like `fee under accepted minimum fee`.

imagobea:
Hello @daveroga , thank you for this.

Because we will include a view on the block explorer that displays/showcases the proposers who are part of the network, and we want to display their fees too, seems easier to have this data on-chain and perform a single request.

Like you suggest, proposers will have to pay fees every time they want to update their "payment". This would cause friction if they had the need to adjust payments often - _do we know anything about it?_

On the other hand, just pointing out that I believe there are 2 fees (?) Deposit fees in ETH, transfer/withdrawal fees in MATIC, so I guess we'll need these 2 variables.

That's it for now, cheers!!

PR #1064, opened 2022-10-06
Proposer fee endpoint and update

daveroga:
Fixes: #936 
- Add/update fee for the proposer
- Get fee for the proposer. Endpoint defined in the proposer app `/proposer/fee`.

Yes. #936.

npm run test-e2e-protocol

## Any other comments?
No

### DIFF
--- a/nightfall-optimist/src/routes/proposer.mjs
+++ b/nightfall-optimist/src/routes/proposer.mjs
@@ -44,14 +44,14 @@ router.post('/register', async (req, res, next) => {
   logger.debug({ msg: '/register endpoint', payload: JSON.stringify(req.body, null, 2) });
 
   try {
-    const { address, url = '' } = req.body;
+    const { address, url = '', fee = 0 } = req.body;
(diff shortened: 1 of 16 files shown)


signorecello (approved):
LGTM!

nightfall-optimist/src/routes/proposer.mjs
```diff
@@ -98,12 +98,14 @@ router.post('/update', async (req, res, next) => {
   logger.debug({ msg: '/update endpoint', payload: JSON.stringify(req.body, null, 2) });
 
   try {
-    const { address, url = '' } = req.body;
+    const { address, url = '', fee = 0 } = req.body;
     if (url === '') {
```
pawelgrzybek, line 102: I am wondering why do we do this check `if (url === '') { … }` here but we don't do it in the block above?

daveroga: Yes, I agree with you. If the `url` should be provided always and different from '' it should be provided also and checked the same in `register` and `update` proposer.

pawelgrzybek (approved):

Issue #948, opened 2022-09-15
Setup folder structure and tools for the new SDK


signorecello:
A suggestion would be to pick up most of the standards from the user SDK and use that as a basis (regarding linting, structure, test framework, etc).

We need to have this set up before starting to migrate functionality from the old SDK

- [ ] Linting
- [ ] TS

- [ ] Folder structure

- [ ] Testing framework (mocha vs jest)

- [ ] Modularization

imagobea:
No longer valid

Issue #945, opened 2022-09-15
Migrate functionality from the old SDK


signorecello:
As part of the new proposer/challenger SDK, most functionality needs to be migrated from the old SDK into this one. Structure should be already in place

signorecello:
duplicate of #892

imagobea:
@signorecello closing this one as either this is a duplicate of #1025, or it refers to all the features that need "re-writing" - for which I'm creating smaller tickets based on dRuiz requirements. Thanks!

PR #1065, opened 2022-10-07
Optimist SDK

signorecello:
This refactors the nf3 library in a way that can be used separately as an SDK, given that we still don't know if this is gonna be an API or not.

Yes, #948 and #945

You can try, but you'll fail. Tests don't run correctly. But if you try, remember to run `npm run build` inside the `optimist-sdk` folder

## Any other comments?
This is just a handover to @imagobea, obviously there are a lot of things wrong here like:

- [ ] Putting this SDK in a proper folder
- [ ] Making tests run correctly (starting by proposer test, apparently changeCurrentProposer doesn't want anything with me)
- [ ] Making tests run directly from the `ts` files instead of building them before running (bah)
- [ ] Cleaning up


signorecello:
Be aware of #1064 and some other changes that may come when rebasing with master

PR #1068, opened 2022-10-07
fix: optimist local test bug

Westlad:

This fixes a bug in the optimist sync test whereby it would complain that `nightfall_3` was not a valid project name and throw, when testing locally.

No

Run up nightfall as usual (`./start-nightfall -g -d`) and run `npm run test-optimist-sync`

## Any other comments?

It's unclear why this bug arose.  It may be that some characters were accidentally deleted because it's unlikely to have ever worked and yet we know that it used to. I changed double quotes to single quotes as well, but that is immaterial to the functioning, just a linter preference,

### DIFF
--- a/test/optimist-resync.test.mjs
+++ b/test/optimist-resync.test.mjs
@@ -51,7 +51,7 @@ describe('Optimist synchronisation tests', () => {
       'docker/docker-compose.ganache.yml',
     ],
     log: process.env.LOG_LEVEL || 'silent',
-    composeOptions: [['-p "nightfall_3"']],
+    composeOptions: [['-p', 'nightfall_3']],
   };
 
   before(async () => {


israelboudoux (approved):
I don't understand how it has been working until today without this fix.

pawelgrzybek (approved):

pawelgrzybek:
@israelboudoux it was working on the pipeline, but not locally. I can't explain to you why, but that was the case.

israelboudoux:
> @israelboudoux it was working on the pipeline, but not locally. I can't explain to you why, but that was the case.

Haha, yeah, don't tell me why IT isn't such an exact area. I myself added this bit and tested locally, but we know things don't work as expected everywhere. :D

PR #1069, opened 2022-10-08
Improvements for speeding up tests

israelboudoux:
The idea is to build and run only the necessary services for the tests.

No

All the tests used in the pipeline can be simulated locally.

## Any other comments?
No


docker/optimist.Dockerfile
```diff
@@ -1,5 +1,5 @@
 # build zokrates from source for local verify
-FROM ghcr.io/eyblockchain/local-zokrates:0.8.2 as builder
+FROM docker.io/israelboudoux/tmp-nightfall-zokrates:0.8.2 as builder
```
druiz0992, line 2: We should never use things from personal repos.

docker/worker.Dockerfile
```diff
@@ -1,5 +1,5 @@
 # build zokrates from source for local verify
-FROM ghcr.io/eyblockchain/local-zokrates:0.8.2 as builder
+FROM docker.io/israelboudoux/tmp-nightfall-zokrates:0.8.2 as builder
```
druiz0992, line 2: We should never use assets from personal repos.

setup-nightfall
```diff
@@ -1,4 +1,25 @@
 #! /bin/bash
+
+usage()
+{
+  echo "Usage:"
+  echo "  -g or --ganache; for a ganache simulator"
+}
+
+if [ ! -z "$1" ]; then
+  while [ -n "$1" ]; do
```
druiz0992, line 10: Why do we need to force this from the setup phase? We did never require it. What happens if we want to deploy to a different blockchain? If you need to just accelerate test, then create a new setup file with you changes, but do not modify this one

druiz0992, line 10: Why do we need to force this from the setup phase? We did never require it. What happens if we want to deploy to a different blockchain? If you need to just accelerate test, then create a new setup file with you changes, but do not modify this one

israelboudoux: This was a mistake and will be removed.

docker/optimist.Dockerfile
israelboudoux: Yes, I was just testing, I'll pull from our ECR.

signorecello:
Too bad the gas test is the bottleneck but damn this makes tests much faster

signorecello (approved):

Issue #1012, opened 2022-09-28
Transfer/Withdraw that can use specific commitment hash


IlyasRidhuan:
Currently, a user supplies a value, token and recipient to the `transfer` and `withdraw` calls. These calls then do the hard work of finding the optimal commitments to be spent (based on selection rubric).

Some use-cases may implement their own commitment selection logic and thus these function calls need to be expanded to support taking in a specific `commitment_hash` or `commitment_hashes`.

PR #1070, opened 2022-10-10
transferring/withdrawing specific commitments

siosw:
this allows users to specify the specific commitments they want to transfer/withdraw.

#1012 

an e2e test can be run like this:
`LOG_LEVEL=warn npx hardhat test test/client/commitments.test.mjs --no-compile`
a unit test is WIP

## Any other comments?
This should be supported through the SDK, see #1071

### DIFF
--- a/nightfall-client/src/services/transfer.mjs
+++ b/nightfall-client/src/services/transfer.mjs
@@ -29,7 +29,7 @@ const NEXT_N_PROPOSERS = 3;
 async function transfer(transferParams) {
   logger.info('Creating a transfer transaction');
   // let's extract the input items
-  const { offchain = false, ...items } = transferParams;
+  const { offchain = false, providedCommitments, ...items } = transferParams;
   const { tokenId, recipientData, rootKey, fee } = generalise(items);
   const ercAddress = generalise(items.ercAddress.toLowerCase());
   const { recipientCompressedZkpPublicKeys, values } = recipientData;
@@ -61,6 +61,7 @@ async function transfer(transferParams) {
     tokenId,
     rootKey,
     maxNumberNullifiers: VK_IDS.transfer.numberNullifiers,
+    providedCommitments,
   });
 
   try {
--- a/nightfall-client/src/services/withdraw.mjs
+++ b/nightfall-client/src/services/withdraw.mjs
@@ -28,7 +28,7 @@ const NEXT_N_PROPOSERS = 3;
 async function withdraw(withdrawParams) {
   logger.info('Creating a withdraw transaction');
   // let's extract the input items
-  const { offchain = false, ...items } = withdrawParams;
+  const { offchain = false, providedCommitments, ...items } = withdrawParams;
   const { tokenId, value, recipientAddress, rootKey, fee } = generalise(items);
   const ercAddress = generalise(items.ercAddress.toLowerCase());
   const shieldContractInstance = await waitForContract(SHIELD_CONTRACT_NAME);
@@ -53,6 +53,7 @@ async function withdraw(withdrawParams) {
     tokenId,
     rootKey,
     maxNumberNullifiers: VK_IDS.withdraw.numberNullifiers,
+    providedCommitments,
   });
 
   try {
--- a/nightfall-client/src/utils/getCommitmentInfo.mjs
+++ b/nightfall-client/src/utils/getCommitmentInfo.mjs
@@ -5,8 +5,10 @@ import constants from '@polygon-nightfall/common-files/constants/index.mjs';
 import Nullifier from '../classes/nullifier.mjs';
 import {
   clearPending,
+  markPending,
   findUsableCommitmentsMutex,
   getSiblingInfo,
+  getCommitmentsByHash,
 } from '../services/commitment-storage.mjs';
 import Commitment from '../classes/commitment.mjs';
 import { ZkpKeys } from '../services/keys.mjs';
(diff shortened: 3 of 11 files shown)


nightfall-client/src/services/transfer.mjs
```diff
@@ -51,16 +56,35 @@ async function transfer(transferParams) {
...
+  const txInfo = {
+    compressedZkpPublicKey,
+    ercAddress,
+    tokenId,
+    ercAddressFee: maticAddress,
+    value,
+    fee: feeValue,
+  };
+  logger.debug(`using user provided commitments: ${providedCommitments === true}`);
+  const commitments = providedCommitments || (await findUsableCommitmentsMutex(txInfo));
```
RogerTaule, line 75: Is there any reason why you are moving `findUsableCommitments` from getCommitmentInfo to transfer.mjs? Since we will also add a couple of new circuits, if possible I'd like to avoid code duplication. Couldn't we pass `providedCommitments` as a parameter for getCommitmentInfo and do this if inside that function?

nightfall-client/src/services/withdraw.mjs
```diff
@@ -45,14 +50,35 @@ async function withdraw(withdrawParams) {
...
+  const txInfo = {
+    compressedZkpPublicKey,
+    ercAddress,
+    tokenId,
+    ercAddressFee: maticAddress,
+    value: valueWithFee,
+    fee: feeValue,
+  };
+  logger.debug(`using user provided commitments: ${providedCommitments === true}`);
+  const commitments = providedCommitments || (await findUsableCommitmentsMutex(txInfo));
```
RogerTaule, line 68: Same as mentioned above for transfer.mjs

nightfall-client/src/services/withdraw.mjs
```diff
@@ -28,7 +33,7 @@ const NEXT_N_PROPOSERS = 3;
 async function withdraw(withdrawParams) {
   logger.info('Creating a withdraw transaction');
   // let's extract the input items
-  const { offchain = false, ...items } = withdrawParams;
+  const { offchain = false, providedCommitments, ...items } = withdrawParams;
```
RogerTaule, line 36: I think that `providedCommitments` should be an array of commitment hashes. Therefore, in order to work properly with the current logic, you will need to get the whole commitment information from the database. Currently, there's no function that gets a commitment by hash, so you'll need to create one and add it to `commitment-storage.mjs`. You can take a look at how `getCommitmentBySalt` works. Then, you'll need to check that the user is the owner of that commitment and otherwise throw an error.

RogerTaule (changes requested):

nightfall-client/src/services/withdraw.mjs
siosw: this is something that I wanted to discuss with you as well. I think having an array of hashes instead of the full objects is a good idea. I basically already have that implemented in the test I'm writing so I'll work on that once the tests are done.

nightfall-client/src/services/transfer.mjs
siosw: not sure what you mean with code duplication? does this change how the circuits work?

my general idea here was to make the commitment logic obvious in the transfer function, that's why I moved it there. but I'm open to moving it back and passing the providedCommitments if they exist.

nightfall-client/src/services/withdraw.mjs
RogerTaule: Great! Passing `providedCommitments` as an array of hashes and handling the commitments inside the client code will simplify things on the user side :)

nightfall-client/src/services/transfer.mjs
RogerTaule: Well, we are also adding burn and tokenise circuits, which may use this functionality and therefore it will be duplicated code. But anyway, I think it is better to handle this logic inside the `getCommitmentInfo` function. In a future we may want to handle some more complex scenarios and it's better to handle all this logic in a single place.

nightfall-client/src/utils/getCommitmentInfo.mjs
```diff
@@ -42,14 +45,36 @@ export const getCommitmentInfo = async txInfo => {
...
-    feeValue,
-  );
+  logger.debug(`using user provided commitments: ${providedCommitments !== undefined}`);
+
+  let commitments;
+  if (providedCommitments) {
+    // look up the hashes
+    const commitmentHashes = providedCommitments.map(c => c.toString());
+    logger.debug({ msg: 'looking up these commitment hashes:', commitmentHashes });
+    const rawCommitments = await getCommitmentsByHash(commitmentHashes);
```
RogerTaule, line 55: I think there are several scenarios to be considered here: 
- What happens if a user provides a fake commitment hash?
- What happens if a user provides a commitment for which he isn't the owner? Or a commitment whose preimage is a different ercAddress? 
- What happens if a commitment is already nullified or pending nullification?
- What happens if the total sum of the commitments isn't enough for what the user is intending to transfer/withdraw?

Hint: Take a look at `verifyEnoughCommitments` function in the commitment storage

RogerTaule (changes requested):

nightfall-client/src/utils/getCommitmentInfo.mjs
```diff
@@ -84,7 +138,9 @@ export const getCommitmentInfo = async txInfo => {
     );
 
     // we may need to return change to the recipient
-    const change = totalInputCommitmentValue - value;
+    const change = totalInputCommitmentValue - (totalValueToSend + addedFee);
```
RogerTaule, line 141: Isn't this the same? `value = totalValueToSend + addedFee`

RogerTaule (changes requested):

PR #1073, opened 2022-10-10
Removing stubs

RogerTaule:
Stubs idea was to speed development and don't get stuck due to the circuit processing time taking too long. However, circuits constraints has been reduced a lot with the use of Poseidon and other optimizations, and generating a proof is fast enough. Therefore, there is no longer need for having those circuits.
This PR removes those circuits and cleans the code


common-files/classes/transaction.mjs
```diff
@@ -79,38 +80,23 @@ class Transaction {
...
-    else commitments = _commitments;
-    if (_nullifiers === undefined || _nullifiers.length === 0)
-      nullifiers = [{ hash: 0 }, { hash: 0 }, { hash: 0 }, { hash: 0 }];
-    else if (_nullifiers.length === 1)
-      nullifiers = [..._nullifiers, { hash: 0 }, { hash: 0 }, { hash: 0 }];
-    else if (_nullifiers.length === 2) nullifiers = [..._nullifiers, { hash: 0 }, { hash: 0 }];
-    else if (_nullifiers.length === 3) nullifiers = [..._nullifiers, { hash: 0 }];
-    else nullifiers = _nullifiers;
+
+    const commitments = utils.padArray(_commitments, { hash: 0 }, numberCommitments);
```
signorecello, line 93: ❤️

siosw:
> Hey, I cheer this idea, removing stubs is less code and less code is good. However I'm worried if it would have any significant impact on the deployment? Do you have a ballpark number on how much this would increase the deployment time?

@signorecello on my machine at least running nightfall with stubs is not faster than with the full circuits. I think thats because snarkjs proof generation is so much faster.

druiz0992 (approved):

PR #1074, opened 2022-10-10
Testnet client

druiz0992:
PR builds a client from AWS container, and verifies testnet client is reachable and working.

During this PR I found that gas test fails with some high probability. The reason is that there is not enough time for the transaction with the withdrawal to be added to the client's database, and the finalizewithdrawal test fails. I added a sleep after the blockProposed event received by the client to give some time to have the block added into the database.

No

None


israelboudoux (approved):

docker/docker-compose.client.dev.yml
```diff
@@ -3,9 +3,6 @@ version: '3.5'
 # bindings.  See the readme for more information.
 services:
   client:
-    build:
```
signorecello, line 6: Without this, I believe local changes on the client dockerfile wouldn't be reflected? The `-d` flag means we always want to build from the local dockerfile instead of using the image

zokrates-worker/src/index.mjs
```diff
@@ -44,7 +44,8 @@ const checkCircuitsOutput = async () => {
       files.map(async f => {
         if (f) {
           const filename = f.split('  ')[1];
-          const circuit = circuits.find(c => filename.includes(c));
+          const _circuit = circuits.find(c => filename.includes(c));
+          const circuit = typeof _circuit === 'undefined' ? '.' : _circuit;
```
signorecello, line 48: I don't understand this line... Why do we set `.`?

signorecello (changes requested):
Just some quick questions

zokrates-worker/src/index.mjs
druiz0992: this is my poor attempt at fixing the error that happens in worker when trying to download a file that doesnt exist. It didnt work, but i forgot to remove it. I will remove it as it is useless.

docker/docker-compose.client.dev.yml
druiz0992: You need to choose. Either we want to simplify user to always have a working version of client without the need to build from a known commit, or they use our provided container.

signorecello: Sorry I'm not getting it. As it is, we can use the `-d` flag to build and use the container, as it is in the current dir, and we bind mount and (I think) use `nodemon` to restart with changes. Without the `-d` flag it pulls the client from somewhere (in this case, AWS) and uses that.

Not sure if this `-d` flag could be added to `start-client` in the very same way

druiz0992: I think you missed the purpose of the PR. The idea is that the client container is not built on the spot anymore. It is always downloaded from AWS. The benefits of this is that now you have always a synced version of the client with testnet,

signorecello: I think I get it, we're assuming someone working on the client would be using the `./start-nightfall` script, and use this docker-compose file only for pulling and using the image from AWS. I'm ok with this then 😊

signorecello (approved):
LGTM

PR #1076, opened 2022-10-12
Optimizing and adding tests for Shield contract

RogerTaule:
This PR adds a few optimizations to the Shield Contract and tests the whole contract.
The Shield Contract was modified as follows:
-` advanceWithdrawal` mapping has been merged into a single mapping of an struct called AdvancedWithdraw and the fees has been reduced to uint96 to fit in a single 32 bytes. This reduces 20.000 gas since we are using a single slot.
- `advanceWithdrawal` function can only be called while the block has not finalized. Otherwise liquidity provider would earn free fees. Same applies to `setAdvanceWithdrawalFee`.
- When `finaliseWithdrawal` is called, if an advanced fee was set it is returned to the owner and the mapping is cleaned.
- `areBlockAndTransactionReal` returns transaction hash so that it is not calculated twice in the same function
- Some requires has been reorganized and some variables have been removed

Some of the future work for the Shield Contract:
- Adding a function that allows users to claim their advanced fee (now it is returned if the transaction is not advanced before the block is finalized, but if during the challenge period a user change its mind cannot get the fee back)
- Document the Smart Contract properly following NatSpec format.
- All the state calls can be converted to assembly, which would save us some gas
No
Tests can be run using `npm run unit-test-shield`
## Any other comments?
Hopefully this tests and utils can be used as an starting point for testing other Smart Contracts. 
Some of the tests uses a hardhat functionality that allows us to manually set a variable in the storage. This functionality needs to know the storage slot, and since we have many contracts inheriting it is not easy to know. Best way is by calling: `npx hardhat check` and looking for the particular variable. Is it possible that if in a future new variables are stored the current storage slots will no longer be valid and need to be updated


Westlad (approved):

daveroga (approved):
LGTM

PR #1085, opened 2022-10-14
Proposer & Optimist little fixes

israelboudoux:
Fixes for Proposer & Optimist that were crashing the applications in certain conditions.

No

The tests are the usual ones performed by Github Actions

## Any other comments?
No


nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -129,16 +130,19 @@ export async function conditionalMakeBlock(proposer) {
...
-          await new Promise(resolve => setTimeout(resolve, 3000)); // eslint-disable-line no-await-in-loop
+          await waitForTimeout(3000); // eslint-disable-line no-await-in-loop
 
-          logger.warn(`Websocket to proposer is closed.  Waiting for proposer to reconnect`);
+          logger.warn(`Websocket to proposer is closed. Waiting for proposer to reconnect`);
 
           increaseProposerWsClosed();
           if (tryCount++ > 100) {
             increaseProposerWsFailed();
-            throw new Error(`Websocket to proposer has failed`);
```
druiz0992, line 139: Why don’t you throw error?

nightfall-optimist/src/event-handlers/transaction-submitted.mjs
```diff
@@ -68,15 +68,20 @@ async function transactionSubmittedEventHandler(eventParams) {
     transaction = await checkAlreadyInBlock(transaction);
     // save transaction if not in block
     if (fromBlockProposer) {
-      saveTransaction({ ...transaction });
+      saveTransaction({ ...transaction }).catch(err =>
```
druiz0992, line 71: There is already a catch of this error. I don't think this code will behave the same

nightfall-optimist/src/event-handlers/transaction-submitted.mjs
```diff
@@ -68,15 +68,20 @@ async function transactionSubmittedEventHandler(eventParams) {
...
     }
 
     await checkTransaction(transaction, true);
+
     logger.info('Transaction checks passed');
 
     // save it
     if (!fromBlockProposer) {
-      saveTransaction({ ...transaction });
+      saveTransaction({ ...transaction }).catch(err =>
```
druiz0992, line 82: Same as above

druiz0992 (changes requested):
I fail to see the purpose of this PR.

nightfall-optimist/src/services/block-assembler.mjs
israelboudoux: Because this crashes the Optimist. I don't think this is the right behavior, no one would like to have its app crashing just because it wasn't possible to connect to a Proposer. There is no `try...catch` to catch this. It should retry until it gets a connection. Btw, I'll add an `logger.error` to warn about this.

nightfall-optimist/src/event-handlers/transaction-submitted.mjs
israelboudoux: No, the `catch` this instruction is inside only handles synchronous calls, this one is an async call and returns a promise that crashes the app if an error happen, then that is one of the possible ways to handle it.

nightfall-optimist/src/event-handlers/transaction-submitted.mjs
israelboudoux: Replied!

nightfall-optimist/src/event-handlers/transaction-submitted.mjs
druiz0992: Then, should be remove the `try-catch` block if its not doing anything?  And shouldnt the `checkTransaction` function catch an error as well?

israelboudoux: We can't remove because `checkAlreadyInBlock` and `checkTransactio` are called with `await` and this `try...catch` handles errors coming from them.

nightfall-optimist/src/event-handlers/transaction-submitted.mjs
```diff
@@ -68,15 +68,20 @@ async function transactionSubmittedEventHandler(eventParams) {
...
+        logger.error({ msg: 'Error while saving transaction', err }),
+      );
```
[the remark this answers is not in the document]
israelboudoux: Hey, sorry @imagobea, I just saw your comment now. Yes, I find it confusing as well, but I didn't take much attention and I was trying to avoid touching things that work.

druiz0992 (approved):

PR #1091, opened 2022-10-17
Proposers unit test

eugypalu:
This PR adds unit tests for the Proposers.sol smart contract and fixes 2 small bugs in the smart contract

No

npm run unit-test-proposers             

## Any other comments?
No


test/unit/SmartContracts/proposers.unit.test.mjs
```diff
@@ -0,0 +1,427 @@
...
+    );
+    const [proposer] = eventTransfer.args;
+
+    expect(proposer).to.equal(addr1.address);
+
+    expect((await state.getProposer(addr1.address)).thisAddress).to.equal(addr1.address);
+
+    await ProposersInstance.deRegisterProposer();
+
+    await ethers.provider.send('evm_increaseTime', [604800]); // + 1 week
```
daveroga, line 330: This only works for ganache. We should write tests that work in all environments. So then we should wait for the blocks after challenge period or skip this test if it's not ganache.

daveroga (changes requested):

test/unit/SmartContracts/proposers.unit.test.mjs
eugypalu: fixed, I added the check on chain id, and the test is run only in case of ganache or hardhat, otherwise the test is skipped

daveroga (approved):

rael-b (approved):
Amazing to see these tests @eugypalu.

Issue #1084, opened 2022-10-14
Fix blocking submitTransaction calls


druiz0992:

druiz0992:
in Nf3, submitTransaction is used to send transactions to blockchain. Among others, the proposer calls this function to submit blocks. The problem that we are facing is that these calls are blocking, so at most 1 L2 block will be sent per L1 block, which will slow the process. The solution needs to workout a scheme where submitTransaction is not blocking, but managed transaction nonces.

Issue #1062, opened 2022-10-06
Optimization in Block class


druiz0992:

druiz0992:
block class defined in optimist contains this:
```    // compute the keccak hash of the proposeBlock signature
    const blockHash = this.calcHash({
      proposer,
      root: updatedTimber.root,
      leafCount: updatedTimber.leafCount,
      blockNumberL2,
      previousBlockHash,
      transactionHashesRoot: await this.calcTransactionHashesRoot(transactions),
      frontierHash: this.calcFrontierHash(updatedTimber.frontier),
    });
    this.localPreviousBlockHash = blockHash;
    // note that the transactionHashes array is not part of the on-chain block
    // but we compute it here for convenience. It needs removing before sending
    // a block object to the blockchain.
    return new Block({
      proposer,
      transactionHashes: transactions.map(t => t.transactionHash),
      transactionHashesRoot: await this.calcTransactionHashesRoot(transactions),
      leafCount: updatedTimber.leafCount,
      root: updatedTimber.root,
      blockHash,
      nCommitments,
      blockNumberL2,
      previousBlockHash,
      frontierHash: this.calcFrontierHash(updatedTimber.frontier),
})
```
It seems redundant to compute `this.calcTransactionHashesRoot(transactions)` and `this.calcFrontierHash(updatedTimber.frontier),` twice

PR #1093, opened 2022-10-17
Multiple L2 blocks per l1 block

druiz0992:
Improves speed at which L2 blocks are generated. It focuses on two parts:
- Block Assembly: Allows to send multiple L2 blocks in a single L1 block. For this, we had to add nonce computation logic and split sign and send transaction functionalities.
- Block Proposed: Removes several unnecessary sleeps

While doing load testing, it was found that test stops after 100 blocks due to the MINIMUM_STAKE for the proposer. Thus, we increased it to 1000000

Yes, #1062 and #1084

Normal tests

### DIFF
--- a/cli/lib/nf3.mjs
+++ b/cli/lib/nf3.mjs
@@ -5,6 +5,7 @@ import WebSocket from 'ws';
 import ReconnectingWebSocket from 'reconnecting-websocket';
 import EventEmitter from 'events';
 import logger from '@polygon-nightfall/common-files/utils/logger.mjs';
+import { Mutex } from 'async-mutex';
 import { approve } from './tokens.mjs';
 import erc20 from './abis/ERC20.mjs';
 import erc721 from './abis/ERC721.mjs';
@@ -27,7 +28,7 @@ import {
 // then there will only be one queue here. The constructor does not need to initialise clientBaseUrl
 // for proposer/liquidityProvider/challenger and optimistBaseUrl, optimistWsUrl for a user etc
 const userQueue = new Queue({ autostart: true, concurrency: 1 });
-const proposerQueue = new Queue({ autostart: true, concurrency: 1 });
+const proposerQueue = new Queue({ autostart: true });
 const challengerQueue = new Queue({ autostart: true, concurrency: 1 });
 const liquidityProviderQueue = new Queue({ autostart: true, concurrency: 1 });
 
@@ -85,6 +86,10 @@ class Nf3 {
   contracts = { ERC20: erc20, ERC721: erc721, ERC1155: erc1155 };
 
   currentEnvironment;
+
+  nonce = 0;
+
+  nonceMutex = new Mutex();
 
   constructor(
     ethereumSigningKey,
@@ -150,6 +155,7 @@ class Nf3 {
   async setEthereumSigningKey(key) {
     this.ethereumSigningKey = key;
     this.ethereumAddress = await this.getAccounts();
+    this.nonce = 0;
   }
 
   /**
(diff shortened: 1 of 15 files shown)


druiz0992:
> Just taking a look out of curiosity, since there are changes in`nf3`.
> 
> Method `submitTransaction` is actually used in multiple places, just thinking a) if we could also benefit from the split somewhere, b) if we should update SDK submit function.
> 
> Cheers!

@imagobea definitely we should add these changes to the proposer app as they ensure that we can add multiple l2 blocks. Regarding user sdk, they will be needed at client side so that multiple transactions can be submitted in same l1 block. My problem is that I believe we are changing this part now, and I prefer to add it once client is finished

cli/lib/nf3.mjs
```diff
@@ -255,35 +261,64 @@ class Nf3 {
...
   This can be found using the getContractAddress convenience function.
+  @returns {Oject} Signed transaction.
+  */
+  async _signTransaction(unsignedTransaction, contractAddress, fee) {
+    let tx;
+    let signed;
+
+    await this.nonceMutex.runExclusive(async () => {
+      // estimate the gasPrice
+      const gasPrice = await this.estimateGasPrice();
```
israelboudoux, line 280: The calls to `estimateGasPrice` and `estimateGas` can be moved outside of the mutex, because they don't share state between threads, and as more operations inside a mutex more time it takes for releasing the lock then slow down the whole operation.

cli/lib/nf3.mjs
```diff
@@ -255,35 +261,64 @@ class Nf3 {
...
+      // Estimate the gasLimit
+      const gas = await this.estimateGas(contractAddress, unsignedTransaction);
+
+      // if we don't have a nonce, we must get one from the ethereum client
+      if (!this.nonce) this.nonce = await this.web3.eth.getTransactionCount(this.ethereumAddress);
```
israelboudoux, line 285: It would be nice to test this on a _near_ real blockchain. According to _[Ethereum Book, Transactions chapter](https://github.com/ethereumbook/ethereumbook/blob/develop/06transactions.asciidoc)_: 
_"Be careful when using the getTransactionCount function for counting pending transactions, because you might run into some problems if you send a few transactions in a row."_
...
_"When you build an application that constructs transactions, it cannot rely on getTransactionCount for pending transactions. Only when the pending and confirmed counts are equal (all outstanding transactions are confirmed) can you trust the output of getTransactionCount to start your nonce counter. Thereafter, keep track of the nonce in your application until each transaction confirms."_

cli/lib/nf3.mjs
```diff
@@ -928,18 +979,22 @@ class Nf3 {
       const { type, txDataToSign, block, transactions, data } = msg;
       logger.debug(`Proposer received websocket message of type ${type}`);
       if (type === 'block') {
+        // First sign transaction, and send it within asynchronous queue. This will
+        // ensure that blockProposed events are emitted in order and with the correct nonce.
+        const tx = await this._signTransaction(
```
israelboudoux, line 984: Maybe the _nonce_ itself can control the sequence of the blocks and we don't need to worry about this, I "guess".

cli/lib/nf3.mjs
```diff
@@ -255,35 +261,64 @@ class Nf3 {
...
+        to: contractAddress,
+        data: unsignedTransaction,
+        value: fee,
+        gas,
+        gasPrice,
+        nonce: this.nonce,
+      };
+      this.nonce++;
+
+      if (this.ethereumSigningKey) {
```
israelboudoux, line 298: As the `gasPrice/estimateGas` methods, this could be moved outside the mutex too, the `tx` variable is safe to be used.

israelboudoux (changes requested):
Looks good, but maybe we need a proper way for managing nonces.

cli/lib/nf3.mjs
druiz0992: I’m not 100% sure. I left it out and then we had unordered transactions. So, I prefer to have this like this.

cli/lib/nf3.mjs
druiz0992: We will certainly test this. But we already applied this solution in the past

druiz0992: In any case, the reason we have this nonce control is exactly the one you posted. If minces don’t match, we take ours

cli/lib/nf3.mjs
druiz0992: Not sure what `this` is. But we need to split the nonce computation and the transaction submission. The nonce computation is done serially, and then we don’t care about the transaction ordering because it comes given by the nonce

nightfall-optimist/src/services/check-block.mjs
```diff
@@ -210,7 +207,6 @@ export async function checkBlock(block, transactions) {
   await checkLeafCount(block);
   // now we have to check the commitment root.
   // For this we can make use of Timber with its optimistic extensions.
-  await new Promise(resolve => setTimeout(resolve, 1000));
```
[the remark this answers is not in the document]
druiz0992: That timeout should not affect performance. It’s there if the WS fails,and we attempt to reconnect. It shouldn’t fail. I’m any case, we need to tackle WS connection in the proposer app

nightfall-optimist/src/classes/block.mjs
```diff
@@ -131,15 +131,18 @@ class Block {
...
     // compute the keccak hash of the proposeBlock signature
     const blockHash = this.calcHash({
       proposer,
       root: updatedTimber.root,
       leafCount: updatedTimber.leafCount,
       blockNumberL2,
       previousBlockHash,
-      transactionHashesRoot: await this.calcTransactionHashesRoot(transactions),
-      frontierHash: this.calcFrontierHash(updatedTimber.frontier),
+      transactionHashesRoot: _transactionHashesRoot,
```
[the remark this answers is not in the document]
druiz0992: Ok, I will change this. Thanks

druiz0992:
> Looks good, but maybe we need a proper way for managing nonces.

What do you mean by proper way? I think this way is correct

israelboudoux (approved):

daveroga (approved):
LGTM

PR #1098, opened 2022-10-18
Obfuscating sensitive data whatever the environment

israelboudoux:
Currently, we don't obfuscate sensitive data for local environment. With this change all sensitive data will be obfuscated whatever the environment.

No

## Any other comments?
No


common-files/utils/utils.mjs
```diff
@@ -25,15 +26,19 @@ export const isLocal = () => {
...
-    (typeof object !== 'object' && typeof object !== 'string') ||
-    !obfuscationSettings ||
     typeof obfuscationSettings !== 'object' ||
     Array.isArray(obfuscationSettings) ||
     obfuscationSettings.length === 0
   ) {
+    throw new Error('Invalid obfuscationSettings object!');
+  }
+
+  if (!object || (typeof object !== 'object' && typeof object !== 'string')) {
```
imagobea, line 41: :100: This new `obfuscate` version is spot on!!

imagobea (approved):
LGTM!!

druiz0992 (changes requested):
Look at logs from erc721 test, line 1291. I can see private keys there. I guess you need to also review the tests

common-files/utils/utils.mjs
signorecello: absolutely, great abstraction right here, spot on @israelboudoux

israelboudoux:
> Look at logs from erc721 test, line 1291. I can see private keys there. I guess you need to also review the tests

I'll verify!

common-files/utils/utils.mjs
israelboudoux: Yes, thanks @imagobea and @signorecello, but this only abstracts obfuscation for the automatic HTTP req/res logging. I don't want to make this automatic for every log because of performance issues, since the obfuscation runs through all the properties of an object recursively, for a too deep object this can be a problem.

druiz0992 (approved):

PR #1100, opened 2022-10-18
Log Adjustments

israelboudoux:
- Removes the fields `pid` and `hostname` from JSON logs;
- Logs are now obfuscated only for non-local environments;
- Removing the `JSON.stringfy` so that the logs will be cleaner;
- Added env variable `LOG_HTTP_PAYLOAD_ENABLED` for allowing disabling the automatic logs of HTTP payloads;
- Removing unnecessary log statements from routes, since they are automatically performed when `LOG_LEVEL == 'debug'` & `LOG_HTTP_PAYLOAD_ENABLED == true`;
- Removing `logger.error(err)` from the `try...catch` in the routes since they are handled in the `errorHandler` middleware;
- Add more unit tests regarding obfuscation;
- Adding handling to `logger` for moving the `correlation-id` to the end of the log message;
- Added configuration for allowing shorten the output of Request/Response logging. This is set via `LOG_HTTP_FULL_DATA` (default to `false`) - It is customizable via environment variable. Output sample:
```
client_1                   | [2022-10-19 15:09:03.912] DEBUG: Call to endpoint '/contract-address/Shield'
client_1                   |     correlationId: "783d321c-580d-4096-b61b-6cc7535eac48"
client_1                   |     params: {}
client_1                   | [2022-10-19 15:09:03.923] DEBUG: Result from endpoint /contract-address/Shield [200]
client_1                   |     correlationId: "783d321c-580d-4096-b61b-6cc7535eac48"
client_1                   |     result: {
client_1                   |       "address": "0x864D794aB771516ADbD058366E72ba697273EB53"
client_1                   |     }
```

No

## Any other comments?
No


common-files/utils/httputils.mjs
```diff
@@ -101,27 +105,49 @@ const applyAxiosDefaults = () => {
   });
 };
 
+/* eslint no-else-return: "off" */
+const getRequestParam = req => {
+  const dataInput = {};
+
+  if (Object.keys(req.query).length !== 0) dataInput.query = req.query;
+  if (Object.keys(req.params).length !== 0) dataInput.params = req.params;
+  if (Object.keys(req.body).length !== 0) dataInput.params = req.body;
```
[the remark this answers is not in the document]
imagobea: Also, maybe this is for readability `Object.keys(req.body).length !== 0`, but I guess I could just be `Object.keys(req.body).length`

docker/client.Dockerfile
```diff
@@ -20,7 +20,6 @@ WORKDIR /app
 COPY nightfall-client/src src
 COPY nightfall-client/docker-entrypoint.sh nightfall-client/package.json nightfall-client/package-lock.json ./
 
-RUN npm link @polygon-nightfall/common-files
```
imagobea, line 23: Just leaving a comment to remind us of this change (since we are modifying this from different PRs)

nightfall-optimist/src/services/check-block.mjs
```diff
@@ -52,7 +52,13 @@ async function checkBlockRoot(block) {
...
       history = await getTreeByLeafCount(block.leafCount);
       logger.debug(`Block has commitments - retrieved history from Timber`);
-      logger.trace(`Timber history was ${JSON.stringify(history, null, 2)}`);
+      logger.trace({
+        msg: 'Timber history was',
+        history,
+      });
+
+      // eslint-disable-next-line no-await-in-loop
+      await new Promise(resolve => setTimeout(resolve, 3000));
```
imagobea, line 61: Why do we need this now? And if we do, didn't you create a util? :upside_down_face:

imagobea (approved):
Hello @israelboudoux , very good! 
I've left only a handful of comments.
The most important in `common-files/utils/httputils.mjs` > when building `dataInput` you seem to overwrite params with body. Please check that one out!!
Cheers :pinata:

common-files/utils/httputils.mjs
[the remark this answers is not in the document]
israelboudoux: You're right, it was a mistake! Nice catch!!!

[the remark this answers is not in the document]
israelboudoux: I wanted to have tests for this but I wasn't able to think in a way to "catch" the log output.

nightfall-optimist/src/services/check-block.mjs
israelboudoux: I didn't change this point, but you're right, however we need a refactoring task for replacing these instructions with the new method in utils. :D

docker/docker-compose.adversary.yml
```diff
@@ -29,11 +29,14 @@ services:
         source: ../config/default.js
         target: /app/config/default.js
     environment:
+      ENVIRONMENT: ${ENVIRONMENT:-local}
```
signorecello, line 32: I think you meant `localhost` instead of `local`?

docker/docker-compose.apps.yml
```diff
@@ -10,6 +10,10 @@ services:
       # to use with postman and etc
       - 8092:8092
     environment:
+      ENVIRONMENT: ${ENVIRONMENT:-local}
```
signorecello, line 13: same, check the other docker-compose files as they have the same var

signorecello (changes requested):
just some quick things

signorecello (approved):

PR #1101, opened 2022-10-18
Adding environment variables to docker-compose

israelboudoux:
This change allows passing environment variables into some docker-compose files.

No

The usual tests performed via pipeline

## Any other comments?
No


druiz0992 (commented):
Does proposer + geth work with this change?

israelboudoux:
> Does proposer + geth work with this change?

Why wouldn't it work? For all env vars there is a default value.

druiz0992:
@israelboudoux because I don’t think it’s working and wondered if this fixed it

daveroga (approved):

druiz0992 (approved):

PR #1105, opened 2022-10-19
feat: added estimate gas fn to common-files

imagobea:

Continues the [task that started in SDK repo](https://github.com/maticnetwork/nightfall-sdk/pull/133). I have moved the `estimateGas` function and its tests to `common-files`.

No

`cd common-files && npm ci`, then run the tests with `npm run test`.

## Any other comments?

First experience with `mocha` and `sinon`, please take a look at the stubs and assertions :pray: Thank you!!


common-files/utils/ethereum/gas.mjs
```diff
@@ -0,0 +1,32 @@
+import logger from '../logger.mjs';
+
+export const TX_GAS_DEFAULT = 4000000;
```
israelboudoux, line 3: I think some of these constants already exist [here](https://github.com/EYBlockchain/nightfall_3/blob/master/cli/lib/constants.mjs).

israelboudoux (approved):
LGTM, there is just a simple detail to double-check.

common-files/utils/ethereum/gas.mjs
pawelgrzybek: @israelboudoux, @imagobea did the right thing. We should not be importing constants from another top-level package. Alternatively we can take from from config. Personally I would keep it as is. Two repetitions doesn't hurt my DRY-mania :)

pawelgrzybek:
Oooo, that is going to be a good opportunity to test if the automated common-files version on merge to master works as expected 🤞

cli/lib/nf3.mjs
```diff
@@ -276,11 +233,6 @@ class Nf3 {
     let signed;
 
     await this.nonceMutex.runExclusive(async () => {
-      // estimate the gasPrice
```
druiz0992, line 279: Why did you remove this?

cli/lib/nf3.mjs
```diff
@@ -292,10 +244,12 @@ class Nf3 {
         to: contractAddress,
         data: unsignedTransaction,
         value: fee,
-        gas,
-        gasPrice,
         nonce: this.nonce,
       };
+      // Estimate the gasLimit
+      const gas = await estimateGas(tx, this.web3);
```
druiz0992, line 250: Don’t we need gasPrice as well?

imagobea: @druiz0992 

A quick search revealed that gas calculations were outdated due to London Upgrade of August 2021. Quick summary [1]:

**Prior to the London upgrade**

```
Total fee = Gas units (limit) * Gas price per unit
```

**After the London upgrade**

```
Total fee = Gas units * (base fee + priority fee)
```

_Where the base fee is a value set by the protocol and the priority fee is a value set by the user as a tip to the validator._

**What this means re transaction object**

Before we had `gas` and `gasPrice`. Now you can set [2]:

- `maxFeePerGas`: (optional, defaulted to (`2 * block.baseFeePerGas) + maxPriorityFeePerGas`) The maximum fee per gas that the transaction is willing to pay in total

- `maxPriorityFeePerGas`: (optional, defaulted to 2.5 Gwei) The maximum fee per gas to give miners to incentivise them to include the transaction (Priority fee)

**About this PR**

Limits gas calculations to `estimateGas` function via web3 - this is for calculating "gas units", a.k.a. `gas`.
In the future, we can consider using dynamic values for `maxPriorityFeePerGas` and `maxFeePerGas`.

imagobea: **References**

[1] https://ethereum.org/en/developers/docs/gas
[2] https://web3js.readthedocs.io/en/v1.7.3/web3-eth.html#id88

signorecello (commented):
LGTM

cli/lib/nf3.mjs
druiz0992: i still seem to remember that when we deployed last time to testnet (beginning of november), we had to update gasPrice because otherwise it wouldn't work.
```
WARN [10-26|14:36:39.709] Served eth_call                          reqid=22578            duration="199.712µs"  err="err: max fee per gas less than block base fee: address 0xe327221Fc90f4D938E23A03C58a4e93A66AAEC20, maxFeePerGas: 20000000000 baseFee: 186671733869 (supplied gas 8000000)"
```

imagobea:
Hopefully, continues in #1262

PR #1136, opened 2022-10-24
Eugenio/state contract unit test

eugypalu:
This PR adds unit tests for the State.sol smart contract 
No
npm run unit-test-state
## Any other comments?
No

### DIFF
--- a/test/unit/SmartContracts/state.unit.test.mjs
+++ b/test/unit/SmartContracts/state.unit.test.mjs
@@ -4,769 +4,769 @@ import hardhat from 'hardhat';
 const { ethers, upgrades } = hardhat;
 
 describe('Proposers contract Proposers functions', function () {
-    let ProposersInstance;
-    let addr1;
-    let addr2;
-    let state;
-    let addressS;
-    let addressC;
-    let block;
-    const transaction1 = {
-        value: '100000000000000',
-        fee: '10',
-        transactionType: '0',
-        tokenType: '0',
-        historicRootBlockNumberL2: [
-            '0x0000000000000000000000000000000000000000000000000000000000000000',
-            '0x0000000000000000000000000000000000000000000000000000000000000000',
-            '0x0000000000000000000000000000000000000000000000000000000000000000',
-            '0x0000000000000000000000000000000000000000000000000000000000000000',
-        ],
-        tokenId: '0x0000000000000000000000000000000000000000000000000000000000000000',
-        ercAddress: '0x000000000000000000000000499d11e0b6eac7c0593d8fb292dcbbf815fb29ae',
-        recipientAddress: '0x0000000000000000000000000000000000000000000000000000000000000000',
-        commitments: [
-            '0x078ba912b4169b22fb2d9b6fba6229ccd4ae9c2610c72312d0c6d18d85fd22cf',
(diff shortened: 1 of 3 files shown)


daveroga (changes requested):

test/unit/SmartContracts/state.unit.test.mjs
```diff
@@ -0,0 +1,772 @@
...
+      0,
+    ]);
+    await state.setNumProposers(2);
+    await state.setStakeAccount(addr1.address, amount, challengeLocked);
+    await state.setStakeAccount(addr2.address, amount, challengeLocked);
+    await state.setCurrentProposer(addr1.address);
+
+    expect((await state.getCurrentProposer()).thisAddress).to.equal(addr1.address);
+
+    await expect(state.changeCurrentProposer()).to.be.reverted;
```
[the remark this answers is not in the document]
eugypalu: I am not sure if this test makes sense.
I included it because maybe it might make sense to include a `require`.
the revert occurs in the `spanProposerSet` function at line `664` when it tries to access the `slots` array.
it tries to access the array by iterating a number of times equal to `proposerSetCount` (which is set to 5).
               
while the number of slots is calculated like this: 
`weight = stake.amount / valuePerSlot;` (with `valuePerSlot` set to 10)

then we could enter a `require` where the stake account must be equal to or greater than 
`proposerSetCount* valuePerSlot`

otherwise there will be fewer slots generated than expected

signorecello:
Closes #1138

test/unit/SmartContracts/state.unit.test.mjs
```diff
@@ -0,0 +1,1057 @@
...
+    const Utils = await ethers.getContractFactory('Utils');
+    const utils = await Utils.deploy();
+    await utils.deployed();
+
+    const State = await ethers.getContractFactory('State', {
+      libraries: {
+        Utils: utils.address,
+      },
+    });
+    state = await upgrades.deployProxy(State, [addr1.address, addressC.address, addressS.address], {
```
RogerTaule, line 116: The addresses passed to the state contract should be the proposer address, challenge address and shield address

RogerTaule (changes requested):

test/unit/SmartContracts/state.unit.test.mjs
eugypalu: I added different addresses from chellenge, shield, and proposer because I wanted to test individual functions directly on the state without going through other smart contract functions, avoiding potential side effects.
to test some functions, like proposeblock I mocked the data of blocks and transaions and I think I entered wrong data.
using the createBlockAndTransactions function and calling proposeBlock (not the bug fixed version, the old one) actually gives error on the last revert.
I will fix using createBlockAndTransactions for the mocked data and testing both reverts.
thanks

RogerTaule (approved):
This PR is ready to go. However, it should not be until the adversary test is fixed so that it works with the fix done in the State contract in PR #1158. This fix was was reverted to unblock all other PRs because that required a major fix in the adversary test side, but on the downside it broke this PR.

daveroga (approved):
Waiting for the fix in State contract to pass the unit-tests. All it's ok.

PR #1142, opened 2022-10-25
Circom 

RogerTaule:
This PR contains the following changes:
- Migrates all the circuits from Zokrates to Circom 2.1.0. 
- Add unit tests for all the circuits and templates
- Creates a new circom docker container to be able to compile the circuits
- Adds Fq2 field operations to the Pairing.sol so that when verifying a proof we check if the proof is built with valid G1 and G2 points and otherwise it returns false. Prior to this change, if the proof was incorrectly built it was unchallengeable.


docker/docker-compose.client.yml
```diff
@@ -18,7 +18,7 @@ services:
       - mongodb
     environment:
       LOG_LEVEL: ${LOG_LEVEL:-debug}
-      ZOKRATES_WORKER_HOST: worker
+      CIRCOM_WORKER_HOST: worker
```
druiz0992, line 21: The availability client test is failing because of this change. It is expecting a different ENV variable to connect to worker. I am fine pushing this as is, and once we deploy to testnet with correct ENV variable (`CIRCOM_WORKER_HOST`), it will work again

druiz0992 (approved):

Issue #1128, opened 2022-10-24
GHA check gas for 32 tx per block sometimes failing


daveroga:
Test check gas for 32 transaction per block sometimes failing with `Error: socket hang up` during the test

![image.png](https://images.zenhubusercontent.com/615b1480539a0a0f1387eadf/059be97b-5e4e-487a-b05b-557764da12b1)

druiz0992:
![Screenshot from 2022-10-24 10-09-04](https://user-images.githubusercontent.com/46495002/197490789-a08aaa82-438d-4c06-a2e2-0544fbbd0a1c.png)

Issue #1125, opened 2022-10-24
GHA optimist-sync test sometimes failing


daveroga:
optimist-sync test sometimes is failing with `Error: socket hang up` just at the beginning of the test.

![image.png](https://images.zenhubusercontent.com/615b1480539a0a0f1387eadf/0b6d0762-23c4-4657-a317-452bbcf07ba2)

PR #1146, opened 2022-10-25
Fixed gas and resync tests

druiz0992:
- gas test was occasionally failing because of a timeout in mongodb. This error seemed to happen after we created a separate mongoDb instance.  I increased this timeout to 1 second.
- optimist-sync test seem to occasionally fail for the same reason.
- Remove unnecessary code from gas test to make it faster. Before, we were testing single and double transfers. But there is no such distinction anymore. To be able to test both transfers, we needed to have two rounds of deposits. I just left one. 

#1128 and #1125

GHA will test these changes, specifically `gas-test` and `optimist-sync` test

### DIFF
--- a/common-files/utils/mongo.mjs
+++ b/common-files/utils/mongo.mjs
@@ -21,7 +21,16 @@ export default {
       });
       connection[url] = await client.connect();
     } else {
-      const client = await new MongoClient(url, { useUnifiedTopology: true });
+      const options = { server:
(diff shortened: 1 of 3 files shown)


druiz0992:
I launched the tests three times without error, so i consider this PR to have fixed the problems with mongoDb timeout.

daveroga (approved):

Westlad (approved):

Issue #1061, opened 2022-10-06
Cache contracts


druiz0992:

druiz0992:
Following https://app.zenhub.com/workspaces/pnf-6295dbd03832a40014480c20/issues/eyblockchain/nightfall_3/925, we see that lots of time is spent retrieving contract information. It would be a good idea once a contract (address, interface, instance) is read, cache it for next time.

The downside is that if a contract is updated, this mechanism will not detect it. Maybe there is a workaround, or maybe we need to force a reboot in optimist

PR #1159, opened 2022-10-27
remove duplicate wait for contract

druiz0992:
Two main changes:
- removed duplicated implementation of `waitForContract` function. Kept the one in `common files`
- Implemented a contract cache mechanism to avoid having to load contracts constantly. The first time a contract is requested, the contract interface will be cached, and from there on, any additional request will be retrieved from the cached version. One endpoint is added to allow to clear cached contracts

#1061 and #1063


daveroga (approved):

RogerTaule (approved):
LGTM

PR #1166, opened 2022-10-27
Removing transaction type

RogerTaule:
This PR gets rid of the `transactionType` property in the transaction struct and replaces it with a new property called `circuitHash`, which is the top 5 bytes of the hash of the circuit added in #1029. 
The reason of this change is to remove the TransactionType enum that was stored in the Smart Contracts, which was forcing us to deploy the contracts every time a new circuit was added. 
When registering the circuit, we do not only register the verificationKey but also to extra properties: `isEscrowRequired` and `isWithdrawing`. If the first one is set to true means that funds needs to be escrowed in order to add the transaction, while only those txs belonging to a circuit that has `isWithdrawing` set to true will be allowed to withdraw funds.
This PR also optimizes the transaction structure by packing the first 4 variables (`uint112 value`, `uint96 fee`, `uint40 circuitHash` and `uint8 tokenType`) into a single variable `uint256 packedInfo`


nightfall-deployer/migrations/2_deploy_upgradeable.js
```diff
@@ -44,7 +44,7 @@ module.exports = async function (deployer) {
 
   await deployProxy(Proposers, [], { deployer, unsafeAllowLinkedLibraries: true });
   await deployProxy(Challenges, [], { deployer, unsafeAllowLinkedLibraries: true });
-  await deployProxy(Shield, [], { deployer, unsafeAllowLinkedLibraries: true });
+  await deployProxy(Shield, { deployer, unsafeAllowLinkedLibraries: true });
```
daveroga, line 47: Why this change here and not in `Proposers` or `Challenges` about the empty parameters `[ ]`?

wallet/src/nightfall-browser/services/commitment-storage.js
```diff
@@ -533,7 +528,7 @@ export async function getWithdrawCommitments() {
   const vals = await db.getAll(COMMITMENTS_COLLECTION);
   const withdraws =
     Object.keys(vals).length > 0
-      ? vals.filter(v => v.isNullified && v.isOnChain >= 0 && v.nullifierTransactionType === '2')
+      ? vals.filter(v => v.isNullified && v.isOnChain >= 0 && v.nullifierCircuitHash === '2')
```
daveroga, line 531: Is this correct? It seems `v.nullifierCircuitHash` is a hash and won't be '2'.

daveroga (changes requested):

nightfall-optimist/src/routes/proposer.mjs
```diff
@@ -350,30 +350,26 @@ router.post('/offchain-transaction', async (req, res) => {
...
-        res.sendStatus(200);
-        break;
-      }
-      default:
-        res.sendStatus(400);
-        break;
+      await enqueueEvent(transactionSubmittedEventHandler, 1, {
+        offchain: true,
+        ...transaction,
+        fee: Number(fee),
```
IlyasRidhuan, line 369: This is an out of scope change for this PR - but this should be `BigInt` in the future if we are support > 64bit fees

test/e2e/tokens/erc721.test.mjs
```diff
@@ -89,7 +89,7 @@ describe('ERC721 tests', () => {
     await emptyL2();
   });
 
-  describe('Deposit', () => {
+  describe.skip('Deposit', () => {
```
IlyasRidhuan, line 92: remove skip

test/adversary/adversary-code/database.mjs
```diff
@@ -85,10 +82,7 @@ const duplicateNullifier = async number => {
   try {
     const connection = await mongo.connection(MONGO_URL);
     const db = connection.db(OPTIMIST_DB);
-    const res = await db
-      .collection(TRANSACTIONS_COLLECTION)
-      .find({ transactionType: { $in: ['1', '2'] } })
-      .toArray();
+    const res = await db.collection(TRANSACTIONS_COLLECTION).find().toArray();
```
IlyasRidhuan, line 85: we will need to filter this for transactions that only have nullifiers for this test

config/default.js
```diff
@@ -95,26 +95,36 @@ module.exports = {
...
       numberCommitments: 2,
+      isEscrowRequired: false,
+      isWithdrawing: true,
     },
     tokenise: {
       txType: 3,
       numberNullifiers: 2,
       numberCommitments: 2,
+      isEscrowRequired: false,
+      isWithdrawing: true,
```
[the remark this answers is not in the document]
RogerTaule: Right, that's what happens when you copy paste xD

PR #1167, opened 2022-10-27
Retrying mongo connection timeout

druiz0992:

### DIFF
--- a/common-files/utils/mongo.mjs
+++ b/common-files/utils/mongo.mjs
@@ -22,11 +22,11 @@ export default {
       connection[url] = await client.connect();
     } else {
       const options = {
-        connectTimeoutMS: 10000,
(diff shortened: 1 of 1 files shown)


israelboudoux (changes requested):
The MongoClient docs mention that the property for connection timeout is `connectTimeoutMS`. `socketTimeoutMS` is used when one already has a socket connection open and defines the timeout to handle a send/receive operation.  

```
   ...
    /** The time in milliseconds to attempt a connection before timing out. */
    connectTimeoutMS?: number;
    /** The time in milliseconds to attempt a send or receive on a socket before the attempt times out. */
    socketTimeoutMS?: number;
    ...
```

israelboudoux (approved):

PR #1176, opened 2022-11-01
Block size bytes

RogerTaule:
WIP


nightfall-deployer/contracts/Shield.sol
```diff
@@ -49,6 +49,14 @@ contract Shield is Stateful, Config, ReentrancyGuardUpgradeable, Pausable, KYC {
                 payIn(t);
             }
         }
+
+        uint256 maxBlockSize = MAX_BLOCK_SIZE;
```
IlyasRidhuan, line 53: I think this must be moved up - possibly above Line 44, otherwise I think the `payIn` will still occur and 1) we will take people's funds even though it will error and 2) it will trigger an escrow of the funds which means a deposit can then be sent off-chain and bypass this check.

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -81,14 +73,53 @@ export async function conditionalMakeBlock(proposer) {
...
-    const unprocessed = await numberOfUnprocessedTransactions();
-    let numberOfProposableL2Blocks = Math.floor(unprocessed / TRANSACTIONS_PER_BLOCK);
-    // if we want to make a block right now but there aren't enough transactions, this logic
-    // tells us to go anyway
-    if (makeNow && unprocessed > 0 && numberOfProposableL2Blocks === 0)
-      numberOfProposableL2Blocks = 1;
-
-    if (numberOfProposableL2Blocks >= 1) {
+    // Get all the mempool transactions sorted by fee
+    const mempoolTransactions = await getSortedMempoolTransactions();
```
IlyasRidhuan, line 77: can we rename this to show this is sorted by fee, e.g. `getSortedByFeeMempoolTransactions()`

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -81,14 +73,53 @@ export async function conditionalMakeBlock(proposer) {
...
+      totalBytes,
+      makeNow,
+    });
+
+    const transactionBatches = [];
+    if (totalBytes > 0) {
+      let bytesSum = 0;
+      for (let i = 0; i < mempoolTransactionSizes.length; ++i) {
+        if (bytesSum + mempoolTransactionSizes[i] > MAX_BLOCK_SIZE) {
+          bytesSum = mempoolTransactionSizes[i];
```
IlyasRidhuan, line 105: not sure this line is needed, at this point we don't need to accumulate or modify `bytesSum`

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -81,14 +73,53 @@ export async function conditionalMakeBlock(proposer) {
...
+          transactionBatches.push(i);
```
IlyasRidhuan, line 106: should be have a `break` after this line at least to terminate the loop early. Otherwise we will continue running the loop despite already reaching the limit for our `transactionBatches`

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -81,14 +73,53 @@ export async function conditionalMakeBlock(proposer) {
...
+        } else {
+          bytesSum += mempoolTransactionSizes[i];
+        }
+      }
+
+      if (bytesSum > 0 && makeNow) {
+        transactionBatches.push(mempoolTransactionSizes.length);
```
IlyasRidhuan, line 113: doesn't this have the possibility of creating a `transactionBatches` that will have a total bytes bigger than what we allow, especially if the previous `for-loop` has already pushed to `transactionBatches`?  Do we need an extra condition in the `if` that is `transactionBatches.length === 0` ?

nightfall-optimist/src/services/database.mjs
```diff
@@ -403,10 +388,14 @@ export async function removeNullifiersFromMemPool(nullifiers) {
 /**
 How many transactions are waiting to be processed into a block?
 */
-export async function numberOfUnprocessedTransactions() {
+export async function getSortedMempoolTransactions() {
```
IlyasRidhuan, line 391: rename to highlight it is sorting by fee

nightfall-deployer/contracts/Shield.sol
RogerTaule: But, isn't the transaction atomic? Meaning if it fails everything is reverted? Anyway, I think it is better to have it prior the loop

PR #1178, opened 2022-11-02
Transaction Structure optimization

### LINKED ISSUE #1174: [FEATURE] Upgrading Transaction structure in the Smart Contracts
When using storage, Solidity automatically packs the variables (so for example if we are storing two `uint128` variables it only uses a single slot of 256 bits. However, when using calldata or memory this same variables are not packed, and therefore we use two memory slots of 256 bits.

Currently, the 4 first variables of our transaction structure are the following: `uint112 value`, `uint112 fee`, `uint8 transactionType` and `uint8 tokenType`. If we do the sum: 112 + 112 + 8 + 8 = 240 < 256. Therefore, one can see that all those 4 fields can be fit in a single uint256 slot. 

The proposal is to get rid of those 4 first variables and create a single uint256 which would contain all four variables concatenated. This means the first 112 bits would correspond to the value, next 112 bits to the fee and so on. In order to do the optimization, there will be two main changes to be done: 
- From the Node code perspective, make sure the solidity struct is built correctly and the calldata is destructured correctly too.
- From the Smart Contract perspective, the way those 4 variables are accessed will need to be modified so that the variable value will be gotten from the uint256. 

This change will get rid of 3 slots in the transaction structure, which implies a reduction at least 1500 gas per transaction (+ a few more gas due to doing keccaks of less slots).

RogerTaule:
This PR optimizes the transaction struct by storing value (uint112), fee (uint96), circuitHash(uint40) and tokenType(uint8) in a single uint256 slot. This should reduce the gas used per transaction in proposeBlock by around 1500 gas. This also reduces the cost of the calldata by around a 10 - 15 % whenever the transaction struct is passed.
Yes, it closes #1174 
It is enough if all the tests go through

### DIFF
--- a/.github/workflows/on-pull-request-master.yml
+++ b/.github/workflows/on-pull-request-master.yml
@@ -2,7 +2,7 @@ name: Check PR
 
 on:
   pull_request:
-    branches: [master, tokenisation-l2]
+    branches: [master, tokenisation-l2, RogerTaule/circuitHash]
 
 jobs:
   dependency-review:
--- a/common-files/classes/transaction.mjs
+++ b/common-files/classes/transaction.mjs
@@ -23,6 +23,15 @@ const arrayEquality = (as, bs) => {
   return false;
 };
 
+const packInfo = (value, fee, circuitHash, tokenType) => {
+  const valuePacked = generalise(value).hex(14).slice(2);
+  const feePacked = generalise(fee).hex(12).slice(2);
+  const circuitHashPacked = generalise(circuitHash).hex(5).slice(2);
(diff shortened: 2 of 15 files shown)


daveroga (approved):

israelboudoux (approved):
LGTM, but I have a suggestion that I think could improve maintainability. A new structure could be created for accommodating the transaction packed info (e.g. 
```
struct TransactionPackedInfo {        
        uint112 value;
        uint96 fee;
        uint40 circuitHash;
        TokenType tokenType;
}
```

then in the `Transaction` struct we could have a method `getPackedInfoStruct()` that returns the built `packedInfo` object and we can refer to it everywhere.

PR #1183, opened 2022-11-03
Applying manual change to common-files version 0.0.7

israelboudoux:
Changes `common-files` current version to `0.0.7`. The trigger solution implemented recently is failing because of the `branch protection rules` in place.

No

No

## Any other comments?
We need to figure out how to do this automatically. Github allows us to create a new `role` with a more relaxed permission that might be used in this case however, this is not the best way.

### DIFF
--- a/common-files/package.json
+++ b/common-files/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@polygon-nightfall/common-files",
-  "version": "0.0.6",
+  "version": "0.0.7",
   "scripts": {
     "test": "NODE_CONFIG_DIR=../config mocha --timeout 0 --bail --exit test/*.mjs"
   },


pawelgrzybek:
SHouldnt that be done automatically on the autocommit step on your recently added automation?

israelboudoux:
> SHouldnt that be done automatically on the autocommit step on your recently added automation?

Yes, but as I explained in the comments the repo has  `branch protection rules` that prevent us from performing it.

israelboudoux:
There are some people asking for a new feature in GA: https://github.com/community/community/discussions/13836

We have this [option](https://github.com/marketplace/actions/branch-protection-bot), but we'd need someone with super-power for creating the TOKEN with full access to the repo.

daveroga (approved):

imagobea (approved):

PR #1191, opened 2022-11-04
x509 based kyc

### LINKED ISSUE #1046: Smart contract based certificate checking.
**Is your feature request related to a problem? Please describe.**
We want to have a smart contract manage whitelisting by allowing entities with a PKI cert that links back to a trusted root to be automatically whitelisted.

**Describe the solution you'd like**
A solution would allow one present a certificate chain of (probably x509) certificates and have a smart contract authenticate them.

**Describe alternatives you've considered**
Other possibilities are Polygon Id as an authorisation method. Additionally, it may be necessary to use lighter-weight certs than x509.

**How can this be tested**
This should have an appropriate unit test.  It is not expected that the solution is integrated into Nightfall initially.

**Additional context**
This is initially an experiment, to determine if such a smart contract can be created and can be gas-efficient.

Westlad:

**Software architecture changes:**
This PR implements KYC checking via x509 certificates. It incorporates two new contracts: `X509.sol` and `DERParser.sol`.  `X509.sol` inherits from `DERParser.sol`.  `DERParser.sol` is a general contract for parsing DER encoded bytes (such as an x.509 certificate). `X509.sol` adds the capability to check the validity of a parsed x.509 certificate. `X509.sol` also inherits from `Whitelisting.sol`, which gives it the functionality to manage the whitelisting functionality the used to be in `Shield.sol`. `X509.sol` is a better home for it because it's not obvious that managing a whitelist should be a concern of `Shield.sol`, whereas it is closely related to KYC functionality. `X509.sol` is required to implement the `KYCInterface.sol` interface.  This exposes a single function `kycCheck(address)` which `Shield.sol` can call to make a KYC check before allowing a transaction (or not). In the future, we could swap out the `X509.sol` certificate check for some other form of KYC and we would not need to alter `Shield.sol` at all.

**Additional Functionality**
Functionality is covered in the README.md file

fixes #1046

Normal tests apply but you can run the specific unit test `npx hardhat test test/unit/SmartContracts/X509.unit.test.mjs` to test these new contracts.

## Any other comments?
Note that currently only RSA certificates are supported. EC could be added in the future.  However RSA is sufficient for now.


druiz0992 (approved):

nightfall-deployer/migrations/2_deploy_upgradeable.js
```diff
@@ -10,11 +10,11 @@ const Proposers = artifacts.require('Proposers.sol');
 const Challenges = artifacts.require('Challenges.sol');
 const State = artifacts.require('State.sol');
 const SimpleMultiSig = artifacts.require('SimpleMultiSig.sol');
-const KYC = artifacts.require('KYC.sol');
+const X509 = artifacts.require('X509.sol');
 
 const config = require('config');
 
-const { RESTRICTIONS, MULTISIG, WHITELIST_MANAGERS } = config;
+const { RESTRICTIONS, MULTISIG, WHITELIST_MANAGERS, RSA_TRUST_ROOTS } = config;
```
druiz0992, line 17: I understand RSA_TRUST_ROOTS somehow initializes the chain of trust between X509 contract and end user. My problem is that I fail how this RSA_TRUST_ROOTS is to be initialized in production.

druiz0992 (changes requested):

cli/lib/nf3.mjs
```diff
@@ -1414,7 +1416,7 @@ class Nf3 {
       address,
     });
     const txDataToSign = res.data;
-    return this.submitTransaction(txDataToSign);
+    return this.submitTransaction(txDataToSign, this.kycContractAddress);
```
druiz0992, line 1419: groupId never used

nightfall-deployer/contracts/X509.sol
```diff
@@ -0,0 +1,345 @@
...
+            certificatePublicKey
+        );
+        expires[msg.sender] = expiry;
+        keysByUser[msg.sender] = subjectKeyIdentifier;
+        addUserToWhitelist(msg.sender); // all checks have passed, so they are free to trade for now.
+    }
+
+    // performs an ongoing KYC check (is the user still in the whitelist? Has the public key been revoked? Is the cert in date?)
+    // We could also remove the user in this function, but that would burn more gas.
+    function kycCheck(address user) external view returns (bool) {
```
druiz0992, line 327: If I understand correctly, a user that has been manually whitelisted but has not validated its certificate will pass `isWhitelisted` check, but will fail `kycCheck`.  Shield contract only cares about `kycCheck` result to allow transactions. So, there is no reason for the SDK to implement whitelisting features (no scenario where a manually whitelisted user without valid certificate can operate) and should focus on KYC, as whitelisting will be done via certificate validation.

nightfall-deployer/contracts/X509.sol
```diff
@@ -0,0 +1,345 @@
...
+        if (
+            !whitelisting ||
+            (!revokedKeys[keysByUser[user]] &&
```
druiz0992, line 330: Do you think there should be a way to recover revoked addresses in case we redeploy contracts?

nightfall-deployer/migrations/2_deploy_upgradeable.js
Westlad: In production RSA_TRUST_ROOTS, would contain the root public keys of the CAs that we have decided to trust ( Digicert, Verisign etc.).  These are the public keys in their self-signed root certificates and they're publicly available.  For example Digicert's is [here](https://www.digicert.com/kb/digicert-root-certificates.htm).  `Owner` can also add remove keys after deployment (although we need to add this functionality to the admin container).
For now, I just made my own CA (self-signed cert) and told the contract to trust it, then made an intermediate CA certificate from that and finally an end user certificate.  At some point soon, we should test with an end-user cert derived from a real CA but that is a little involved (identity checks etc) and so, in the interestsof time, I haven't done that here.

cli/lib/nf3.mjs
Westlad: Good spot, thanks! `X509.sol` should give itself a groupId.

nightfall-deployer/contracts/X509.sol
Westlad: I agree it's not worth implementing now, if we don't already have it. However, it would be useful if we decided to remove certificate checking because then we'd replace X509.sol with a much simpler contract exposing the kycCheck interface that just checked `isWhitelisted`.

nightfall-deployer/contracts/DerParser.sol
```diff
@@ -0,0 +1,181 @@
...
+            Tag memory,
+            uint256,
+            uint256
+        )
+    {
+        bytes1 tagClass = tagByte & 0xC0;
+        bool isConstructed = (tagByte & 0x20) != 0;
+        bytes1 tagType = tagByte & 0x1F;
+        uint256 headerLength = 0;
+        require(tagType < 0x1F, 'Tag is Long Form, which is not supported');
```
imagobea, line 41: I think we have started to add the name of the contract at the beginning of error messages. You might want to apply this to all new instances.

```suggestion
        require(tagType < 0x1F, 'DerParser: Tag is Long Form, which is not supported');
```

nightfall-deployer/contracts/X509.sol
```diff
@@ -0,0 +1,345 @@
...
+    bytes1 usageBitMaskEndUser;
+    bytes1 usageBitMaskIntermediate;
+
+    function initialize() public override(Whitelist) initializer {
+        Whitelist.initialize();
+        usageBitMaskEndUser = 0xC0;
+        usageBitMaskIntermediate = 0x06;
+    }
+
+    function setUseageBitMaskEndUser(bytes1 _usageBitMask) external onlyOwner {
```
imagobea, line 33: (?)

```suggestion
    function setUsageBitMaskEndUser(bytes1 _usageBitMask) external onlyOwner {
```

nightfall-deployer/contracts/X509.sol
```diff
@@ -0,0 +1,345 @@
...
+        usageBitMaskEndUser = _usageBitMask;
+    }
+
+    function setUseageBitMasIntermediate(bytes1 _usageBitMask) external onlyOwner {
```
imagobea, line 37: usage, mask

```suggestion
    function setUsageBitMaskIntermediate(bytes1 _usageBitMask) external onlyOwner {
```

imagobea (commented):
Just a few questions, observations. Great feature indeed. Thank you!

nightfall-deployer/contracts/X509.sol
Westlad: Yes, good call.  The list of revoked addresses is public so it can be read anyway, so nothing to do to the smart contract but it's a useful admin function to include.  I'll add it to the admin PR but not here.

cli/lib/nf3.mjs
```diff
@@ -1439,6 +1441,36 @@ class Nf3 {
...
+        'sha256',
+        Buffer.from(ethereumAddress.toLowerCase().slice(2), 'hex'),
+        {
+          key: privateKey,
+          padding: crypto.constants.RSA_PKCS1_PADDING,
+        },
+      );
+    }
+    // now validate the cert
+    const res = await axios.post(`${this.clientBaseUrl}/whitelist/validate`, {
```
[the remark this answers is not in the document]
Westlad: The route is `/whitelist` but only for historical reasons.  I can update it to `/kyc` which is more accurate.

cli/lib/nf3.mjs
```diff
@@ -1439,6 +1441,36 @@ class Nf3 {
     });
     return res.data.isWhitelisted;
   }
+
+  /**
+   Validates an X509 (RSA) certificate
+   */
+  async validateCertificate(certificate, ethereumAddress, derPrivateKey) {
```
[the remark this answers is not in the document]
Westlad: No problem!:
- Yes, type is Buffer but I suspect a string will also work.  The web3js api will probably work it out.
- There's no problem, you can whitelist as many addresses as you like. It's not free to do that though, so you would probably want to limit how many you register.
- So, your certificate contains a public key, which the certificate attests to.  derPrivateKey is the corresponding private key to the public key in the certificate.  It's nothing to do with your Ethereum private key. It needs to be available to Nf3 so that Nf3 can use it to sign your Ethereum address.  This action proves that the owner of the Ethereum address is also the person attested to in the certificate.

cli/lib/nf3.mjs
Westlad: I've removed this, and related, whitelist functions entirely because they are no longer required - `X509.sol` subsumes this functionality.

nightfall-deployer/contracts/X509.sol
```diff
@@ -0,0 +1,345 @@
...
+            12 -
+            (3 * ((_year + 4900 + (_month - 14) / 12) / 100)) /
+            4 -
+            OFFSET19700101;
+
+        _seconds = uint256(__days) * SECONDS_PER_DAY;
+    }
+
+    // this function finds and checks the Not Before and Not After tlvs
+    // TODO this and the subsequent function loop over the tlvs twice - this is inefficient - refactor code
```
israelboudoux, line 148: I don't see a problem with the loop in these functions, the worst Time Complexity is `O(n)` and I might be wrong but the `tlvs` array doesn't seem to get too large, anyways inside the loop you don't do any intensive operation. I can't see much room for improvement honestly, then, if you agree, you can remove the TODO. :)

nightfall-deployer/contracts/X509.sol
Westlad: Good spot, thanks!

israelboudoux (approved):
Hey @Westlad, what a terrific work. congrats! there are just a few things to verify.

nightfall-client/src/routes/kyc.mjs
```diff
@@ -45,4 +50,23 @@ router.post('/remove', async (req, res, next) => {
...
 });
 
+/**
+ Validate a certificate (which will also add the user to the whitelist if the certificate is valid and the signature over
+ their ethereum address checks out). The signature can be falsey if we don't want to whitelist the address but are just validating
+ the certificate.  We might want to do this for an intermediate certificate for example.
+ */
+router.post('/validate', async (req, res, next) => {
+  const { certificate, ethereumAddressSignature } = req.body;
+  if (!certificate.type === 'Buffer') next(new Error('Certificate is not a buffer'));
```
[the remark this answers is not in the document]
Westlad: I don't really follow I'm afraid.  I'd expect an error if you don't sent a certificate so that seems like correct behaviour.

doc/ssl-contract.md
```diff
@@ -0,0 +1,25 @@
+Ethereum uses the EC secp256k1 curve.
+
+Could just have a CA directly add addresses to a whitelist. Extend this idea so that you can have chains of addresses, each added by the next one up in the turust heirachy.
+
+ssl certs x509 signing algos.
```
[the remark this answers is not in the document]
Westlad: Oh, that shouldn't be in the PR, I'll delete, thanks.

nightfall-client/src/services/kyc.mjs
```diff
@@ -1,23 +1,38 @@
...
 
 export async function removeUserFromWhitelist(address) {
-  const shieldContractInstance = await waitForContract(SHIELD_CONTRACT_NAME);
-  return shieldContractInstance.methods.removeUserFromWhitelist(address).encodeABI();
+  const kycContractInstance = await waitForContract(KYC_CONTRACT_NAME);
+  return kycContractInstance.methods.removeUserFromWhitelist(address).encodeABI();
+}
+
+export async function validateCertificate(certificate, ethereumAddressSignature) {
+  const kycContractInstance = await waitForContract(KYC_CONTRACT_NAME);
```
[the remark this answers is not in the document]
Westlad: No,  if it's null then the call will throw as expected. There's no need to check that.  The whole point of the call is to validate the certificate, it will throw if the certificate is invalid, for whatever reason.

nightfall-deployer/contracts/DerParser.sol
```diff
@@ -0,0 +1,181 @@
...
+            for (uint256 i = 0; i < MAX_DEPTH; i++) {
+                if (pointer == depthChangesAt[i]) depth--;
+            }
+        } while (pointer < derBytes.length);
+        return (tlvs);
+    }
+
+    /*
+    This function is like walkDerTree but it doesn't store any tlvs. It can be used without gas cost
+    To compute the length of the tlv array (we could make a nodejs version of this in the future)
```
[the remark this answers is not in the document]
Westlad: It's because it doesn't change the state of the blockchain, thus it can be computed locally and costs no gas.

nightfall-deployer/contracts/X509.sol
Westlad: I was thinking that the functions each perform a similar loop so maybe they could be combined in future.  I'll remove it though because it's not needed.

nightfall-client/src/routes/kyc.mjs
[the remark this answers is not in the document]
israelboudoux: Yes, an error that will happen will be something like this `Property type can't be read from certificate`, then adding a validation for this would raise a more explicit message.

nightfall-deployer/contracts/DerParser.sol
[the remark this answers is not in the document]
signorecello: that's smart @Westlad, I only saw the "we could do this in nodejs" and wasn't seeing the point in not doing it now... but yeah solidity is much faster and this func runs locally :) great idea

config/default.js
```diff
@@ -57,6 +57,14 @@ module.exports = {
     { address: '0xfCb059A4dB5B961d3e48706fAC91a55Bad0035C9', groupId: 2 },
```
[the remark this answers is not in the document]
Westlad: Actually, I'll remove it. It's no longer needed.

nightfall-deployer/contracts/Whitelist.sol
```diff
@@ -3,28 +3,27 @@ import './Ownable.sol';
...
 pragma solidity ^0.8.0;
 
-contract KYC is Ownable {
-   
+contract Whitelist is Ownable {
     bool public whitelisting;
     mapping(address => uint256) public users;
     mapping(address => uint256) public managers;
 
     // groupIds cannot be zero, although we don't specifcally chack for this because assigning a groupId of zero has
```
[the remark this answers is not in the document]
Westlad: `Whitelist.md` explains `groupIds`.  They are a candidate for removal but I'll leave them for now; they just make it easier to tell which whitelist manager added the user. It's information that can also be gleaned from the transaction logs and currently there is only one whitelist manager in any case (the x509 certificate).

druiz0992 (approved):

signorecello (approved):
Amazing work, thank you!

PR #1199, opened 2022-11-07
Adding api authentication using an uuid

### LINKED ISSUE #1162: Add API keys for authentication
As proposer, I want to restrict some REST routes from the general public, such as:

- [ ] Make block

- [ ] Register proposer

- [ ] Update stake

- [ ] Etc

So an API authentication need to be defined.

- [ ] Find out which routes need to be protected

- [ ] Implement API authentication scheme using Passport or other state-of-the-art authentication

- [ ] Create an express middleware that uses it to protect some routes

signorecello:
This implements a way of protecting certain routes with an API key. This API key needs to be set on the `X-APP-TOKEN` header of every request. Without it, the routes will return a `401 unauthorized`.

Yes, it closes #1162 

Just normal tests. I added a test that sets invalid API keys just to see it fail spectacularly.

## Any other comments?
I added a method to `nf3` that basically sets the API keys to something. This method will be irrelevant once Nightfall Node is done, since `nf3` will no longer call the optimist.

### DIFF
--- a/nightfall-optimist/src/routes/proposer.mjs
+++ b/nightfall-optimist/src/routes/proposer.mjs
@@ -130,7 +130,7 @@ router.post('/update', auth, async (req, res, next) => {
 /**
  * Returns the current proposer
  */
-router.get('/current-proposer', auth, async (req, res, next) => {
+router.get('/current-proposer', async (req, res, next) => {
   try {
     const stateContractInstance = await getContractInstance(STATE_CONTRACT_NAME);
     const { thisAddress: currentProposer } = await stateContractInstance.methods
@@ -146,7 +146,7 @@ router.get('/current-proposer', auth, async (req, res, next) => {
 /**
  * Returns a list of the registered proposers
  */
-router.get('/proposers', auth, async (req, res, next) => {
+router.get('/proposers', async (req, res, next) => {
   try {
     const proposers = await getProposers();
 
--- a/nightfall-optimist/src/routes/proposer.mjs
+++ b/nightfall-optimist/src/routes/proposer.mjs
@@ -26,6 +26,7 @@ import {
 } from '../services/database.mjs';
 import transactionSubmittedEventHandler from '../event-handlers/transaction-submitted.mjs';
 import getProposers from '../services/proposer.mjs';
+import auth from '../utils/auth.mjs';
 
 const router = express.Router();
 const { TIMBER_HEIGHT, HASH_TYPE } = config;
@@ -42,7 +43,13 @@ export function setProposer(p) {
  * amount.  The user must post the address being registered.  This is for the
  * Optimist app to use for it to decide when to start proposing blocks.  It is * not part of the unsigned blockchain transaction that is returned.
  */
+<<<<<<< HEAD
 router.post('/register', async (req, res, next) => {
+=======
+router.post('/register', auth, async (req, res, next) => {
+  logger.debug({ msg: '/register endpoint', payload: JSON.stringify(req.body, null, 2) });
+
+>>>>>>> b0867d6a (feat: adding api authentication using an uuid)
   try {
     const { address, url = '', fee = 0 } = req.body;
     if (url === '') {
@@ -96,7 +103,13 @@ router.post('/register', async (req, res, next) => {
 /**
  * Function to update proposer's URL
  */
+<<<<<<< HEAD
 router.post('/update', async (req, res, next) => {
+=======
+router.post('/update', auth, async (req, res, next) => {
+  logger.debug({ msg: '/update endpoint', payload: JSON.stringify(req.body, null, 2) });
+
+>>>>>>> b0867d6a (feat: adding api authentication using an uuid)
   try {
     const { address, url = '', fee = 0 } = req.body;
     if (url === '') {
@@ -147,7 +160,13 @@ router.get('/proposers', async (req, res, next) => {
  * Function to return a raw transaction that de-registers a proposer.  This just
  * provides the tx data. The user has to call the blockchain client.
  */
+<<<<<<< HEAD
 router.post('/de-register', async (req, res, next) => {
+=======
+router.post('/de-register', auth, async (req, res, next) => {
+  logger.debug({ msg: '/de-register endpoint', payload: JSON.stringify(req.body, null, 2) });
+
+>>>>>>> b0867d6a (feat: adding api authentication using an uuid)
   try {
     const { address = '' } = req.body;
     const proposersContractInstance = await getContractInstance(PROPOSERS_CONTRACT_NAME);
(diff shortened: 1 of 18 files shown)


nightfall-optimist/src/routes/proposer.mjs
```diff
@@ -256,7 +268,9 @@ router.post('/payment', async (req, res, next) => {
  * for a parameterless function - therefore it's a constant and could be pre-
  * computed by the app that calls this endpoint.
  */
-router.get('/change', async (req, res, next) => {
+router.get('/change', auth, async (req, res, next) => {
```
druiz0992, line 271: I think this function doesn’t need to be protected. Any one can call changeProposer

test/e2e/protocol/proposer.test.mjs
```diff
@@ -127,6 +127,20 @@ describe('Basic Proposer tests', () => {
     }
   });
 
+  it('should fail to register a proposer without a valid API key', async () => {
```
druiz0992, line 130: How do you test that an end user without the right api key can’t call a private endpoint?

druiz0992 (changes requested):
Asked some questions

nightfall-optimist/src/routes/proposer.mjs
signorecello: I thought about that. On the other hand, I wonder if there's any situation where the proposer may not want to call changeProposer, for example to save on gas costs. I'm happy to open this route to the public though, if it makes more sense to you

test/e2e/protocol/proposer.test.mjs
signorecello: You mean another test with no token at all? That makes sense, yes. I'll add it tomorrow

signorecello: Actually, if we want to be comprehensive, this test accounts for:

- You can't provide a wrong API key and access a private endpoint
- You must provide a correct API key to access a private endpoint

It should also account for:

- You can't access a private endpoint with an inexistent key
- You can access a public endpoint with, without, or with an invalid API key

Not sure if I'm covering all cases here, though

pawelgrzybek:
Does this passport strategy do anything more than just checking for a presence of a correct token on a header? If its as simple as that maybe we can drop it. Passport shines when you combine multiple strategies, or use more complicated ones, but this one is not more complicated than `req.headers.token === env.token ?  next() : return 401`. Maybe it does something extra that I am unaware of?

nightfall-optimist/src/routes/proposer.mjs
druiz0992: @daveroga please, comment on this

signorecello:
Yes you're absolutely right. It could be just a simple comparison. However, I think using passport will pave the way to a later proposer UI we may want to develop, maybe we want to use multiple strategies and etc. 

I'm happy not using passport at all, as less code = 🤩

nightfall-optimist/src/routes/proposer.mjs
signorecello: So I just submitted a new commit. As we're going for a 1:1 implementation, I simply opened the three routes that are currently being used by the proposer app:

```
/offchain-transaction
/mempool
/:contract
```

Everything else should have the `auth` middleware, for consistency's sake. I'm happy to open other routes if @daveroga thinks it's worth it, some of them make perfect sense to be open (like `getCurrentProposer` and etc)

test/e2e/protocol/proposer.test.mjs
signorecello: Just submitted these missing tests

nightfall-optimist/src/routes/proposer.mjs
daveroga: `/fee` also was added for the block explorer and should be available for the user to know the fees

daveroga: `changeCurrentProposer` could be called by anyone, although the interested one is the next proposer. There is no issue in someone to call this. If the current proposer is still in the challenge period it will revert and if not it will change to the next proposer.

test/e2e/protocol/proposer.test.mjs
```diff
@@ -127,6 +127,37 @@ describe('Basic Proposer tests', () => {
...
+      bootProposer.setApiKey(environment.AUTH_TOKEN);
+    }
+  });
+
+  it('should access any public route with any API key', async () => {
+    bootProposer.resetApiKey();
+    await bootProposer.unprocessedTransactionCount();
+    bootProposer.setApiKey(environment.AUTH_TOKEN);
+    await bootProposer.unprocessedTransactionCount();
+  });
```
imagobea, line 159: Is this test to demonstrate that unprotected routes are available both with and without API key? If so I think these are actually 2 tests - one per scenario.

Specially since right/wrong API key seems that won't matter as the middleware is not present.

Also, I guess we are relying on Promises being resolved, but IMO we should try to write assertions. Here you could assert against status code `200` for example.

imagobea (changes requested):
Good stuff @signorecello specially since you put this together in the blink of an eye! (Apololgies that I took so long to review..)

Requesting changes:

- Please align your branch and review the proposer file
- Remove unnecessary packages
- I made a few other suggestions but these can be re-visited later since we are merging into feat-branch

Many thanks!! :nerd_face:

nightfall-optimist/src/routes/proposer.mjs
```diff
@@ -42,7 +43,9 @@ export function setProposer(p) {
  * amount.  The user must post the address being registered.  This is for the
  * Optimist app to use for it to decide when to start proposing blocks.  It is * not part of the unsigned blockchain transaction that is returned.
  */
-router.post('/register', async (req, res, next) => {
+router.post('/register', auth, async (req, res, next) => {
+  logger.debug({ msg: '/register endpoint', payload: JSON.stringify(req.body, null, 2) });
```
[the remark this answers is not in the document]
signorecello: It is, I'll remove this one and the others as well

nightfall-optimist/src/utils/auth.mjs
```diff
@@ -0,0 +1,5 @@
+export default function auth(req, res, next) {
+  const token = req.headers['x-app-token'];
+  if (token === process.env.AUTH_TOKEN) return next();
+  return res.status(401).send('Unauthorized');
+}
```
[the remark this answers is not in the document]
signorecello: I discussed this with @israelboudoux , and in principle I also agree to avoid duplication. However, in this case there would be no reuse for it. As of now this is the only service using API auth so maybe we could leave that change if we ever implement the same for other services

nightfall-optimist/src/utils/auth.mjs
```diff
@@ -0,0 +1,11 @@
+import config from 'config';
+import crypto from 'crypto';
```
velkoskis, line 2: Smooth and short implementation w/o passport.js. I'm glad we went for the hashing option. Nice :)

velkoskis (approved):
Great job Ze. 🙌 
Did a local run on the tests and they pass. After merging this swagger docs we will have to be updated with the API key.

druiz0992 (approved):

PR #1200, opened 2022-11-08
feat: node can sign and send txs including making blocks (mvp)

### LINKED ISSUE #1163: Implement an MVP - One proposer route
As a proposer, I want to be sure the work is done without breaking changes. For that, I need to provide a simple MVP of the new changes.

- [x] Instead of sending the result of `encodeABI()` for the `registerProposer` call to the blockchain, sign it directly on the optimist

- [x] Remove the signing from the `nf3` file

To cover min functionality as described in #1111. Investigate how the tx queue/propose block processes work/can be achieved.

imagobea:

- Allows optimist to operate autonomously as a proposer, without the need of additional services/websockets
- Autonomy is achieved by signing and sending txs directly from optimist, using existing env var `PROPOSER_KEY`
- Updated: routes `register`, `/de-register`, `/update`, `change`; services `conditionalMakeBlock`
- With the above changes, the "node" continues to produce blocks when the time is right (`TX_PER_BLOCK`)
- The "node" can also produce blocks any time via `/blck/make-now`

**You will find commented code which can be deleted if we all agree**. This is especially true for `nf3`, where endpoint calls remain but the rest of the logic has been disabled. This is to keep tests compatibility.

**Tests have also been updated**. Although the approach has been a bit more radical and a few tests have been removed/updated completely.

**We have removed second/third proposers in protocol/administrator tests**. These did not spin multiple optimists so we agreed that these were not a true representation of Nightfall.

**At the time of opening this PR some actions are failing** (`adversary`, `optimist-syn` and `ping-pong`). A few tests have been marked with `skip`.

Also note that the whole create tx > sign tx > send tx flow can be refactored to avoid duplication. **Current approach is to `await` for everything** (txs and db operations), but we can try to handle queues on a separate PR.

Closes #1163 

Run Nightfall

```
bin/start-nightfall
```

Interact with optimist from the command line, examples

```
curl http://localhost:8081/proposer/current-proposer
curl http://localhost:8081/proposer/proposers
curl -d "url=pepe&stake=999999999" http://localhost:8081/proposer/register
curl -d "" http://localhost:8081/proposer/de-register
curl http://localhost:8081/block/make-now
```

You can also run the tests as per the commands in the `package.json`.

To send transactions you can use the SDK (mind the env vars)

```
npm run eg:ganache:deposit
```

## Any other comments?

n/a


nightfall-optimist/src/routes/proposer.mjs
```diff
@@ -43,32 +45,58 @@ export function setProposer(p) {
...
     /*
       when we get to here, either the proposer was already registered (txDataToSign === '')
       or we're just about to register them. We may or may not be registed locally
       with optimist though. Let's check and fix that if needed.
      */
-    if (!(await isRegisteredProposerAddressMine(address))) {
+    if (!(await isRegisteredProposerAddressMine(ethAddress))) {
       logger.debug('Registering proposer locally');
-      await setRegisteredProposerAddress(address, url); // save the registration address
+      await setRegisteredProposerAddress(ethAddress, url); // save the registration address
```
imagobea, line 99: NB: Pretty sure the code beyond this line (in this route) can be deleted

signorecello:
Hey @imagobea just a question about multiple optimists. I also do agree these multiple optimists tests should be done on a CI/CD repo only and should be removed from here. Just wondering if @daveroga agrees, as he's the mentor of the PoS functionality. If he does, I do so too 😊

In my view, to do this on a CI/CD repo would massively increase the reliability of the services, as optimists could be spinned with their own separate mongos as well

imagobea:
> Hey @imagobea just a question about multiple optimists. I also do agree these multiple optimists tests should be done on a CI/CD repo only and should be removed from here. Just wondering if @daveroga agrees, as he's the mentor of the PoS functionality. If he does, I do so too blush
> 
> In my view, to do this on a CI/CD repo would massively increase the reliability of the services, as optimists could be spinned with their own separate mongos as well

Hello @signorecello , yes protocol-proposer/administrator were discussed with @daveroga and he agreed. On the other hand now ping-pong test I think introduces this concept of many optimists. But ping-pong will have to be updated according to the new "nighfall-node" approach anyways. Let's discuss also with @israelboudoux next week. Cheers!

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -116,46 +125,56 @@ export async function conditionalMakeBlock(proposer) {
...
           )
           .encodeABI();
+        const { address } = web3.eth.accounts.privateKeyToAccount(ethPrivateKey);
+        const blockStake = await stateContractInstance.methods.getBlockStake().call();
+        const tx = {
+          from: address,
+          to: stateContractAddress,
+          data: txDataToSign,
+          value: blockStake,
+          gas: 8000000,
```
druiz0992, line 141: gas cant be a fixed number in final version

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -116,46 +125,56 @@ export async function conditionalMakeBlock(proposer) {
...
+        };
+        const signedTx = await web3.eth.accounts.signTransaction(tx, ethPrivateKey);
```
druiz0992, line 143: i know you mention this on the description, but final version must ensure that multiple L2 blocks can fit in the same L1 block.

imagobea: Noted :memo: Thanks for taking a look

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -116,46 +125,56 @@ export async function conditionalMakeBlock(proposer) {
...
+        const receipt = await web3.eth.sendSignedTransaction(signedTx.rawTransaction);
+        logger.debug(`Transaction receipt ${receipt}`);
```
signorecello, line 145: ❤️

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -69,11 +76,13 @@ async function makeBlock(proposer, number = TRANSACTIONS_PER_BLOCK) {
...
 /**
  * This function will make a block iff I am the proposer and there are enough
- * transactions in the database to assembke a block from. It loops until told to
+ * transactions in the database to assemble a block from. It loops until told to
  * stop making blocks. It is called from the 'main()' routine to start it, and
  * should not be called from anywhere else because we only want one instance ever
  */
 export async function conditionalMakeBlock(proposer) {
+  const stateContractInstance = await waitForContract(STATE_CONTRACT_NAME);
+  const stateContractAddress = await getContractAddress(STATE_CONTRACT_NAME);
```
signorecello, line 85: Just for later reviewing, all this pattern of getting stateContractInstances is happening all over the place... Wondering if we could abstract it somewhere and simply return an object like:

```
{
    "state": "0x1234",
    "proposers", "0x5678"
}
````

We could then call this func somewhere in the ```app.mjs``` file and reuse it.

For later, though :)

signorecello (approved):

velkoskis (approved):
Good job with the implementation as well as the PR review instructions 👏 . All the endpoints are tested and working good, including make-block. 
Sorry for reviewing this later than it should have been reviewed. 
Thank you for your patience. 🙏

imagobea:
> Good job with the implementation as well as the PR review instructions clap . All the endpoints are tested and working good, including make-block. Sorry for reviewing this later than it should have been reviewed. Thank you for your patience. pray

Thank you for taking the time Stefan, cheers

PR #1203, opened 2022-11-08
wallet withdraw tx type

druiz0992:
additional location where wallet uses tx type 3 for withdrawal instead of 2

### DIFF
--- a/wallet/src/components/Transactions/index.jsx
+++ b/wallet/src/components/Transactions/index.jsx
@@ -365,7 +365,7 @@ const Transactions = () => {
                     </a>
                   </div>
                 </Row>
-                {tx.txType === '3' && tx.isOnChain > 0 && !tx.withdrawState ? (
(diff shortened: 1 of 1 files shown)


pawelgrzybek (approved):

wallet/src/components/Transactions/index.jsx
```diff
@@ -365,7 +365,7 @@ const Transactions = () => {
                     </a>
                   </div>
                 </Row>
-                {tx.txType === '3' && tx.isOnChain > 0 && !tx.withdrawState ? (
+                {tx.txType === '2' && tx.isOnChain > 0 && !tx.withdrawState ? (
```
israelboudoux, line 368: Could we use constants or even methods (e.g. `tx.isWithdraw()`) instead of comparing with scalar values?

israelboudoux (changes requested):

israelboudoux (approved):
Don't mind, this isn't the right moment for this change I've asked for.

PR #1184, opened 2022-11-03
 make adversary proposer to submit bad transaction

LijuJoseJJ:
## What does this implement/fix? Explain your changes.
before this PR adversary proposer never use to submit bad transaction it create rather just put bad tx is block this made block propose fail because of transactionHash isEscrowed check from shiled contract.

## Does this close any currently open issues?
fixes #https://github.com/EYBlockchain/nightfall_3_private/issues/292.

## What commands can I run to test the change? 
run Adversary test

## Any other comments?
temporary workflow ran on changes https://github.com/EYBlockchain/nightfall_3/actions/runs/3384741762/jobs/5623106944
another workflow ran on this branch https://github.com/EYBlockchain/nightfall_3/actions/runs/3387286233/jobs/5627830238

PR #1208, opened 2022-11-09
make adversary wait and count required no of rollbacks

LijuJoseJJ:
This PR change adversary.test.mjs to wait for certain no. of rollback before checking for expected balance

this PR also comment IncorrectHistoricRoot type bad transaction creation as block propose is failing.

this Pr also fix two bugs introduced by my PR https://github.com/EYBlockchain/nightfall_3/pull/1184 bug could catch local and github action because adversary.test was not waiting for required no. of rollbacks.
bug was `IncorrectProofDeposit` and `IncorrectPublicInputDepositCommitment` bad tx creation logic was not wait for tx to reach and get store in db before adversary could propose a block.

PR also moved submit bad tx (signing and submittx) from clii to transpiler.

fixes # https://github.com/EYBlockchain/nightfall_3_private/issues/248


cli/lib/nf3.mjs
```diff
@@ -1075,36 +1075,6 @@ class Nf3 {
 
       if (type === 'rollback') proposeEmitter.emit('rollback', data);
 
-      // this is used by adversary proposer for submitting bad transaction.
```
ChaitanyaKonda, line 1078: Thank you for removing it here and making it part of transpiled code.

test/adversary.test.mjs
```diff
@@ -56,6 +62,8 @@ describe('Testing with an adversary', () => {
   // const value = 10;
   // const value1 = 1000;
   const value2 = 5;
+  let rollbackCount = 0;
+  const minExpectedRollbacks = 6;
```
ChaitanyaKonda, line 66: In a new issue please make `minExpectedRollbacks` equal to the number of bad transaction types that exist by configuring the adversary test to run good transaction(s) for each bad transaction such that there is a rollbacked block for each bad transaction type. This ensures testing each type of bad transaction.

ChaitanyaKonda (approved):

test/adversary.test.mjs
LijuJoseJJ: understood, I will make the changes in next PR

RogerTaule (approved):

PR #1213, opened 2022-11-11
Nightfallnode/swagger js docs

luizoamorim:
Add openapi 3.0.0 documentation for proposer and block routes.

Not yet. We still need to finalize more 5 functions in the other files.

The documentation is generated automatically based on the openapi jsDoc above each route.


nightfall-optimist/src/routes/block.mjs
```diff
@@ -16,6 +13,133 @@ import { setMakeNow } from '../services/block-assembler.mjs';
 
 const router = express.Router();
 
+/**
```
israelboudoux, line 16: Hey friends, I'm not sure if this is the best way, but having such big block of comments inside the source-code doesn't look good to me. Maybe, moving them to a YML file apart from the source-code it is better. Having this separated might be easier to maintain. We should discuss that with the team, maybe I'm wrong and this is the right of of doing it for JS apps.

israelboudoux (changes requested):
We have to discuss with the team the best approach for adding these swagger docs.

nightfall-optimist/src/app.mjs
```diff
@@ -12,6 +16,49 @@ import {
...
+            },
+          },
+        },
+      },
+    },
+  },
+  apis: ['src/routes/**/*.mjs'],
+};
+
+const spec = swaggerJsDoc(options);
```
imagobea, line 60: We could try to move this outside `app.mjs` and simply export/import `spec` from the new file into this one

nightfall-optimist/src/routes/block.mjs
imagobea: Like we talked yesterday, they implemented these docs first on a separate file - like you suggest, and people seemed to complain. I don't think we're happy with the result... Let's decide later. Since we're working on a feature branch we have some time to reflect/research/decide.
:pray:

nightfall-optimist/src/routes/proposer.mjs
```diff
@@ -370,7 +702,20 @@ router.post('/encode', async (req, res, next) => {
...
+ *   post:
+ *     tags:
+ *     - Proposer
+ *     summary: Offchain Transaction.
+ *     description: Offchain transaction executed by a client.
+ *     responses:
+ *       200:
+ *         description: OK
+ *       500:
+ *         description: An error occured
```
imagobea, line 717: :nerd_face: Please update all instances

```suggestion
 *         description: An error occurred
```

imagobea (changes requested):
image

Blocking since I couldn't reach the docs. Perhaps I'm being dumb, but you could have easily provided some instructions on the PR "instructions for testing" section. PR descriptions exist to help your reviewers to understand and test your changes. This should also inform future developers to understand what happened on a particular PR.

This is what I did:

```
npm ci && bin/start-nightfall -g
```

Then visited URL as per image, ie optimist url + `/api-docs`.
Please let me know what I'm doing wrong. Thank you.

nightfall-optimist/src/routes/block.mjs
```diff
@@ -36,12 +185,67 @@ router.get('/make-now', async (req, res, next) => {
...
+ *                      transactions:
+ *                        type: array
+ *                        example: null
+ *                      index:
+ *                        type: number
+ *                        example: null
+ *        401:
+ *          $ref: '#/components/responses/UnauthorizedError'
+ *        500:
+ *          description: Some error ocurred.
```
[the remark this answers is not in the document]
velkoskis: Luiz did it one way, I did another way, none of them correctly 😄  . Thank you for noticing these things, sometimes they get away.  🙏

velkoskis (approved):

velkoskis (approved):
LGTM!

PR #1214, opened 2022-11-11
Fix state unit test

eugypalu:
Uses the function `getMinimumStake` and `getValuePerSlot` so as not to cause problems on any changes to the configuration values in `config.sol`

npm run unit-test-state


RogerTaule (approved):
LGTM. Unit tests went through

ChaitanyaKonda (approved):

PR #1218, opened 2022-11-14
Mutliple adversary fixes

ChaitanyaKonda:
Rollback should retrieve transactions of an L2 block by its L2 block number and not just transaction hash because it is not unique in the event of duplicate transactions.

Today, transactions from mempool are removed after creating an L2 block during block assembly and receiving an L2 block as part of L2 block proposed event. It is important to note that commitments and nullifiers can have duplicates in mempool that have not been proposed yet, these transactions should also be removed from mempool. The PR fixes this.

Refactor check-block code.

Challenger might catch wrong proof error when the testing requires catching of duplicate nullifier/commitment. As long as the bad transaction is caught, we do not care which of the two reasons it is caught under. 

239, 242, 243, 247 

Github action

### DIFF
--- a/nightfall-optimist/src/services/block-assembler.mjs
+++ b/nightfall-optimist/src/services/block-assembler.mjs
@@ -11,6 +11,8 @@ import constants from '@polygon-nightfall/common-files/constants/index.mjs';
 import { waitForContract } from '@polygon-nightfall/common-files/utils/contract.mjs';
 import {
   removeTransactionsFromMemPool,
+  removeCommitmentsFromMemPool,
+  removeNullifiersFromMemPool,
   getMostProfitableTransactions,
   numberOfUnprocessedTransactions,
 } from './database.mjs';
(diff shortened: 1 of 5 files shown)


nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -159,6 +161,10 @@ export async function conditionalMakeBlock(proposer) {
         // remove the transactions from the mempool so we don't keep making new
         // blocks with them
         await removeTransactionsFromMemPool(block.transactionHashes);
+        await removeCommitmentsFromMemPool(
+          transactions.map(transaction => transaction.commitments),
+        );
+        await removeNullifiersFromMemPool(transactions.map(transaction => transaction.commitments));
```
RogerTaule, line 167: I think this is wrong and it should be : 
```suggestion
        await removeNullifiersFromMemPool(transactions.map(transaction => transaction.nullifiers));
```

RogerTaule (changes requested):

RogerTaule (approved):

LijuJoseJJ (approved):

PR #1220, opened 2022-11-14
Daveroga/fix multiproposer ping pong

daveroga:
Fix accounts and control some checks in case other proposers running while running the ping-pong test. 

No

npm run ping-pong

## Any other comments?
No


docker/docker-compose.proposers.yml
```diff
@@ -115,8 +111,8 @@ services:
       CLIENT_HOST: client
       CLIENT_PORT: 8080
       PROPOSER_HOST: proposer_1
-      PROPOSER_PORT: 8092
-      PROPOSER_KEY: '0x4775af73d6dc84a0ae76f8726bda4b9ecf187c377229cb39e1afa7a18236a69d'
+      PROPOSER_PORT: 8093
+      PROPOSER_KEY: '0xcbbf1d0686738a444cf9f66fdc96289035c384c4e8d26768f94fa81f3ab6596a'
```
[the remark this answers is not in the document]
daveroga: This is because the infra was using already the proposer1 key for the existing proposer and I'd like the test to work also in the infra.

doc/pos.md
```diff
@@ -6,6 +6,9 @@ and Tendermint's proposer selection algorithm, Nightfall will select next propos
 Terminology:
 
 - **Slots**. Units assigned to proposers based on their stake.
+- **Value per slot**. Amount of the stake for each slot to calculate the slots that a proposer will
+  have based on their stake. If a proposer has 10K staked and the value per slot is 1K then the
+  proposer will have 10 slots to calculate their stake power in the weighted round robin algorithm.
```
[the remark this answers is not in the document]
daveroga: Yes. I'll explain it.

bin/export-test-env
```diff
@@ -0,0 +1,6 @@
+#! /bin/bash
```
druiz0992, line 1: I would provide a more descriptive name than mere `test`. Maybe `multiproposer-test` or something like that

bin/start-test-env
```diff
@@ -0,0 +1,82 @@
+#! /bin/bash
```
druiz0992, line 1: more descriptive name

docker/docker-compose.tests.yml
```diff
@@ -0,0 +1,350 @@
+version: '3.5'
```
druiz0992, line 1: better name

example.test.env
```diff
@@ -0,0 +1,7 @@
+CLIENT1_API_URL=http://localhost:8083
```
druiz0992, line 1: better name

test/ping-pong/ping-pong.test.mjs
```diff
@@ -1,156 +1,324 @@
...
 const environment = config.ENVIRONMENTS[process.env.ENVIRONMENT] || config.ENVIRONMENTS.localhost;
+const { TEST_ERC20_ADDRESS } = process.env;
 
-let result;
+const nf3Users = [];
+let blockStake;
+const listAddresses = [];
+const listTransfersTotal = [];
+const TEST_LENGTH = 4;
+const value = 10;
```
Westlad, line 23: Should this be a random sequence, rather than a single value, to avoid the issue of a false pass if both users fail to send?

daveroga: This is only used for the initial deposits to be included in a block. The first part of the test creates TEST_LENGTH deposits of 10 in this case. Then for every loop of the TEST_LENGTH we do 1 transfer and 1 deposit of a random value between 1 and 9 in this case. 
```
// Returns a random integer from 1 to value - fee
const valueToTransfer = Math.floor(Math.random() * (value - fee)) + 1; 
```
I do like this because I defined 1 for the fee. So then with the first deposits included in a block, every user will have enough to cover the randoms values + 1 of the fee and at maximum it will be 10 every loop.

druiz0992 (approved):
waiting for @Westlad resolution, but im fine with the code as it is

PR #1229, opened 2022-11-17
(adversary) make all bad tx to be created and challenge and rollback individually 

LijuJoseJJ:
this PR fix adversary code it add validTransaction after each bad tx which make rollback of individual bad transaction


LijuJoseJJ:
PR is outdated fix are in master already

PR #1246, opened 2022-11-21
ChangeCurrentProposer - calculation of the next proposers

### LINKED ISSUE #1223: changeCurrentProposer - Calculation of the next proposers
Instead of leaving it to the `changeCurrentProposer` each time to calculate from scratch the `nextProposer`, the next n are calculated all at once, so that we know in advance what the next ones will be, and stimulate to execute the function for changing proposer

- [x] Smart contract functions

- [ ] gas cost comparison

- [ ] Test

eugypalu:
Instead of leaving it to the `changeCurrentProposer` each time to calculate from scratch the `nextProposer`, the next n are calculated all at once, so that we know in advance what the next ones will be, and stimulate to execute the function for changing proposer

- [x]  Smart contract functions
- [x]  gas comparison
- [x]  Test

### Gas comparison:
Gas comparison is made on two functions modified in this pr (`setNumProposers`, `changeCurrentProposer`).

#### setNumProposers(np)

| | Previous Version | New Version |
| :---         |     :---:      |          ---: |
| np != 2   | 33208  | 33246   |
| np == 2    | 36008       | 346445      |

In this case, the new version is more expensive in case `numproposers` = 2 because it is calculated the list of next proposers.

The first time the setNumProposers is executed with np == 2 and the changeCurrentProposer is called the cost in gas is:
| min | max | avg |
| :---         |     :---:      |          ---: |
| 637981  | 1047968  | 759665   |

#### changeCurrentProposer

| | Previous Version | New Version |
| :---         |     :---:      |          ---: |
| currentSprint != sprintsInSpan   | 315195 avg  | 118882 avg   |
| currentSprint == sprintsInSpan    | 315195 avg       | 325525 avg     |

Unlike the previous version, the `changeCurrentProposer` does not calculate the next proposer from scratch, but only calculates it when `currentsprint == sprintsInSpan`. so the cost is much lower in the case where `currentsprint != sprintsInSpan`, since the next proposer is simply taken from the list, while in the case where `currentsprint == sprintsInSpan` the cost is about the same as before since the whole list is calculated.

Gas cost was also evaluated with 10/20/30/50 proposers, but the number does not impact gas cost

close #1223 
npm run unit-test-state

### DIFF
--- a/nightfall-deployer/contracts/State.sol
+++ b/nightfall-deployer/contracts/State.sol
@@ -36,7 +36,7 @@ contract State is ReentrancyGuardUpgradeable, Pausable, Key_Registry, Config {
     address[] public slots; // slots based on proposers stake
     ProposerSet[] public proposersSet; // proposer set for next span
     uint256 public currentSprint; // the current sprint of the span
-    address[] public proposersList;
+    address[] public spanProposersList;
 
     function initialize() public override(Pausable, Key_Registry, Config) {
         Pausable.initialize();
@@ -543,7 +543,7 @@ contract State is ReentrancyGuardUpgradeable, Pausable, Key_Registry, Config {
             currentSprint = 0; // We don't have sprints because always same proposer
             addressBestPeer = currentProposer.thisAddress;
         } else {
-            addressBestPeer = proposersList[currentSprint];
+            addressBestPeer = spanProposersList[currentSprint];
             currentSprint += 1;
 
             if (currentSprint == sprintsInSpan) {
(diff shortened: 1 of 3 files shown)


daveroga (approved):

nightfall-deployer/contracts/State.sol
```diff
@@ -365,7 +366,12 @@ contract State is ReentrancyGuardUpgradeable, Pausable, Key_Registry, Config {
     }
 
     function setNumProposers(uint256 np) public onlyProposer {
+        uint256 prevNumProposers = numProposers;
         numProposers = np;
+        if (numProposers == 2 && prevNumProposers == 1) {
```
Westlad, line 371: What is the significance of numProposers being `==2`?

eugypalu: when changing from 1 proposer to 2 proposers is the situation where it becomes necessary to build the `spanProposersList`, to allow the `changecurrentproposer` to directly access the next proposers from the list.
We only handle the specific case that goes from 1 to 2 proposers, because all other cases are handled by the `changeCurrentProposer` function

Westlad: Got it, thanks!

Westlad (approved):

PR #1247, opened 2022-11-21
Historic root challenge

IlyasRidhuan:
Expands the historicRootBlockNumber challenge to also handle the case where the number of historic roots specified is greater than the number of nullifiers.

This doesn't really represent an attack vector, it's just for completeness
Run the adversary test or the contract unit tests.

### DIFF
--- a/nightfall-deployer/contracts/Challenges.sol
+++ b/nightfall-deployer/contracts/Challenges.sol
@@ -268,13 +268,33 @@ contract Challenges is Stateful, Config {
             transaction.transactionSiblingPath
         );
 
-        for (uint256 i = 0; i < transaction.transaction.nullifiers.length; ++i) {
-            uint64 historicblockNumberL2 = Utils.getHistoricRoot(
+        // Unpack the historic block numbers, they are packed in 256bits and so 4 x 64 historicRootBlockNumbers are unpacked.
+        uint64[] memory historicRootBlockNumbers = new uint64[](
+            transaction.transaction.historicRootBlockNumberL2.length * 4
+        );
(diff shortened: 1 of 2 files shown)


RogerTaule (approved):

nightfall-deployer/contracts/Challenges.sol
```diff
@@ -268,13 +268,33 @@ contract Challenges is Stateful, Config {
...
+                return;
+            }
+        }
 
-            if (uint256(historicblockNumberL2) >= state.getNumberOfL2Blocks()) {
+        // Historic Block Numbers are packed in 256bits and so always results in multiples of 4
+        // if nullifiers are not multiple of 4 we need to ensure these "extra" historic block numbers are zero.
+        for (
+            uint256 i = historicRootBlockNumbers.length;
+            i < transaction.transaction.nullifiers.length;
```
daveroga, line 294: I think this is probably wrong. I would expect `i > transaction.transaction.nullifiers.length`.  So `>` instead of `<`.

daveroga (changes requested):

nightfall-deployer/contracts/Challenges.sol
IlyasRidhuan: Good catch! Yep it should be the other way - I'll push up a fix

IlyasRidhuan: I think I also need to be `i = historicRootBlockNumbers.length -1` in the first line

daveroga (approved):

Issue #1152, opened 2022-10-26
L2 transfrom functionality


siosw:
this is part of the L2 tokenisation effort tracked here #1009

### What is this?
We want to allow users to combine L2 commitments into different, new L2 commitments according to rules defined by the user. This is needed to represent IRL supply chain processes within nightfall.

This will require a new service within client as well as a new circuit.

PR #1252, opened 2022-11-22
transform service

### LINKED ISSUE #1009: L2 Tokenisation Task List
Here is the task list that defines what needs to completed as part of L2 Tokenisation. Each should form their own issues with corresponding PR. They are broadly grouped by their contribution to the end goal.

### Performance
- [ ] Migration to snarkjs backend #985 
- [ ] Enable rapidsnark as an option for server deployments.

### Flexibility
- [ ] Dynamic Commitment + Nullifier Arrays #986 
- [ ] Remove `TRANSACTION_TYPE` enum and replace `Key_Registry` to use a content-based hash #988
- [ ] Convert maximum transactions in `proposeBlock` to instead work on the `num_bytes` of the block + transactions.
- [ ] Metadata Array in On-chain Transaction Structure

### Usability
- [ ] Transfer/Withdraw that can use specific commitment hash
- [ ] Private Mint / Burn Zokrates circuit

siosw:
Adds a transform transaction, combine N L2 tokens into M L2 tokens.

closes  #1009 and #1152
 
`npx hardhat test test/unit/circuits/transform.circom.mjs --no-compile`

`npx hardhat test test/e2e/tokens/l2tokenisation.test.mjs --grep 'Transform' --no-compile`


druiz0992 (approved):

nightfall-client/src/services/transform.mjs
```diff
@@ -0,0 +1,234 @@
...
+  const { rootKey, compressedZkpPublicKey, fee } = generalise(items);
+
+  const shieldContractInstance = await waitForContract(SHIELD_CONTRACT_NAME);
+  const maticAddress = generalise(
+    (await shieldContractInstance.methods.getMaticAddress().call()).toLowerCase(),
+  );
+
+  const circuitHash = await getCircuitHash('transform');
+
+  // the last two nullifiers are reserverd for covering the fee
```
IlyasRidhuan, line 38: im not a big fan of this logic.

Reserving nullifiers and commitments to be used as part of fees leads to unnecessary rigidity, reduces anonymity and means that on-chain transactions will be padded such that non-zero hashes are between 0-hashes.

IlyasRidhuan: If the transaction takes up to N nullifiers and C commitments, then we should allow for:

- Up to (N-1) nullifiers related to the fee and up to 1 commitment related to the change or;
- Up to  N nullifiers related to the token transformation and up to C commitments related to the token change.

in my mind the number of commitments, C, should be at least N + 1 (maybe even N + 2) to account for a partial transformation of input commitments and the fee.

siosw: > in my mind the number of commitments, C, should be at least N + 1 (maybe even N + 2) to account for a partial transformation of input commitments and the fee.
I agree with this, right now this is set to C=5 N=6 but N=7 seems reasonable to me as well.

My reason for reserving the last two commitments for the fee were to match the burn and tokenise transaction. 
Another concern I had was that checking if a commitment is a fee or not on the circuit level would increase the number of constraints.

Depending on your thoughts on this I'm happy to change it to be completely agnostic to the order of commitments / nullifiers.

nightfall-client/src/routes/transform.mjs
```diff
@@ -0,0 +1,20 @@
+import express from 'express';
+import logger from '@polygon-nightfall/common-files/utils/logger.mjs';
+import transform from '../services/transform.mjs';
+
+const router = express.Router();
+
+router.post('/', async (req, res, next) => {
+  try {
+    const { rawTransaction: txDataToSign, transaction } = await transform(req.body);
+    logger.debug('returning raw transaction');
```
israelboudoux, line 10: You don't need these log instructions anymore, on the routes the logging of the payload/response is carried out automatically.

nightfall-client/src/routes/transform.mjs
```diff
@@ -0,0 +1,20 @@
...
+    logger.trace(`raw transaction is ${JSON.stringify(txDataToSign, null, 2)}`);
+    // convert commitment from GN to hex form for transmission
+    res.json({ txDataToSign, transaction });
+  } catch (err) {
+    logger.error(err);
```
israelboudoux, line 15: This `logger.error` can be dropped out, the error will caught by the error middleware and will perform `logger.error`.

nightfall-client/src/queues/transform.mjs
```diff
@@ -0,0 +1,31 @@
...
+      data: null,
+    };
+
+    try {
+      logger.debug(`Transform queue handler ${message.content.toString()}`);
+
+      const txDataToSign = await transform(JSON.parse(message.content.toString()));
+      logger.debug({
+        msg: 'Returning raw transaction',
+        rawTransaction: JSON.stringify(txDataToSign, null, 2),
```
israelboudoux, line 19: The `JSON.stringify` isn't needed, you can pass the object directly, it is properly handled by the log lib.

nightfall-client/src/services/transform.mjs
```diff
@@ -0,0 +1,234 @@
...
+    return submitTransaction(
+      optimisticTransformTransaction,
+      commitmentInfo,
+      rootKey,
+      shieldContractInstance,
+      true, // offchain
+    );
+  } catch (error) {
+    await Promise.all(commitmentInfo.oldCommitments.map(o => clearPending(o)));
+    throw new Error(error);
```
israelboudoux, line 230: You can just re-throw the error.

nightfall-client/src/services/transform.mjs
```diff
@@ -0,0 +1,234 @@
...
+      nullifiers: commitmentInfo.nullifiers,
+      proof,
+      numberNullifiers: VK_IDS.transform.numberNullifiers,
+      numberCommitments: VK_IDS.transform.numberCommitments,
+      isOnlyL2: true,
+    });
+
+    logger.debug({
+      msg: 'Client made transaction',
+      transaction: JSON.stringify(optimisticTransformTransaction, null, 2),
```
israelboudoux, line 218: No need to use `JSON.stringify` anymore.

nightfall-client/src/services/transform.mjs
```diff
@@ -0,0 +1,234 @@
...
+      msg: 'witness input is',
+      witness: JSON.stringify(witness, 0, 2),
+    });
+
+    // call a worker to generate the proof
+    const res = await generateProof({ folderpath: 'transform', witness });
+
+    logger.debug({
+      msg: 'Received response from generate-proof',
+      response: JSON.stringify(res.data, null, 2),
```
israelboudoux, line 198: No need to use `JSON.stringify` anymore.

nightfall-client/src/services/transform.mjs
```diff
@@ -0,0 +1,234 @@
...
+      privateData,
+      commitmentInfo.roots,
+      maticAddress,
+      VK_IDS.transform.numberNullifiers,
+      VK_IDS.transform.numberCommitments,
+    );
+
+    logger.debug({
+      msg: 'witness input is',
+      witness: JSON.stringify(witness, 0, 2),
```
israelboudoux, line 190: No need to use `JSON.stringify` anymore.

IlyasRidhuan (approved):

PR #1255, opened 2022-11-22
chore: remove hashing function

velkoskis:
This removes the existing hash function from the auth logic . It serves the purpose of simplifying and using an .env variable as the auth token for the proposer routes.
no
npm run test-e2e-protocol            
## Any other comments?
It is suggested for the deployer to use UUID(https://www.uuidgenerator.net/) as the authentication token. However, the code doesn't stop the deployer to use whatever string they want.

Updated administrator and circuits tests.

### DIFF
--- a/cli/lib/nf3.mjs
+++ b/cli/lib/nf3.mjs
@@ -5,7 +5,6 @@ import Queue from 'queue';
 import Web3 from 'web3';
 import WebSocket from 'ws';
 import crypto from 'crypto';
-import config from 'config';
 import ReconnectingWebSocket from 'reconnecting-websocket';
 import EventEmitter from 'events';
 import logger from '@polygon-nightfall/common-files/utils/logger.mjs';
--- a/cli/lib/nf3.mjs
+++ b/cli/lib/nf3.mjs
@@ -107,10 +107,7 @@ class Nf3 {
     this.ethereumSigningKey = ethereumSigningKey;
     this.zkpKeys = zkpKeys;
     this.currentEnvironment = environment;
-    axios.defaults.headers.common['X-APP-TOKEN'] = crypto
-      .createHash('sha256')
-      .update(environment.PROPOSER_KEY)
-      .digest('hex');
+    axios.defaults.headers.common['X-APP-TOKEN'] = environment.PROPOSER_KEY;
   }
 
   /**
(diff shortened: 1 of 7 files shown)


cli/lib/nf3.mjs
```diff
@@ -107,10 +107,7 @@ class Nf3 {
     this.ethereumSigningKey = ethereumSigningKey;
     this.zkpKeys = zkpKeys;
     this.currentEnvironment = environment;
-    axios.defaults.headers.common['X-APP-TOKEN'] = crypto
-      .createHash('sha256')
-      .update(environment.PROPOSER_KEY)
-      .digest('hex');
+    axios.defaults.headers.common['X-APP-TOKEN'] = environment.PROPOSER_KEY;
```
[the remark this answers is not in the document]
velkoskis: Got it! I am going to update this. I misunderstood.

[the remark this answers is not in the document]
velkoskis: @druiz0992 Do we have a preference if this random content should be hashed in any way or it is okay to be just a string that the user configures (preferably uuid)?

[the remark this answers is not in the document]
druiz0992: String is ok

config/default.js
```diff
@@ -187,6 +187,7 @@ module.exports = {
         process.env.CHALLENGER_KEY ||
         process.env.BOOT_CHALLENGER_KEY ||
         '0xd42905d0582c476c4b74757be6576ec323d715a0c7dcff231b6348b7ab0190eb',
```
druiz0992, line 189: Why is it defined as part of the environment? If you maintain it there, can you also add an entry in aws?

nightfall-optimist/src/utils/auth.mjs
```diff
@@ -1,11 +1,5 @@
...
-import crypto from 'crypto';
-
-const environment = config.ENVIRONMENTS[process.env.ENVIRONMENT] || config.ENVIRONMENTS.localhost;
-
 export default function auth(req, res, next) {
-  const token = req.headers['x-app-token'];
-  if (token === crypto.createHash('sha256').update(environment.PROPOSER_KEY).digest('hex'))
-    return next();
+  const token = req.headers['X-APP-TOKEN'];
+  if (token === process.env.AUTH_TOKEN) return next();
```
druiz0992, line 3: Should we take this from config?

config/default.js
velkoskis: Yes, let's add it to aws and polygonEdge!

nightfall-optimist/src/utils/auth.mjs
velkoskis: no, if we maintain it in the environments, we need to call it from there.

daveroga (approved):

imagobea (approved):
Thank you for taking care of this :raised_hands:

PR #1256, opened 2022-11-23
make duplicate commitments/nullifiers tx to check for proposer payment

LijuJoseJJ:
this PR fixes #https://github.com/EYBlockchain/nightfall_3_private/issues/240

elaborating changes in this PR.
commitment and nullifier check once checkTransaction passed (in TransactionSubmitEventHandler and in RollbackEventHandler - case where transaction are put in mempool), check is incoming transaction has duplicate commitments or nullifiers against a transaction already in mempool and if so compare their proposer provided fee, 
For TransactionSubmitEventHandler - save incoming transaction only if has higher payment.
For RollbackEventHandler - consider transaction as invalid and delete/(do not put in mempool) if payment is not higher


ChaitanyaKonda (dismissed):
`checkTransaction` is used by 1) `transactionSubmittedEventHandler` when a transaction is received and when it is called from inside `blockProposedEventHandler` when a transaction that the optimist does not know if is received as part of an L2 block 2) `blockProposedEventHandler` when it calls `checkBlock` to check the validity of L2 transactions 3) `rollbackEventHandler` to assess which transactions in an L2 block are invalid to delete them from transaction database.

The issue picked in this PR is about ensuring amongst the transactions with same nullifier/commitment, we only have one that is higher in payment to be proposed. Adding a transaction to mempool only is done during the receipt of a to be proposed transaction which happens when a user sends the transaction to proposer handled during situation (1) and when a transaction is added back to mempool which happens during situation (3). It is not relevant to handle this for situation (2).

As a result handling this change inside `checkTransaction` which applies to (1), (2) and (3) is incorrect. It needs to be handled only during (1) and (3). 

Please remove this from inside `checkTransaction` and make the relevant changes to apply to only the relevant situations

Westlad (approved):

druiz0992 (approved):

PR #1259, opened 2022-11-23
Add extended key usage checks

Westlad:

Our current method of checking x509 certificates to provide an identity check is quite good, but it can be made more secure. That's what this PR does.

Currently, when you submit a certificate we check the following are correct:
1. it's correctly signed by a CA that we trust (or, more accurately, `X509.sol` trusts)
2. it's in date
3. the key usage is `Certificate Sign`, if it's an intermediate certificate, and `Digital Signature` if it's an end user certificate

This is quite good but unfortunately the X509 key usage flags are quite broad-brush. `Digital Signature` can be set in a certificate that represents a very high level of identity check but also in one where the only check the CA made was that you own the email domain referred to in the certificate. This means that `nigerian.prince777@scam.com` could possibly be accepted (although we'd hope most reputable CAs would not accept @scam.com).

To be extra sure, we need to look into the Extended Key Usage data. This gives us a more granular view of the certificate's use and hence the level of checking that the CA did before issuing it. Unfortunately there is no overarching standard for what CAs put here; each has their own set of Object Ids and their CPS determines which they use. That's not a big problem, provided we make sure that `X509.sol` is loaded with the correct ones; it's just fiddly.
The OIDs that we wish to add are specified in the config on a per ETH_NETWORK basis.  It's an array of arrays, representing one group of OIDs for each CA. The group to use can be selected when the certificate is sent for validation.  This is safe because specifying the wrong group will cause the validation to fail - it's just lower gas cost than having the smart contract try every possible group of OIDs.

Even then, this may not give us enough assurance, so this PR also provides a check against the Certificate Policies extension.  It works in the same way to the Extended Key Usage but consumes Certificate Policy OIDs.

For test purposes, as we aren't using a real CA, I just chose OIDs for Code Signing, Time Stamping and Email Protection, and added made-up OIDs for the Certificate Policies, then issued a test end-user certificate with these OIDs set. Note that certificates without Key Usage and Certificate Policy extensions will not validate. This ensures that we only validate certificates which provide the appropriate level of assurance.

No

Normal CI tests will be adequate.  The X509 unit test has been updated and a new end user certificate created which exercises this new functionality.


Westlad:
I've added Mock certificates build from Digicert and Entrust certificate profiles, so they should be close the real thing. I've also removed my OID encoding app as I've now found a more comprehensive one online (linked in the docs about adding a certificate). Finally I improved X509 so that it copes better when items are marked as critical in the certificate.

druiz0992 (approved):

test/unit/SmartContracts/X509.unit.test.mjs
```diff
@@ -107,16 +161,34 @@ describe('DerParser contract functions', function () {
...
+      0,
     );
-    // now presenting the end user cert should work because the smart contract now trusts the Intermediate CA public key
+    // now presenting the end user cert should also work
     await X509Instance.validateCertificate(
       certChain[0].derBuffer,
       certChain[0].tlvLength,
       signature,
       true,
+      0,
```
israelboudoux, line 172: Hi @Westlad, this the index of the `oidGroup` right? I didn't like much this because the user must know the index. Wouldn't we be able to use the `OID Group` value as the index? (e.g. `2.16.840.1.114412.3.21.2`)

Westlad: Ah, no, that's not what the `oidGroup` is.  The example you gave is a single OID.  However a set (array) of OIDs must be applied to the certificate check.  This set is dependent on the CA being used. So for example Digicert might be `oidGroup = 1`, Entrust is `oidGroup = 2` etc. . Each index choice gives a different array of OIDs.  It's true that the user will need to know which `oidGroup` index to use with their certificate but we expect the person deploying the contracts to publish a mapping of `CA => oidGroup`.

israelboudoux: You're right, I made confusion.

israelboudoux (approved):

PR #1261, opened 2022-11-23
several fixes for aws deployment

druiz0992:
- `setContractState` and `transferOwnership` usually fail in goerli deployment. The reason is because there is no gas estimation being done
- Allow passing the name for both optimist and commitments db. The reason is to allow different optimists and clients to use the same AWS cluster and just have different names.
- `PROPOSER_MAX_BLOCK_PERIOD_MILIS` configures the maximum time that a proposer can wait to generate a block if there are pending transactions.
- AWS document db requires different format for db projections
- circuit registration usually failed in Goerli deployment because there was no gas estimation being done
- In some tests i found that the proposer wasn't set even though addresses matched. This case is fixed now
- Correct `proposer` and `hosted-utils-api-server` containers were not used. When nightfall is deployed with `start-nightfall`, old containers were used instead.
- Fixed client so that i downloads new circuits
- Enabled client liveness test back
- Recovered delete `build:testnet` from wallet package.json
- Make naming consistent between S3 and wallet
- Tentatively working testnet wallet
- Cilent start http ports at the beginning, but returns `400` while syncing. This is to ensure that AWS ECS service can verify container is alive

No

### DIFF
--- a/common-files/utils/web3.mjs
+++ b/common-files/utils/web3.mjs
@@ -54,13 +54,25 @@ export default {
     if (!config.WEB3_OPTIONS.from) throw Error('config WEB3_OPTIONS.from is not set');
     if (!config.ETH_PRIVATE_KEY) throw Error('config ETH_PRIVATE_KEY not set');
 
+    const fromAddress = await this.web3.eth.accounts.privateKeyToAccount(config.ETH_PRIVATE_KEY);
+
     const tx = {
+      from: fromAddress.address,
       to: contractAddress,
       data: rawTransaction,
       value,
-      gas: config.WEB3_OPTIONS.gas,
       gasPrice: config.WEB3_OPTIONS.gasPrice,
     };
+    let gas;
+    try {
+      gas = await this.web3.eth.estimateGas(tx);
+      logger.debug({ msg: 'Gas estimated at', gas });
+    } catch (error) {
+      gas = config.WEB3_OPTIONS.gas;
+      logger.warn({ msg: 'Gas estimation failed, use default', gas });
+    }
+
+    tx.gas = Math.ceil(gas * 2); // 50% seems a more than reasonable buffer
 
     const signed = await this.web3.eth.accounts.signTransaction(tx, config.ETH_PRIVATE_KEY);
     return this.web3.eth.sendSignedTransaction(signed.rawTransaction);
--- a/common-files/utils/web3.mjs
+++ b/common-files/utils/web3.mjs
@@ -47,6 +47,19 @@ export default {
     this.web3.currentProvider.connection.close();
   },
 
+  async estimateGas(tx) {
+    let gas;
+    try {
+      gas = await this.web3.eth.estimateGas(tx);
+      logger.debug({ msg: 'Gas estimated at', gas });
+    } catch (error) {
+      gas = config.WEB3_OPTIONS.gas;
+      logger.warn({ msg: 'Gas estimation failed, use default', gas });
+    }
+    gas = Math.ceil(gas * 2); // 50% seems a more than reasonable buffer
+    return gas;
+  },
+
   // function only needed for infura deployment
   async submitRawTransaction(rawTransaction, contractAddress, value = 0) {
     if (!rawTransaction) throw Error('No tx data to sign');
@@ -63,16 +76,7 @@ export default {
       value,
       gasPrice: config.WEB3_OPTIONS.gasPrice,
     };
-    let gas;
-    try {
-      gas = await this.web3.eth.estimateGas(tx);
-      logger.debug({ msg: 'Gas estimated at', gas });
-    } catch (error) {
-      gas = config.WEB3_OPTIONS.gas;
-      logger.warn({ msg: 'Gas estimation failed, use default', gas });
-    }
-
-    tx.gas = Math.ceil(gas * 2); // 50% seems a more than reasonable buffer
+    tx.gas = await this.estimateGas(tx);
 
     const signed = await this.web3.eth.accounts.signTransaction(tx, config.ETH_PRIVATE_KEY);
     return this.web3.eth.sendSignedTransaction(signed.rawTransaction);
--- a/config/default.js
+++ b/config/default.js
@@ -6,8 +6,8 @@ function configureAWSBucket() {
 }
 
 module.exports = {
-  COMMITMENTS_DB: 'nightfall_commitments',
-  OPTIMIST_DB: 'optimist_data',
+  COMMITMENTS_DB: process.env.COMMITMENTS_DB || 'nightfall_commitments',
+  OPTIMIST_DB: process.env.OPTIMIST_DB || 'optimist_data',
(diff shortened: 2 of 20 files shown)


daveroga (approved):

common-files/utils/web3.mjs
```diff
@@ -54,13 +54,25 @@ export default {
...
     const tx = {
+      from: fromAddress.address,
       to: contractAddress,
       data: rawTransaction,
       value,
-      gas: config.WEB3_OPTIONS.gas,
       gasPrice: config.WEB3_OPTIONS.gasPrice,
     };
+    let gas;
+    try {
```
israelboudoux, line 67: This block of code is repeated in other parts, please, create an utility method so that we can reuse it.

nightfall-deployer/src/circuit-setup.mjs
```diff
@@ -169,13 +169,23 @@ async function setupCircuits() {
...
+        const fromAddress = await web3.eth.accounts.privateKeyToAccount(config.ETH_PRIVATE_KEY);
         const tx = {
-          from: process.env.FROM_ADDRESS,
+          from: fromAddress.address,
           to: keyRegistry.options.address,
           data: call.encodeABI(),
-          gas: config.WEB3_OPTIONS.gas,
           gasPrice: config.WEB3_OPTIONS.gasPrice,
         };
+        try {
```
israelboudoux, line 180: Duplicated code!

nightfall-deployer/src/circuit-setup.mjs
```diff
@@ -199,13 +209,23 @@ async function setupCircuits() {
...
```
israelboudoux, line 220: Duplicated code!

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -79,7 +80,9 @@ export async function conditionalMakeBlock(proposer) {
     or we're no-longer the proposer (boo).
    */
 
-  logger.info(`I am the current proposer: ${proposer.isMe}`);
+  if (proposer.isMe) {
```
israelboudoux, line 83: The next `if` statament is equal to this one, please, move the log statement to there and remove this `if`.

config/default.js
```diff
@@ -136,6 +136,7 @@ module.exports = {
     DEFAULT_CIRCUIT_FILES_URL: 'https://nightfallv3-proving-files.s3.eu-west-1.amazonaws.com',
     DEFAULT_CONTRACT_FILES_URL: 'https://nightfallv3-proving-files.s3.eu-west-1.amazonaws.com',
   },
+  PROPOSER_MAX_BLOCK_PERIOD_MILIS: Number(process.env.PROPOSER_MAX_BLOCK_PERIOD) || 60 * 1000,
```
israelboudoux, line 139: Isn't 60 secs too low?

israelboudoux (changes requested):
LGTM, but there are a few things to check.

config/default.js
druiz0992: It’s a config value for testing. In deployment this value is customized. There’s no right value

nightfall-optimist/src/services/block-assembler.mjs
druiz0992: done

common-files/utils/web3.mjs
israelboudoux: What I mean is that this block
```
    try {
      gas = await this.web3.eth.estimateGas(tx);
      logger.debug({ msg: 'Gas estimated at', gas });
    } catch (error) {
      gas = config.WEB3_OPTIONS.gas;
      logger.warn({ msg: 'Gas estimation failed, use default', gas });
    }
```
 can be a utility function that can be used everywhere it is needed. e.g.
```
async function estimateGas(tx) {
    try {
      gas = await this.web3.eth.estimateGas(tx);
      logger.debug({ msg: 'Gas estimated at', gas });
    } catch (error) {
      gas = config.WEB3_OPTIONS.gas;
      logger.warn({ msg: 'Gas estimation failed, use default', gas });
    }
    return gas;
}

...
gas = estimateGas(tx);
...
```

druiz0992: i don't think this is needed

nightfall-optimist/src/routes/proposer.mjs
```diff
@@ -305,6 +312,12 @@ router.get('/mempool', async (req, res, next) => {
   }
 });
 
+/**
+ * Function to Propose a state update block  This just
+ * provides the tx data, the user will need to call the blockchain client
+ * @deprecated - this is now an automated process - no need to manually propose
+ * a block
+ */
```
imagobea, line 320: FYI

I removed this one myself. It says "deprecated" but it is not true since the adversary (or at least the old one) uses this endpoint (see test util `createBadBlock`).

At the same time, it will clash with the Swagger so...

nightfall-optimist/src/routes/proposer.mjs
```diff
@@ -79,13 +81,18 @@ router.post('/register', async (req, res, next) => {
...
           await enqueueEvent(() => logger.info('Start Queue'), 0); // kickstart the queue
         }
       }
+    } else if (address === currentProposer.thisAddress && !proposer.isMe) {
+      logger.warn(
+        'Proposer was already registered on the blockchain and with this Optimist instance, but proposer flag wasnt set - setting isMe flag',
+      );
+      proposer.isMe = true;
+      proposer.address = address;
+      await enqueueEvent(() => logger.info('Start Queue'), 0); // kickstart the queue
```
imagobea, line 95: Can you comment on this?
What is the situation we are trying to resolve :pray: Thank you

nightfall-optimist/src/routes/proposer.mjs
druiz0992: what do i do @imagobea ?

nightfall-optimist/src/routes/proposer.mjs
druiz0992: there is a scenario that i found in testing that this proposer's address was the current proposer's address, but the flag `proposer.isMe` wasn't set. It only happened at boot time from time to time.... so i just forced it to be the same

nightfall-optimist/src/routes/proposer.mjs
imagobea: IMO we can live without it

druiz0992: so, i just remove the `deprecated` or should i do something else @imagobea ?

israelboudoux (approved):
LGTM

nightfall-optimist/src/routes/proposer.mjs
imagobea: Sorry I missed your latest comment @druiz0992 
I would remove the entire docstring from L315-320, thank you (remember it will be properly documented when we have the swagger). Thanks again!

Issue #1260, opened 2022-11-23
Nightfall-node requested changes


imagobea:
- [ ] Make sure that we're happy with eth addresss + keys management
- [ ] Make sure that we're happy with gas estimations
- [ ] Refactor repeated code around tx
- [ ] Clean the logs
- [ ] New tests round

PR #1198, opened 2022-11-07
feat: nightfall node dev branch

imagobea:
## What does this implement/fix? Explain your changes.

With this PR, the Proposer app is no longer needed. The optimist itself can perform this role by calling the `proposer/register` endpoint once the optimist is online. 

**Important changes in this PR**

- Protects almost all optimist routes with an API Key (#1199, #1255)
- Handles proposer related transactions autonomously, using `PROPOSER_KEY` to sign transactions (#1200, #1225)
- Adds Swagger js docs to Optimist API (#1213)
- Updates `nf3`
- Refactors tests: note that one `optimist-sync` test was discontinued, but will be restored when tackling challenger

_Note that this PR is only the beginning towards getting rid of `nf3 Proposer` functions. Not all tests were updated to discontinue this practice. Similarly, proposer app was kept for now to facilitate launching proposer in ping-pong test._

## Does this close any currently open issues?

See related Epic on ZenHub

## What commands can I run to test the change? 

```bash
# Note the params and update as appropriate
curl -k -X POST -H "X-APP-TOKEN: 0ce4fee0-c765-43d6-973c-d404bfdde2e9" -d "url=[optimist-url]&stake=999999999&fee=10" http://localhost:8081/proposer/register
```

For further details, please refer to each PR.

## Any other comments?

n/a

PR #1262, opened 2022-11-24
refactor: nightfall-node requested changes

imagobea:

Refactor from transaction creation to submission:

- Call a function to create tx (includes gas estimation), using mutex

- Submit transaction and add the resulting promise to a queue (endpoint will return the `tx hash`, which is available right after signing)

- Actions that need to happen after the transaction has settled, will only take place if the tx was successful

#1260 

Like the other `nightfall-node` PRs, you can rely on curl to make calls to the optimist, or run the tests. Note that now routes are protected with an API key (instead of the sha256 derived from the proposer key).

## Any other comments?

Inspired by #1093


nightfall-optimist/src/services/transaction-sign-send.mjs
```diff
@@ -0,0 +1,46 @@
+import { web3 } from '@polygon-nightfall/common-files/utils/contract.mjs';
+import { estimateGas, estimateGasPrice } from '@polygon-nightfall/common-files/utils/gas.mjs';
+import logger from '@polygon-nightfall/common-files/utils/logger.mjs';
+import app from '../app.mjs';
```
imagobea, line 4: Had to try this anyways :joy:

cli/lib/nf3.mjs
```diff
@@ -1066,6 +1066,7 @@ class Nf3 {
       fee,
     });
     logger.debug(`Proposer /update response ${res}`);
+    return res?.data;
```
imagobea, line 1069: :memo:

Have updated tests with call to `/register`, so that these "await" for the register tx to be mined (event `NewCurrentProposer` can only be guaranteed in an env with 1 proposer).

Does `nf3` need to be updated? Tests using `nf3.registerProposer` seem to be passing anyway.
Keep an eye.

test/utils.mjs
```diff
@@ -522,3 +522,9 @@ export const emptyL2 = async ({ nf3User, web3, logs }) => {
   }
   await new Promise(resolve => setTimeout(resolve, 6000));
 };
+
+export async function isTransactionMined(txHash, web3) {
```
israelboudoux, line 526: The name of this function is a bit confusing because it gives the impression that a boolean value is gonna be returned. Maybe changing it to `waitTransactionToBeMined` or something similar will match with the function's goal.

test/utils.mjs
```diff
@@ -522,3 +522,9 @@ export const emptyL2 = async ({ nf3User, web3, logs }) => {
...
+  const receipt = await web3.eth.getTransactionReceipt(txHash);
+  if (receipt !== null) return;
+  await isTransactionMined(txHash, web3);
```
israelboudoux, line 529: This function is recursive, but there isn't a limit for retrying, maybe one could add a timeout, even if it is a small one, just to avoid too many recursive calls and consequently prevent the heap from being out of memory.

nightfall-optimist/src/services/transaction-sign-send.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+export async function getAddressNonce(ethAddress) {
+  let nonce;
+  try {
+    nonce = await web3.eth.getTransactionCount(ethAddress);
+  } catch (err) {
+    logger.error({
+      msg: 'Error obtaining address nonce',
+      err,
+    });
```
israelboudoux, line 29: If an error happens, the `nonce` will be `null`. At this point, you could throw the error you captured.

nightfall-optimist/src/services/transaction-sign-send.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+const { GAS, GAS_MULTIPLIER, GAS_ESTIMATE_ENDPOINT, GAS_PRICE, GAS_PRICE_MULTIPLIER } = config;
+
+const nonceMutex = new Mutex();
+
+const isListening = async () => {
+  try {
+    await web3.eth.net.isListening();
+  } catch (error) {
+    throw new Error('Web3 ws is disconnected, try again later');
```
israelboudoux, line 17: We could get the message from the `error` variable and pass in the exception.

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -168,25 +169,51 @@ export async function conditionalMakeBlock(proposer) {
...
+        );
+
+        // Submit tx and update db if tx is successful
+        // CHECK - db ops
+        txsQueue.push(async () => {
+          try {
+            const receipt = await sendSignedTransaction(signedTx);
+            logger.debug({ msg: 'Block proposed', receipt });
+
+            await removeTransactionsFromMemPool(block.transactionHashes);
```
israelboudoux, line 198: Maybe we could make use of `Promise.all([])` taking the these three calls inside it. e.g. 
```
await Promise.all([
  removeTransactionsFromMemPool(block.transactionHashes),
  removeCommitmentsFromMemPool(transactions.map(t => t.commitments.filter(c => c !== ZERO)).flat(Infinity)),
  removeNullifiersFromMemPool(transactions.map(t => t.nullifiers.filter(c => c !== ZERO)).flat(Infinity)),
]);
```
The code would be a bit cleaner and the behaviour it is the same.

nightfall-optimist/src/app.mjs
```diff
@@ -17,18 +16,21 @@ import {
...
 
 const app = express();
 
 const ethPrivateKey = environment.PROPOSER_KEY;
 const { address } = web3.eth.accounts.privateKeyToAccount(ethPrivateKey);
+const nonce = await getAddressNonce(address);
 
 app.set('ethPrivateKey', ethPrivateKey);
 app.set('ethAddress', address);
+app.set('nonce', nonce);
```
israelboudoux, line 31: Sorry, I didn't get the idea around this constant `nonce`.

nightfall-optimist/src/routes/proposer.mjs
```diff
@@ -91,20 +95,35 @@ router.post('/register', auth, async (req, res, next) => {
...
+        ethPrivateKey,
+        ethAddress,
+        proposersContractAddress,
+        txDataToSign,
+        stake,
+      );
+
+      // Submit tx and update db if tx is successful
+      // CHECK - I think the code beyond L132 can be removed (?) then db op could be moved here
+      txsQueue.push(async () => {
```
israelboudoux, line 116: This piece of code is repeated many times, changing only the log message sometimes and the calls inside. Maybe we can think in something to reuse it and make it less verbose. ;)

israelboudoux (changes requested):
LGTM, there are just a few things to check.

common-files/utils/gas.mjs
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * The underlying RPC `eth_estimateGas` will check the balance of the sender,
+ * this means that even though the call doesn't consume any gas,
+ * the `from` address must have enough gas to execute the tx
+ *
+ * @async
+ * @function estimateGas
+ * @param {object} tx A tx object
+ * @param {Web3} web3
+ * @param {number} gasDefault Has default
```
velkoskis, line 17: @param {number} gasDefault default gas price

nightfall-optimist/src/app.mjs
velkoskis: I think that it is needed to create a signed transaction for making a block if the current proposer is the actual running proposer.

nightfall-optimist/src/services/transaction-sign-send.mjs
```diff
@@ -0,0 +1,84 @@
...
+  // Check if web3 ws is available
+  await isListening();
+
+  logger.debug({ msg: 'Create transaction object...' });
+
+  let signedTx;
+  await nonceMutex.runExclusive(async () => {
+    // Update nonce if necessary
+    const _nonce = await getAddressNonce(from);
+    if (nonce < _nonce) {
```
velkoskis, line 45: why is this needed?

velkoskis (approved):
Good job on the implementation. LGTM

test/utils.mjs
imagobea: True. Somehow I thought that given the recursive nature, the name might be okay. But I agree that using `wait` is better. Thank you.

test/utils.mjs
imagobea: Agree. But not entirely sure about what to do.
When mentioning `setTimeOut`, do you mean inside the recursion? ie

```
// await getTransactionReceipt
// if condition for exiting
// else setTimeOut(waitTransactionToBeMined, 3000)
```

A counter might be possible too. Please let me know :pray:

nightfall-optimist/src/services/transaction-sign-send.mjs
imagobea: Tbh I think this implementation is a bit off, I'll take another look :pray:

nightfall-optimist/src/app.mjs
imagobea: I guess @velkoskis that is what I was trying to do, but since I am not refreshing the value stored in the state of the app, this doesn't make much sense. Thank @israelboudoux , I will re-think this one (alongside nonce error-handling above).
Thanks again both

nightfall-optimist/src/routes/proposer.mjs
imagobea: :eyes: That is true, but I would appreciate if I could do this in another PR :pray:

nightfall-optimist/src/services/transaction-sign-send.mjs
daveroga: I think this is to synchronize the nonce if another transaction has been sent before processing this transaction. If not then a revert will be thrown because the account has a higher nonce.

test/utils.mjs
daveroga: When I say a timeout I refer to a timeout from the perspective of the caller of the function `waitTransactionToBeMined` in in the case of the loop function it would be a counter.

nightfall-optimist/src/routes/proposer.mjs
israelboudoux: Sure! :D

nightfall-optimist/src/services/transaction-sign-send.mjs
velkoskis: @daveroga isn't that what nonceMutex.runExclusive() is taking care of?

test/utils.mjs
imagobea: Okay: removed recursion, added timer and counter.

nightfall-optimist/src/services/transaction-sign-send.mjs
imagobea: FYI `isListening` can return a boolean but also fail for many reasons
For the happy path, I am now returning the result from calling the web3 method, then I only proceed with the transaction if the result is "truthy"
If there's an error, the exception will be raised re-using the error message as per your suggestion

imagobea:
This is it sirs, thank you for your comments @velkoskis @daveroga @israelboudoux (and apologies again, like I said the page went "stale" and I was replying to some comments while missing others).
Have requested a re-review from Isra, but feel free to comment or ask further questions.
Thanks again :pray:

nightfall-optimist/src/routes/proposer.mjs
```diff
@@ -134,7 +156,13 @@ router.post('/register', auth, async (req, res, next) => {
         }
       }
     }
-    res.json({ receipt });
+
+    if (signedTx) {
```
israelboudoux, line 160: This `if` is not needed. If `signedTx` is `null`, then something wrong happened, right?

nightfall-optimist/src/services/block-assembler.mjs
```diff
@@ -221,13 +251,14 @@ export async function conditionalMakeBlock(proposer) {
...
         // blocks with them
-        await removeTransactionsFromMemPool(block.transactionHashes);
-        await removeCommitmentsFromMemPool(
-          transactions.map(t => t.commitments.filter(c => c !== ZERO)).flat(Infinity),
-        );
-        await removeNullifiersFromMemPool(
-          transactions.map(t => t.nullifiers.filter(c => c !== ZERO)).flat(Infinity),
-        );
+
+        // await removeTransactionsFromMemPool(block.transactionHashes);
```
israelboudoux, line 255: Pls, remove these code commented out.

nightfall-optimist/src/services/transaction-sign-send.mjs
```diff
@@ -0,0 +1,85 @@
...
+    const isListening = await web3.eth.net.isListening();
+    return isListening;
+  } catch (error) {
+    throw new Error(error.message);
+  }
+};
+
+export async function getAddressNonce(ethAddress) {
+  try {
+    const nonce = await web3.eth.getTransactionCount(ethAddress);
```
israelboudoux, line 24: You can return directly.

nightfall-optimist/src/services/transaction-sign-send.mjs
```diff
@@ -0,0 +1,85 @@
...
+import logger from '@polygon-nightfall/common-files/utils/logger.mjs';
+import { Mutex } from 'async-mutex';
+
+const { GAS, GAS_MULTIPLIER, GAS_ESTIMATE_ENDPOINT, GAS_PRICE, GAS_PRICE_MULTIPLIER } = config;
+
+const nonceMutex = new Mutex();
+
+const isWeb3Listening = async () => {
+  try {
+    const isListening = await web3.eth.net.isListening();
```
israelboudoux, line 15: You can return directly.

nightfall-optimist/src/services/transaction-sign-send.mjs
```diff
@@ -0,0 +1,85 @@
...
+  } catch (error) {
+    throw new Error(error.message);
+  }
+}
+
+// TODO document
+export async function createSignedTransaction(nonce, ethPrivateKey, from, to, data, value = 0) {
+  // Check if web3 ws is available
+  const result = await isWeb3Listening();
+  if (result) {
```
israelboudoux, line 35: I'd invert this `if` so that we return the negative asap since it is just a piece of code, while the code inside the `if` statement is pretty big. e.g. 
`if(! result) {
    throw new Error('Web3 ws not listening, try again later');
}
... // the rest...

nightfall-optimist/src/app.mjs
imagobea: We finally agreed to skip all nonce tracking for now and monitor behaviour

nightfall-optimist/src/routes/proposer.mjs
imagobea: Addressed in these commits [d8c542f](https://github.com/EYBlockchain/nightfall_3/pull/1262/commits/d8c542f8a22ab5163b9921975cfb2eb1203646f5), [5fb74e3](https://github.com/EYBlockchain/nightfall_3/pull/1262/commits/5fb74e347080f688f49bb85abe3e9f9c953d3d9e) (I cleaned up all nonce stuff as discussed)

nightfall-optimist/src/services/block-assembler.mjs
imagobea: Will address in #1198

nightfall-optimist/src/services/transaction-sign-send.mjs
imagobea: Of course

nightfall-optimist/src/services/transaction-sign-send.mjs
imagobea: Like we said it is not actually possible to return straight from an `asyn` call within a `try/catch` block (see https://jakearchibald.com/2017/await-vs-return-vs-return-await/).

In any case - like you pointed out - both methods `isWeb3Listening` and `getAddressNonce` were used locally and I've ended up removing both.

Thanks

nightfall-optimist/src/services/transaction-sign-send.mjs
imagobea: Removed, see comment re `isWeb3Listening`

nightfall-optimist/src/services/transaction-sign-send.mjs
israelboudoux: Amazing, thank you!

israelboudoux (approved):
LGTM

PR #1264, opened 2022-11-24
Chait/optimist fixes

ChaitanyaKonda:
`checkAlreadyInBlock` inside `transactionSubmittedEventHandler` is not be relevant because checking for duplicate transaction hashes is an incomplete check. Duplicacy check is done by commitments and nullifiers which is handled inside `checkTransaction` already. Also, this function caters to remine logic before the code transitioned to waiting for certain number of confirmation blocks before assuming the event is final to be processed.. Also error thrown from `checkAlreadyInBlock` would create a challenge incorrectly because it does not hold the necessary arguments required by `services/challenges.mjs`.  On a different note, if optimist receives a transaction as part of an L2 block that it was not aware of, then optimist  does not save `l2BlockNumber` whilst saving this transaction to it’s database. As a result, when `rollbackEventHandler` is called during a rollback of an L2 block, if an incorrect transaction, for example duplicate nullifier/commitment is being deleted, it risks deleting the original correct transaction too.
An even more important implication of this, when `removeTransactionsFromMemPool` which removes transactions from mempool is called inside `blockProposedEventHandler` for transactions in an L2 block, it does not work correctly because this function updates transactions by querying based on `blockNumberL2` too which in this case is unset. 

Fixes the issues
https://github.com/EYBlockchain/nightfall_3_private/issues/236
https://github.com/EYBlockchain/nightfall_3_private/issues/234

github actions

### DIFF
--- a/nightfall-optimist/src/event-handlers/transaction-submitted.mjs
+++ b/nightfall-optimist/src/event-handlers/transaction-submitted.mjs
@@ -2,11 +2,7 @@
  * Module to handle new Transactions being posted
  */
 import logger from '@polygon-nightfall/common-files/utils/logger.mjs';
-import {
-  saveTransaction,
-  getBlockByTransactionHash,
-  getTransactionByTransactionHash,
-} from '../services/database.mjs';
+import { saveTransaction } from '../services/database.mjs';
 import checkTransaction from '../services/transaction-checker.mjs';
 import TransactionError from '../classes/transaction-error.mjs';
 import { getTransactionSubmittedCalldata } from '../services/process-calldata.mjs';
--- a/nightfall-optimist/src/event-handlers/transaction-submitted.mjs
+++ b/nightfall-optimist/src/event-handlers/transaction-submitted.mjs
@@ -11,7 +11,7 @@ import { getTransactionSubmittedCalldata } from '../services/process-calldata.mj
  * This handler runs whenever a new transaction is submitted to the blockchain
  */
 async function transactionSubmittedEventHandler(eventParams) {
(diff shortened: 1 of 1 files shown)


ChaitanyaKonda:
Please close the relevant issues in the other repo when merged

nightfall-optimist/src/event-handlers/transaction-submitted.mjs
```diff
@@ -2,52 +2,16 @@
...
-
-  return transaction; // but it's otherwise ok
-}
-
 /**
  * This handler runs whenever a new transaction is submitted to the blockchain
  */
 async function transactionSubmittedEventHandler(eventParams) {
-  const { offchain = false, fromBlockProposer, ...data } = eventParams;
+  const { offchain = false, fromBlockProposer, blockNumberL2 = -1, ...data } = eventParams;
```
RogerTaule, line 14: Just a minor thing: shouldn't we initialize `fromBlockProposer = false` ? When `transactionSubmittedEventHandler` is called in `proposer.mjs`, this variable is not specified, so that the condtions holds because the parameter is `undefined`, but I think it will be more "correct" if we initialize it to `false`

RogerTaule (changes requested):

nightfall-optimist/src/event-handlers/transaction-submitted.mjs
ChaitanyaKonda: This variable wasn't introduced in this PR but the suggestion is still valid.
Pushed this change

RogerTaule (approved):

LijuJoseJJ (approved):

PR #1265, opened 2022-11-25
New adversary

RogerTaule:
This PR modifies the adversary paradigm. It is now composed of two services: `lazy-optimist` and `bad-client`
- The `lazy-optimist` is an optimist that doesn't check the transaction's validity when received, so that it can create blocks with invalid transactions. It can also create blocks with invalid properties such as an invalid leaf count. This is done by modifying the raw proposeBlock transaction, so that it doesn't mess with the database information. In order to create a bad block, one can call `/block/make-now/:badBlockType?` and specify one of the following types: `IncorrectTreeRoot`, `IncorrectLeafCount`, `IncorrectFrontierHash`, `DuplicateTransaction`.
- The `bad-client` is a client that has the ability to create bad transactions. If nothing is specified, it will create valid ones. In order to create a bad transaction, one simply needs to pass an extra parameter to the action he wants to perform (deposit, transfer...) called `transactionType`. It can have 4 different values:
1. `DuplicateNullifier` - It uses commitments that has been previously nullified to create the transaction. Does not apply to deposits. It will fail if there's no previous transactions.
2. `IncorrectInput` - It randomly modifies a transaction value after the proof has been created.
3. `IncorrectProof` - It modifies the proof to be a fake one (currently disabled)
4. `IncorrectHistoricRootNumber` - It modifies the historic root numbers before creating the proof (so that the proof contained in the transaction is valid)

Both services has been created by transpiling the original services and adding some minor modifications using regex. Very minor changes has been done to the client code to make the transpilation and changes easier.  Additionally, nf3 is also transpiled and slightly modified to support calls to the bad client.

This PR also introduces a few fixes:
- `verify` function in the `Verifier.sol` contract now checks if the verification key length is at least 33. Without that check, if the verification key length is smaller the verification function was reverting with panic code.
- If `circuitHash` is invalid (meaning there's no verification key stored in the smart contract, the `TransactionError` is thrown prior to even calling verify from snarkjs (because it would throw an unexpected error)
- Code for creating challenges for for duplicate nullifiers and commitments that are in the same block is fixed.
- [NEW] Client subscribes to Challenges Smart Contract so that it actually listens and performs the proper rollbacks
- [NEW] Rollback with duplicate commitments and nullifiers inside the same block is handled correctly. All duplicates needs to be removed, not just two.


daveroga (commented):
Good job @imagobea! Only few comments to think about.

test/utils.mjs
```diff
@@ -221,6 +221,15 @@ export class Web3Client {
       if (timeout === 0) throw new Error('Timeout in waitForEvent');
     }
 
+    if (expectedEvents[0] === 'blockProposed') {
```
daveroga, line 224: I would rewrite `waitForEvent` instead of doing this explicitly for `blockProposed` insde. Instead of always looking at eventLogs[0], I'd search in all the array for the event and do the push or shift operations based on the specific `expectedEvents` passed.

daveroga (changes requested):

test/utils.mjs
daveroga: Also in the `start-nightfall` the `nightfall_3_mongodb-adversary-db` volume should be removed if present. If it's not removed then the test only works the first time.

daveroga (approved):

test/adversary.test.mjs
```diff
@@ -198,132 +218,276 @@ describe('Testing with an adversary', () => {
...
-    it('Test incorrect public input transfer commitment', async () => {
-      console.log('Testing incorrect public input transfer commitment...');
-      await nf3AdversarialProposer.makeBlockNow('IncorrectPublicInputTransferCommitment');
-      console.log('Waiting for rollback...');
-      await waitForRollback();
-      console.log('Rollback incorrect public input transfer commitment completed');
-      expect(challengeSelector).to.be.equal(challengeSelectors.challengeProofVerification);
-    });
+    describe('Withdraw rollbacks', async () => {
+      it.skip('Test duplicate transaction withdraw', async () => {
```
israelboudoux, line 375: This one is being skipped, is that correct?

test/adversary.test.mjs
```diff
@@ -198,132 +218,276 @@ describe('Testing with an adversary', () => {
...
+      });
 
-    it('Test incorrect historic root', async () => {
-      console.log('Testing incorrect root...');
-      await nf3AdversarialProposer.makeBlockNow('IncorrectHistoricRoot');
-      console.log('Waiting for rollback...');
-      await waitForRollback();
-      console.log('Rollback incorrect historic root completed');
-      expect(challengeSelector).to.be.equal(challengeSelectors.challengeHistoricRoot);
+      it.skip('Test incorrect proof withdraw', async () => {
```
israelboudoux, line 440: This test is being skipped, is that correct?

israelboudoux (approved):
LGTM

druiz0992 (approved):

PR #1271, opened 2022-11-29
fix: bumping circom version to 2.1.2

RogerTaule:
Updates circom circuits from version 2.1.0 to 2.1.2, which fixes a bug in the witness generator (although it doesn't impact our project because we are using wasm files to generate the witness)


IlyasRidhuan (approved):

siosw (approved):
LGTM

PR #1272, opened 2022-11-29
Fees are only paid in L2

RogerTaule:
This PR simplifies the fees in Nightfall, so that all fees will now be paid only in L2. This change is done for multiple reasons: 
- Simplicity: It will simplify the fees concept since all the fees would be paid always the same way.
- Optimization: Submit transaction function will be cheaper. ProposeBlock function will also be cheaper so that the cost per transaction will be less.
- Not having to deal with FX rates. When assembling the block, we are sorting by fee. However, the way it is done right now is wrong because only considers L2 fees but we also allow for L1 fees. So to do it properly we'd need to introduce FX rates, which I think is undesired
- There's no impact in the user side. They already had to deposit L2 token so that fees can be paid for L2 operations such as transfer, so there's no difference in doing that deposit earlier. Also, there's no issue in doing the first deposit since two weeks ago I modified the code to be suitable (fees can already be paid in L2 when depositing L2 token)
- Since the Smart Contracts will be frozen, either this change is done now or never


imagobea (approved):
LGTM (except for the description, please update!)
Just left a couple suggestions and a question.
Thanks @RogerTaule , good work :muscle:

druiz0992 (approved):

PR #1275, opened 2022-11-30
Client was not doing rollbacks

RogerTaule:
Since `emit Rollback` was moved into Challenges and the client was only listening for events for the State contract, rollback events weren't caught by the client, creating discrepancies between optimist and client database. 
In order to avoid similar issues in a future, the client is now subscribed to all the contracts. This doesn't have any impact to the client behaviour since all those events that are not handled are simply ignored.
State sync event handling has been fixed for both: client and optimist.
An additional log has been added to transaction checker to help us detect root discrepancies easier.
N/A

N/A
## Any other comments?
N/A

### DIFF
--- a/nightfall-client/src/event-handlers/subscribe.mjs
+++ b/nightfall-client/src/event-handlers/subscribe.mjs
@@ -7,13 +7,28 @@ import { waitForContract } from '@polygon-nightfall/common-files/utils/contract.
 import constants from '@polygon-nightfall/common-files/constants/index.mjs';
 import logger from '@polygon-nightfall/common-files/utils/logger.mjs';
 
-const { STATE_CONTRACT_NAME } = constants;
+const {
+  STATE_CONTRACT_NAME,
+  CHALLENGES_CONTRACT_NAME,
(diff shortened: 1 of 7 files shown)


nightfall-client/src/event-handlers/subscribe.mjs
```diff
@@ -7,13 +7,28 @@ import { waitForContract } from '@polygon-nightfall/common-files/utils/contract.
...
+    PROPOSERS_CONTRACT_NAME,
+  ];
+  const contracts = await Promise.all(contractNames.map(c => waitForContract(c)));
+  const emitters = contracts.map(e => {
+    const emitterC = e.events.allEvents();
+    emitterC.on('changed', event => callback(event, args));
+    emitterC.on('data', event => callback(event, args));
+    return emitterC;
+  });
   logger.debug('Subscribed to layer 2 state events');
```
daveroga, line 32: Better `Subscribed to layer 2 contracts events` because now it's not only the state contract.

daveroga (approved):
LGTM. Only this change in the message

PR #1276, opened 2022-11-30
Change current proposer automatically called by proposer

### LINKED ISSUE #1249: change proposer added to proposer so that proposer is changed automatically

eugypalu:
this pr adds to the proposer app the ability to automatically call the changeCurrentProposer.
The changeCurrentProposer is automatically called by the next proposer, or, if the next proposer does not make the call, it waits for a period of time after which it is called by any other active proposer

close #1249 

`npm run ping-pong
`then in another terminal: 
`docker logs -f proposer_1`
this last command will log the `sprintNumber`, the `nextProposer` from the list and who is the `proposer` that calls the `changeCurrentProposer` function

### DIFF
--- a/apps/proposer/src/proposer.mjs
+++ b/apps/proposer/src/proposer.mjs
@@ -51,7 +51,7 @@ export default async function startProposer(nf3, proposerBaseUrl) {
   logger.info('Listening for incoming events');
 }
 
-export default async function checkAndChangeProposer(nf3) {
+export async function checkAndChangeProposer(nf3) {
   const maxBlockWaitTime = 5;
   logger.info('Checking Proposer...');
   const proposerStartBlock = await nf3.proposerStartBlock();
--- a/apps/proposer/src/proposer.mjs
+++ b/apps/proposer/src/proposer.mjs
@@ -4,10 +4,12 @@ Module that runs up as a proposer
 */
 import logger from '@polygon-nightfall/common-files/utils/logger.mjs';
 
+const TIMER_CACP = process.env.TIMER_CACP || 30;
+const MAX_ROTATE_TIMES = process.env.MAX_ROTATE_TIMES || 2;
 /**
 Does the preliminary setup and starts listening on the websocket
 */
-export async function startProposer(nf3, proposerBaseUrl) {
+export default async function startProposer(nf3, proposerBaseUrl) {
   logger.info('Starting Proposer...');
   // Mnemonic are only required for services connecting to a client that
   // can generate a compressed PKD.
@@ -27,6 +29,7 @@ export async function startProposer(nf3, proposerBaseUrl) {
 
   // If the emitter is not defined it causes the process to exit
   const blockProposeEmitter = await nf3.startProposer();
+  checkAndChangeProposer(nf3);
   blockProposeEmitter
     .on('receipt', (receipt, block) => {
       logger.debug(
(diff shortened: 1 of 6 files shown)


apps/proposer/src/proposer.mjs
```diff
@@ -1,9 +1,43 @@
 /* eslint-disable import/no-unresolved */
+/* eslint-disable no-await-in-loop */
 /**
 Module that runs up as a proposer
 */
 import logger from '@polygon-nightfall/common-files/utils/logger.mjs';
 
+const TIMER_CACP = process.env.TIMER_CACP || 30;
```
druiz0992, line 8: Add these two variables to config

apps/proposer/src/proposer.mjs
```diff
@@ -1,9 +1,43 @@
...
+const MAX_ROTATE_TIMES = process.env.MAX_ROTATE_TIMES || 2;
+
+async function checkAndChangeProposer(nf3) {
```
druiz0992, line 11: Can you add a brief description of this function?

druiz0992 (changes requested):
Left some comments

apps/proposer/src/proposer.mjs
eugypalu: added, thanks

imagobea (commented):
Hello @eugypalu (and everyone involved for that matter)
Thanks for this.

Just leaving a comment since I don't mean to block anybody. Also making some notes bearing in mind "nightfall-node".

- The new env vars are a bit too "mysterious". I assume `CAPC` in TIMER_CACP stands for "Check And Change Proposer", perhaps TIMER_CHANGE_PROPOSER (?) Could also contain the unit in which one is meant to express this time, ie milliseconds, etc.

- As per MAX_ROTATE_TIMES.. no clue. I would have to deep-dive in the logic. Please consider a different name for this one too.

- Today we talked about a variable that was causing some trouble `PROPOSER_MAX_BLOCK_PERIOD_MILIS`, because it was automating the make-block process but leaving the proposer without further control. Perhaps something to bear in mind.

- With "nightfall-node" we will lose the proposer app, so we will have to put the logic in the optimist (as we all know)

- Last but not least: `Nf3` was typically a place for API calls and whatnot. I can see that now we are using it more and more to reach contract functions/public values. Perhaps this is worth a conversation :pray: :pray: :pray: 

Many thanks again

apps/proposer/src/proposer.mjs
druiz0992: if you are going to change @imagobea, then also use config vars instead of env vars (that's why we defined them).

The idea to have vars in config is that one can see which things are configurable. If configuration is only from env vars, is difficult to get a full picture of all the parameters needed.

eugypalu:
> Hello @eugypalu (and everyone involved for that matter) Thanks for this.
> 
> Just leaving a comment since I don't mean to block anybody. Also making some notes bearing in mind "nightfall-node".
> 
> * The new env vars are a bit too "mysterious". I assume `CAPC` in TIMER_CACP stands for "Check And Change Proposer", perhaps TIMER_CHANGE_PROPOSER (?) Could also contain the unit in which one is meant to express this time, ie milliseconds, etc.
> * As per MAX_ROTATE_TIMES.. no clue. I would have to deep-dive in the logic. Please consider a different name for this one too.
> * Today we talked about a variable that was causing some trouble `PROPOSER_MAX_BLOCK_PERIOD_MILIS`, because it was automating the make-block process but leaving the proposer without further control. Perhaps something to bear in mind.
> * With "nightfall-node" we will lose the proposer app, so we will have to put the logic in the optimist (as we all know)
> * Last but not least: `Nf3` was typically a place for API calls and whatnot. I can see that now we are using it more and more to reach contract functions/public values. Perhaps this is worth a conversation 🙏 🙏 🙏
> 
> Many thanks again

Hello @imagobea  thanks for the comments

- fixed, you are right, `CACP` is not very explanatory as a name :)
- about `MAX_ROTATE_TIMES`, Every time the `checkAndChangeProposer` function is called, it checks if the proposer should call the `changecurrentproposer`, and the call is automatically made by the directly involved proposer (the proposer that will become the currentProposer right after). In case it is not available, the other proposers have to wait a number of blocks equal to `MAX_ROTATE_TIMES * rotateProposerBlocks` and then they are allowed to make the `changecurrentproposer` call.
- Regarding `NF3`, I saw that some calls to smart contract views were already present, while calls to functions were structured as API calls, so I thought I would follow this method, but absolutely we can talk about it if there is any concern

druiz0992 (approved):

Issue #1234, opened 2022-11-19
X509 


druiz0992:
X509 required activities
- [x] Assign good certificase to some users
- [x] Assign bad certificates to some users
- [x] Add final CA roots
- [ ] Other...

israelboudoux:
I have created a script for creating the certificates - it creates the Root CA certificate and the X users' certificates. The certificates are being generated correctly with the expected flags, but when I tried to test them using the X509 Unit Test, it fails for the test _**'Should validate the user certificate'**_, while passes the other two tests - this is a modified version for the X509 unit test that I changed because I'm not generating an intermediate CA (branch: israelboudoux/test-x509). The error is the following: 

```
     Error: VM Exception while processing transaction: reverted with reason string 'X509: Decrypt does not have a leading zero octets'
      at X509.onlyOwner (nightfall-deployer/contracts/Ownable.sol:26)
      at X509.validateSignatureAndExtractMessageHash (nightfall-deployer/contracts/X509.sol:91)
      at X509.checkSignature (nightfall-deployer/contracts/X509.sol:268)
      at X509.validateCertificate (nightfall-deployer/contracts/X509.sol:300)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

The point is that when I comment out the following two `require` validation [here](https://github.com/EYBlockchain/nightfall_3/blob/master/nightfall-deployer/contracts/X509.sol#L97) the test passes successfully. I understand that some parameters are expected after decrypting the signature, but I don't what to do for these parameters comply with the expected validations in the X509 contract. 

Follows the certificate + priv_key attached that I used for testing, but the code for generating it can be seeing in the PR at [https://github.com/EYBlockchain/nightfall_3_private/pull/313](https://github.com/EYBlockchain/nightfall_3_private/pull/313).

israelboudoux:
Hey @Westlad, would you mind adding the flags you used to create each of the certs (CA & end-user)? Maybe I'm missing some of them.

Thanks!

Westlad:
Hi @israelboudoux.  For current Master they are:
For a CA - key usage flag: Certificate Sign, CRL Sign
For an end user cert: Digital Signature, Non repudiation .
Once #1259 is merged, you will also need:
Extended Key Usage must be whatever you tell X509.sol to use when you deploy (must be something).  The test uses: Code Signing, E-mail Protection, Time Stamping (probably not what we would use in a real deploy).
[...]

PR #1279, opened 2022-11-30
Scripts for generation of the certificates

israelboudoux:
Creates the scripts for generating the x509 certificates

#1234

The pipeline has a unit test for this

## Any other comments?
No


test/_certificates/conf/extfile.conf
```diff
@@ -0,0 +1,4 @@
+[extensions]
+subjectKeyIdentifier=hash
+authorityKeyIdentifier=keyid,issuer
+keyUsage=digitalSignature,nonRepudiation
```
Westlad, line 4: I think you will also need `extendedKeyUsage` and `certificatePolicies` to be set for these certificates to work.  The details don't matter provided the OIDs you use are passed into the smart contract at deployment via `addExtendedKeyUsage` and `addCertificatePolicies`

test/_certificates/root_ca.crt
```diff
@@ -0,0 +1,34 @@
+-----BEGIN CERTIFICATE-----
```
Westlad, line 1: Would it not be better to output the certs in DER format so that they are compatible with `X509.sol`?

test/_certificates/root_ca.pub_key
```diff
@@ -0,0 +1,14 @@
+-----BEGIN PUBLIC KEY-----
+MIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEAqo/ZXI+Xfx7U4wX5m3q9
```
Westlad, line 2: Suggest extracting the modulus, exponent and keyIdentifier for compatibility with `X509.sol`'s `RSAPublicKey` struct.

Westlad (changes requested):
suggest the private keys exported are in PKCS#1 format if not already.

Westlad:
Would it not be good to have unit tests for the certificates to ensure they work with `X509.sol`?

test/_certificates/root_ca.crt
israelboudoux: As this certificate is the main Root of Trust, it isn't sent to the Smart Contract, then I didn't export it in DER format, only the end-user ones are in DER format.

test/_certificates/root_ca.pub_key
israelboudoux: These information are extracted to the following files, with exception of the modulus that uses the default one:
- test/_certificates/root_ca.pub_key.modulus
-  test/_certificates/root_ca.authority_key

test/_certificates/root_ca.crt
Westlad: That's true, we don't actually need the certificate for nightfall to work, but it's a bit inconsistent to output it in PEM format when the others are DER.

test/_certificates/root_ca.pub_key
Westlad: Ah, sorry. Didn't spot those.

test/_certificates/root_ca.crt
israelboudoux: Yes, I'll remove this one because we don't need it.

israelboudoux:
> suggest the private keys exported are in PKCS#1 format if not already.

israelboudoux:
> Would it not be good to have unit tests for the certificates to ensure they work with `X509.sol`?

Done!

Westlad (approved):

test/_certificates/gen-end_user-bad-certificates.sh
```diff
@@ -0,0 +1,56 @@
...
+
+user_name=user-$1
+
+echo "Generating self-signed certificate for user '$user_name'"
+
+openssl genpkey -outform DER -pkeyopt rsa_keygen_bits:4096 -algorithm RSA -out user/invalid/$user_name.priv_key -quiet
+
+# generates a certification request
+openssl req -new \
+  -subj "/C=IN/ST=Mumbai/O=Polygon Technology/OU=Nightfall Team/CN=$user_name/emailAddress=$user_name@polygon.technology" \
```
signorecello, line 25: Just remember to remove references to Polygon (also on the email, etc)

signorecello (changes requested):
Just some quick change regarding naming

test/_certificates/gen-end_user-bad-certificates.sh
israelboudoux: Very good spot @signorecello!

signorecello (approved):
LGTM master Isra

PR #1282, opened 2022-12-01
Adding function so that proposer can modify time between blocks

RogerTaule:
Introduces a route so that a proposer can decide the time between blocks. If set to 0, the timer isn't used and the algorithm will wait until the block is full to create a block
N/A
Start nightfall and a proposer and then run `curl http://localhost:8088/block/block-time -H 'Content-Type: application/json' -d '{"timeMs": 60000}'`

## Any other comments?
N/A


daveroga (approved):

velkoskis (approved):

imagobea (changes requested):
Thank you @RogerTaule Please consider the suggested changes :pray: Let me know if you have any questions. Thanks again!

imagobea (approved):
Not ideal but please remove the validation for now. Thank you for your time!!

PR #1286, opened 2022-12-01
nightfall-node: misc updates

imagobea:

This PR mostly fixes almost all tests failing on the feature branch. It also cleans up to some extend almost all test suites by:

- Removing unnecessary tests, e.g. duplicates, tests that don't make sense on a single optimist environment
- Improves consistency across test suites

The `ping-pong` still fails, but hopefully we can pair with @daveroga He might be able to help too in identifying a way around `rollbacks` monitoring, which was relying on certain websocket events.
We should also study the `optimist-resync` test to agree on what should be the behaviour.

No

Check GitHub Actions.

## Any other comments?

No


common-files/utils/gas.mjs
```diff
@@ -37,15 +37,15 @@ export async function estimateGas(tx, web3, gasDefault = 4000000, gasMultiplier
...
- * @param {number} gasPriceMultiplier Buffer to apply to estimated gas price - Has default
+ * @param {string} [gasEstimateEndpoint] URL for querying gas prices - Has default
+ * @param {number} [gasPriceDefault] Has default
+ * @param {number} [gasPriceMultiplier] Buffer to apply to estimated gas price - Has default
  * @returns {Promise<number>}
  */
 export async function estimateGasPrice(
-  gasEstimateEndpoint,
   web3,
+  gasEstimateEndpoint = 'https://vqxy02tr5e.execute-api.us-east-2.amazonaws.com/production/estimateGas',
```
druiz0992, line 48: Where is this endpoint located?

imagobea: GM @druiz0992 the value is the same as per the config, but since common files is a library, I prefer to avoid using the config. Or do you think we will stop this service? We could pass also empty string and will default to the `web3` option - this function has 2 backup options.

test/circuits.test.mjs
```diff
@@ -1,265 +1,249 @@
...
-    );
-    expectTransaction(singleTransfer);
-
-    await emptyL2({ nf3User: nf3Users[0], web3: web3Client, logs: eventLogs });
-
-    const afterBalances = await getBalance();
-
-    expect(afterBalances[0] - beforeBalances[0]).to.be.equal(-8);
-    expect(afterBalances[1] - beforeBalances[1]).to.be.equal(7);
-  });
```
imagobea, line 96: Removed skipped test

test/circuits.test.mjs
```diff
@@ -1,265 +1,249 @@
...
-
-  it('Test that all circuits are working without fees', async () => {
```
imagobea, line 98: Transformed this one into `describe` for readability

test/e2e/protocol/proposer.test.mjs
```diff
@@ -244,94 +249,38 @@ describe('Basic Proposer tests', () => {
...
-        }
-
-        // const res = await secondProposer.changeCurrentProposer();
-        // expectTransaction(res);
-        numChanges++;
-      } catch (err) {
-        console.log(err);
-      }
-    }
-    expect(numChanges).to.be.equal(CHANGE_PROPOSER_NO_TIMES);
```
imagobea, line 317: Removed skipped test

test/e2e/tokens/erc20.test.mjs
```diff
@@ -29,286 +28,183 @@ const {
...
-
-      await nf3Users[0].makeBlockNow();
-      await web3Client.waitForEvent(eventLogs, ['blockProposed']);
-
-      const afterZkpPublicKeyBalance =
-        (await nf3Users[0].getLayer2Balances())[erc20Address]?.[0].balance || 0;
-      expect(afterZkpPublicKeyBalance - currentZkpPublicKeyBalance).to.be.equal(
-        transferValue - fee,
-      );
-    });
```
imagobea, line 134: This test is no longer needed since deposit fees are paid in L2

test/gas.test.mjs
```diff
@@ -212,44 +207,15 @@ describe('Gas test', () => {
...
-        ).to.be.greaterThan(0);
-        const res = await nf3Users[0].finaliseWithdrawal(withdrawal);
-        expectTransaction(res);
-        const endBalance = await web3Client.getBalance(nf3Users[0].ethereumAddress);
-        expect(parseInt(endBalance, 10)).to.be.lessThan(parseInt(startBalance, 10));
-        logger.debug(`The gas used for finalise withdraw, back to L1, was ${res.gasUsed}`);
-      } else {
-        logger.debug('Not using a time-jump capable test client so this test is skipped');
-        this.skip();
-      }
```
imagobea, line 244: This test is not needed since it exists already for each token in `e2e/tokens`

test/e2e/protocol/proposer.test.mjs
```diff
@@ -18,16 +17,24 @@ chai.use(chaiAsPromised);
...
+const {
+  tokenConfigs: { tokenType, tokenId },
+  mnemonics,
+  signingKeys,
+  fee,
+  transferValue,
+} = config.TEST_OPTIONS;
 
 const web3Client = new Web3Client();
+const web3 = web3Client.getWeb3();
```
velkoskis, line 29: maybe this line is redundant, don't you have the same object in web3Client after line 28?

test/e2e/protocol/proposer.test.mjs
velkoskis: don't we need a change current proposer test?

velkoskis (approved):
Good job. Left some minor comments. Thanks for refactoring and throwing away code that was not used in the tests! They look much better now.

test/e2e/protocol/proposer.test.mjs
imagobea: We do. But for that we need to launch an environment with multiple optimists and this test won't do that. I believe there's some kind of ping-pong where proposer "rotation" is being tested.

test/e2e/protocol/proposer.test.mjs
imagobea: Not sure what you mean, perhaps you could do `web3Client.web`, but some tests had this approach so I just followed this pattern ¯\_(ツ)_/¯

test/e2e/protocol/proposer.test.mjs
velkoskis: okay, thanks

test/e2e/protocol/proposer.test.mjs
velkoskis: fair enough!

test/utils.mjs
```diff
@@ -20,7 +20,7 @@ export const waitForTimeout = async timeoutInMs => {
 
 export const topicEventMapping = {
   BlockProposed: '0x566d835e602d4aa5802ee07d3e452e755bc77623507825de7bc163a295d76c0b',
-  Rollback: '0xea34b0bc565cb5f2ac54eaa86422ae05651f84522ef100e16b54a422f2053852',
+  Rollback: '0x03404b6979bc5ed297ee1fc6daf00fcb29bbd74b934488ef775c1d70c5e7cd8d',
```
imagobea, line 23: :fire:

common-files/utils/gas.mjs
israelboudoux: This is a good point, as we will deploy this somewhere the service should be consumed from this "new" stack. We should create a new environment var for this endpoint.

CC: @imagobea @druiz0992

PR #1296, opened 2022-12-07
Client minor fixes

RogerTaule:
This PR does the following: 
- Removes unncessary logs in the client files
- In the client side, `new Transaction` was called twice and there was no need for that. This PR removes the second one and modifies the code accordingly.
- Fixes `proof.mjs` class in common files


IlyasRidhuan (approved):

nightfall-client/src/utils/getCommitmentInfo.mjs
```diff
@@ -126,14 +126,20 @@ export const getCommitmentInfo = async txInfo => {
     spentCommitments.push(...oldCommitmentsFee);
     oldCommitments.push(...validatedProvidedCommitments);
 
+    logger.debug({
```
israelboudoux, line 129: I believe you forgot to remove the log that follows to that one.

israelboudoux (approved):
LGTM

PR #1302, opened 2022-12-12
fix deployer get circuits in some cases

daveroga:
In some cases the deployer gets stucked checking the circuits and downloading them from the worker.
It has been detected that the function that sometimes is blocking the `deployer` is `getVerificationKeyByCircuitPath` from the `worker` in `filling.mjs` in the line when the verification key is exported with snarkjs (`snarkjs.zKey.exportVerificationKey`).

```
export const getVerificationKeyByCircuitPath = async circuitPath => {
  const vkPath = `${outputPath}/${circuitPath}/${path.basename(circuitPath)}.zkey`;
  if (fs.existsSync(vkPath)) {
    const vk = await snarkjs.zKey.exportVerificationKey(vkPath);
    return vk;
  }

  logger.warn({ msg: 'Unable to locate file', vkPath });

  return null;
};
```
If we don't do all the calls in parallel to get the verification keys of the circuits with `promise.all` and we do them sequentially then it's working fine.

No

./bin/start-nightfall

## Any other comments?
No

### DIFF
--- a/worker/src/utils/filing.mjs
+++ b/worker/src/utils/filing.mjs
@@ -13,16 +13,8 @@ export const getVerificationKeyByCircuitPath = async circuitPath => {
   if (fs.existsSync(vkPath)) {
     logger.debug(`Exporting verification key with snarkjs for ${vkPath}`);
 
-    let vk;
-    try {
-      vk = await snarkjs.zKey.exportVerificationKey(vkPath);
-      logger.debug(`Exported verification key for ${vkPath}`);
-    } catch (error) {
-      logger.error({
-        msg: 'Error snarkjs exportVerificationKey',
-        error,
-      });
-    }
+    const vk = await snarkjs.zKey.exportVerificationKey(vkPath);
+    logger.debug(`Exported verification key for ${vkPath}`);
     return vk;
   }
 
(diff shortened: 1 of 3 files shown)


worker/src/utils/filing.mjs
```diff
@@ -11,7 +11,18 @@ const outputPath = `/app/output`;
...
   if (fs.existsSync(vkPath)) {
-    const vk = await snarkjs.zKey.exportVerificationKey(vkPath);
+    logger.debug(`Exporting verification key with snarkjs for ${vkPath}`);
+
+    let vk;
+    try {
+      vk = await snarkjs.zKey.exportVerificationKey(vkPath);
+      logger.debug(`Exported verification key for ${vkPath}`);
+    } catch (error) {
+      logger.error({
```
israelboudoux, line 21: Hey my friend, if an error happens the exception isn't re-throwed and the `vk` variable isn't set to anything. Is that expected?

israelboudoux (approved):
There is just a comment to verify!

imagobea (approved):
Hello @daveroga I think there's no need to add the logs in `vk` routes. Both request and response are logged by default. So `folderpath` should be there already. It's approved anyways, cheers! :raised_hands:

PR #1331, opened 2022-12-19
AWS deployment

druiz0992:
Updates for AWS deployment
- Proposer periodically checks if its part of registered set and re-registers it not
- Add try/catch in nf3 challenger emitter
- Add X509 config to all network types
- Dev mode in hosted-utils

No

### DIFF
--- a/apps/proposer/src/proposer.mjs
+++ b/apps/proposer/src/proposer.mjs
@@ -4,6 +4,7 @@
 Module that runs up as a proposer
 */
 import logger from '@polygon-nightfall/common-files/utils/logger.mjs';
+import { waitForTimeout } from '@polygon-nightfall/common-files/utils/utils.mjs';
 import config from 'config';
 
 const { TIMER_CHANGE_PROPOSER_SECOND, MAX_ROTATE_TIMES } = config;
@@ -50,10 +51,10 @@ async function checkAndChangeProposer(nf3) {
           await nf3.changeCurrentProposer();
         }
       } catch (err) {
-        logger.info(err);
+        logger.error(err);
       }
     }
-    await new Promise(resolve => setTimeout(resolve, TIMER_CHANGE_PROPOSER_SECOND * 1000));
+    await waitForTimeout(TIMER_CHANGE_PROPOSER_SECOND * 1000);
   }
 }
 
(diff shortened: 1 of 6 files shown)


daveroga (approved):

apps/proposer/src/proposer.mjs
```diff
@@ -56,6 +57,32 @@ async function checkAndChangeProposer(nf3) {
...
+      logger.info('Attempting to register proposer');
+      const blockStake = await nf3.getBlockStake();
+      const minimumStake = await nf3.getMinimumStake();
+
+      logger.info(`blockStake: ${blockStake}, minimumStake: ${minimumStake}`);
+
+      try {
+        await nf3.registerProposer(proposerBaseUrl, minimumStake);
+      } catch (err) {
+        logger.info(err);
```
israelboudoux, line 78: Please replace with `logger.error`

apps/proposer/src/proposer.mjs
```diff
@@ -56,6 +57,32 @@ async function checkAndChangeProposer(nf3) {
...
+      }
+    }
+
+    await new Promise(resolve => setTimeout(resolve, CHECK_REGISTER_PROPOSER_SECOND * 1000));
```
israelboudoux, line 82: The function `waitForTimeout` from `common-files/utils/utils.mjs` can be used here.

cli/lib/nf3.mjs
```diff
@@ -1298,7 +1298,19 @@ class Nf3 {
...
           } catch (err) {
-            challengeEmitter.emit('error', err, type, txSelector);
+            logger.error({
+              msg: 'Error while trying to challenge a block',
+              type,
+              err,
+            });
+            try {
+              challengeEmitter.emit('error', err, type, txSelector);
+            } catch (err2) {
```
israelboudoux, line 1308: I'm not sure if this will work, when one emits an `error`, the Emitter manager will try to verify if there is any listener for `error`, if not it will log the error and will stop the process - it doesn't throw an error. The best way would be to use the method `createEmitter` when creating the `challengerEmitter` (similar to the proposer), then the server would not stop if the emitter emits an `error`.

israelboudoux (changes requested):

apps/proposer/src/proposer.mjs
druiz0992: done

israelboudoux (approved):

PR #1341, opened 2023-01-04
invert x509 logic to be on by default

### LINKED ISSUE #1278: Invert whitelisting control logic.
Currently whitelisting checks are off by default. This is convenient for testing where we don't want to have whitelisting enabled but would be more secure if whitelisting was on by default (guards against accidental deployment without whitelisting).

Westlad:

Currently, the requirement to have a valid X509 document signing certificate is off by default.  This change turns it on by default so that a deployer has to actively override this security feature.

fixes #1278 

You can use the integrations tests running locally to test this change.  They should now *fail* unless you actively turn of whitelisting by setting the `WHITELISTING` environment variable to `disable`:

```sh
export WHITELISTING=disable
```
The reason for this failure is that the legacy integration tests (with the exception of the x509 tests) do not support whitelisting.  We may change that in future, but it would delay this PR, which should be incorporated pre-launch.

## Any other comments?

This PR also fixes a bug in the erc20 integration test whereby an evergreen test was passing and should have failed. A small change to the test script and a small change to the code (to fix the actually failing code) has been made.


daveroga (approved):

docker/docker-compose.dev.yml
```diff
@@ -69,6 +69,7 @@ services:
         target: /app/entrypoint.sh
     environment:
       TEST_SANCTIONS_CONTRACT: 'mock'
+      WHITELISTING: ${WHITELISTING}
```
israelboudoux, line 72: It is better to move this env var to the `docker/docker-compose.yml` file, since this one is exclusively for tests/dev.

israelboudoux (changes requested):
There is just a little change to do.

docker/docker-compose.dev.yml
Westlad: Normally I would agree, however, that will make code that defaults to WHITELISTING=disable and somewhat invalidates the point of this change, which is to ensure that if people turn of this important security feature, they do so knowingly.

Westlad: As discussed.  `WHITELISTING` is now disabled in dev mode so that people don't need to remember to set then environment variable.

israelboudoux (approved):

Issue #1237, opened 2022-11-19
PoS/Decentralization


druiz0992:
- [x] Initial configuration PoS. #1248
- [x] Change proposer added to proposer so that proposer is added automatically. #1249
- [x] Estimate the cost the proposer has to charge in fees based in changeCurrentProposer gas cost or other related values. #1257
- [x] proposers claim fees periodically
- [x] proposers set a different minimum fee
- [x] challengers claim rewards periodically
- [ ] Other

PR #1354, opened 2023-01-16
periodic payment cron job for proposer and challenger

LijuJoseJJ:
This PR implement a feature to `cli` SDK for proposer and challenger to start a cron job which withdraw their pending withdraw from state contract.
as part of the implementation it fixes few existing bugs:-
1.  fix state contract addPendingWithdrawal function - not to override existing added pendingWithdraw for an address
2. update `nightfall_deployer` migration files to do `state.setFeeL2TokenAddress` function
3. fix optimist router.get('/pending-payments') api to work with address not passed in lower case. As in our optimist DB we save EOA address in lowercase unlike  in ethereum where address in case sensitive
4. fix optimist router.get('/withdraw') api:- to  router.post('/withdraw') also correct logic to call withdraw function from state contract, before api had logic to call non-existing withdraw from proposer contract
5. fix optimist router.post('/payment')

https://github.com/EYBlockchain/nightfall_3/issues/1237

### DIFF
--- a/.github/workflows/on-pull-request-master.yml
+++ b/.github/workflows/on-pull-request-master.yml
@@ -722,6 +722,19 @@ jobs:
       - name: Build adversary
         run: npm run build-adversary
 
+      - name: Perform npm link
+        run: |
+          cd common-files
+          npm link
+          cd ..
+          npm link @polygon-nightfall/common-files
+          cd cli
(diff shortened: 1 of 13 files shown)


Westlad (approved):

Westlad:
Can you clear the conflict please?

PR #1360, opened 2023-01-24
Adding authentication layer

israelboudoux:
This authentication layer is something good to have that can be used by any of the microservices, it just needed to set the `AUTHENTICATION_KEY` environment variable with the API Key desired, which will be used to authenticate requests against the `X-API-Key` header. 
The env var `ENDPOINTS_WHITELISTED` can be used to whitelist endpoints - it may contain a comma-separated-list of endpoints and is able to accept regex. 
At the moment this capability is configured only for `nightfall-client`, since it is likely that a deployer/enterprise wants to use this layer to protect the application endpoints.

This change implies changes in SDK so that the calls can be authenticated.

No

A new test was added to Github Action

## Any other comments?
No


common-files/utils/httputils.mjs
```diff
@@ -234,18 +236,59 @@ const responseLogger = (req, res, next) => {
...
   routesDefiner,
   addHealthCheck = true,
   useFileUpload = true,
 ) => {
+  app.use(requestLogger);
+
+  if (process.env.AUTHENTICATION_KEY) {
+    app.set(
+      ENDPOINTS_WHITELISTED,
+      (process.env.ENDPOINTS_WHITELISTED ?? '').split(',').flatMap(v => v.trim()),
```
pawelgrzybek, line 287: Regular `.map` should be sufficient. I cannot imagine someone passing there a string that will result in nested arrays.

pawelgrzybek (changes requested):

common-files/utils/httputils.mjs
```diff
@@ -234,18 +236,59 @@ const responseLogger = (req, res, next) => {
...
 
+const isEndpointWhitelisted = req => {
+  let result = false;
+  req.app.get(ENDPOINTS_WHITELISTED).forEach(e => {
+    if (!result) {
+      result = req.url.match(e) != null;
+    }
+  });
+
+  return result;
```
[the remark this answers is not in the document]
israelboudoux: Sorry, but this doesn't work in this case because the value of `ENDPOINTS_WHITELISTED` accepts regular expressions.

common-files/utils/httputils.mjs
israelboudoux: Done!

common-files/utils/httputils.mjs
```diff
@@ -234,18 +236,59 @@ const responseLogger = (req, res, next) => {
...
+};
+
+const authenticationHandler = (req, res, next) => {
+  if (
+    req.url === '/healthcheck' ||
```
pawelgrzybek, line 252: Same question here, what is someone makes a requests to `/healthcheck/`. This should be handled by `new URL` or `req.url.startsWith('/healthcheck')`.

common-files/utils/httputils.mjs
```diff
@@ -234,18 +236,59 @@ const responseLogger = (req, res, next) => {
...
+ * the environment variable AUTHENTICATION_KEY is set: this is the key that will be used to authenticate requests
+ * against the request header 'X-API-Key'. Endpoints can be whitelisted by using the environment variable
+ * ENDPOINTS_WHITELISTED listing the whitelisted endpoints (e.g. ENDPOINTS_WHITELISTED="/commitment/save, /commitment/delete").
+ * The env var ENDPOINTS_WHITELISTED accepts regex expression values.
  *
- * @param {*} app - an instance of 'expressjs'.
+ * @param {*} app - an instance of 'expressjs'
+ * @param routesDefiner - an annonymous function that receives an expressjs instance and defines the routes
+ * @param addHealthCheck - adds a /healthcheck endpoint automatically
+ * @param useFileUpload - adds a middleware for file uploading
```
pawelgrzybek, line 274: Since we are here can we add types to this function? Like so…

```
 * @param {*} app - an instance of 'expressjs'
  * @param {Function} routesDefiner - an annonymous function that receives an expressjs instance and defines the routes
  * @param {boolean} addHealthCheck - adds a /healthcheck endpoint automatically
  * @param {boolean} useFileUpload - adds a middleware for file uploading
```

israelboudoux: Good one!

test/e2e/nf3lib/nf3.test.mjs
```diff
@@ -0,0 +1,39 @@
...
+
+const { mnemonics, signingKeys } = config.TEST_OPTIONS;
+
+describe('Nf3 Test', () => {
+  describe('Client Authenticated calls tests', () => {
+    it('Should fail since no authentication key is set', async () => {
+      const nf3User = new Nf3(signingKeys.user1, environment);
+      try {
+        await nf3User.init(mnemonics.user1);
+        fail();
```
imagobea, line 23: I think we should add some explanation at the top of this file. Because I don't think this test suite makes much sense outside the Github Action test `client-authentication-test` context. So one could simply bump into it and be like...

image

If you run the suite via `npm run test-e2e-nf3lib`, this particular test fails but not because of the described reason, because `fail()` will raise an `Assertion` error.

imagobea (changes requested):
Hello @israelboudoux 

Interesting feature indeed.

Requesting changes because of the `test/e2e/nf3lib/nf3.test.mjs` file. I think we should add a comment, providing some context of what we are trying to demonstrate (outside the GitHub Action the tests pass but what's stated isn't really true).

Similarly, I was wondering why do we need a 3rd env var, especific to nf3 (`CLIENT_AUTHENTICATION_KEY`).

Happy to discuss any of it, just let me know! :hugs: 
Cheers

test/e2e/nf3lib/nf3.test.mjs
imagobea: Just to be clear
Test is `Should fail since no authentication key is set`
This passes, ie it fails, but due to `fail()` raising exception :see_no_evil:

bin/setup-nightfall
```diff
@@ -2,8 +2,10 @@
 
 set -e
 
-# Install node dependencies
-npm ci
+if [ "$SKIP_NPM_CI" == "" ]; then
```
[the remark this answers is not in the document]
israelboudoux: Yes, just for us, developers, when we want to speed up things locally we can use them.

docker/docker-compose.yml
```diff
@@ -33,11 +33,12 @@ services:
     ports:
       - 8080:80
     depends_on:
-      - deployer
       - worker
       - rabbitmq
       - mongodb
     environment:
+      AUTHENTICATION_KEY: ${AUTHENTICATION_KEY}
+      ENDPOINTS_WHITELISTED: ${ENDPOINTS_WHITELISTED}
```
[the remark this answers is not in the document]
israelboudoux: Good catch, I'll add it to the deployment. The authentication is not required, which is the reason why there isn't a fallback value. It is for the flavor of the deployer, if he wants to enable it, he supplies a value.

.github/workflows/on-pull-request-master.yml
```diff
@@ -96,6 +96,106 @@ jobs:
...
+          fi
+
+      - name: Authentication Test - non-whitelisted endpoint OK
+        run: |
+          RESULT=$(curl -w "%{http_code}" -o /dev/null --silent --header "X-Api-Key: $(echo $AUTHENTICATION_KEY)" http://localhost:8080/commitment/)
+          if [ $RESULT != "200"]; then
+            exit 1
+          fi
+
+      - name: Authentication Test - non-whitelisted endpoint failure
```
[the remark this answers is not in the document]
israelboudoux: Done!

.github/workflows/on-pull-request-master.yml
```diff
@@ -96,6 +96,106 @@ jobs:
...
+          command: |
+            docker wait nightfall_3_deployer_1
+          attempt_limit: 100
+          attempt_delay: 20000
+
+      - name: Debug logs - after image builds
+        if: always()
+        run: cat authentication-test.log
+
+      - name: Authentication Test - heathcheck OK
```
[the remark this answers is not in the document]
israelboudoux: Done!

test/e2e/nf3lib/nf3.test.mjs
daveroga: Yes, I agree with @imagobea. This is failing because of fail(). We need better explanation of this test.

test/e2e/nf3lib/nf3.test.mjs
```diff
@@ -0,0 +1,39 @@
...
+      } catch (err) {}
+    });
+
+    it('Should pass since the authentication key is set', async () => {
+      process.env.CLIENT_AUTHENTICATION_KEY = process.env.AUTHENTICATION_KEY;
```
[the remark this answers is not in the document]
daveroga: I suppose if you are not passing `CLIENT_AUTHENTICATION_KEY` the test won't fail either, so I don't understand the test.

daveroga (changes requested):

bin/setup-nightfall
```diff
@@ -13,7 +15,9 @@ if [ $OS_ARCH != "x86_64" ]; then
   NO_CACHE_FLAG='--no-cache'
 fi
 
-docker build ${NO_CACHE_FLAG} -t ghcr.io/eyblockchain/local-circom -f docker/circom.Dockerfile .
+if [ "$SKIP_CIRCOM_BUILD" == "" ]; then
```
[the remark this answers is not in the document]
israelboudoux: This is for local development only. Have you wondered how many times you run `setupNightfall` and it goes thorough all the process unnecessarily? That is it, for us testing things locally we can use this to be faster when developing.

test/e2e/nf3lib/nf3.test.mjs
israelboudoux: Yes, you're right. The tests are not well designed and miss some sort of documentation. Working on it! ;)

israelboudoux: Yes, you're right. The tests are not well designed and miss some sort of documentation. Working on this! ;)

israelboudoux: Yes, you're right. The tests are not well designed and miss some sort of documentation. Working on this! ;)

imagobea: I don't know what's going on on GitHub these days... Comments etc appear a gazillion times :rofl:

imagobea (approved):
LGTM!
Just one more thing about `nf3.test.mjs` : we could close the user instance at the end of each test with   `await nf3User.close()`.
:raised_hands:

test/e2e/nf3lib/nf3.test.mjs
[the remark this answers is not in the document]
israelboudoux: I have re-designed the test so that this env var is not needed anymore.

bin/setup-nightfall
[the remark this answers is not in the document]
daveroga: I rather do setup-nightfall because when I remove images I don't remove local-circom. But it's ok for development.

daveroga:
> LGTM! Just one more thing about `nf3.test.mjs` : we could close the user instance at the end of each test with `await nf3User.close()`. raised_hands

I agree with @imagobea

daveroga (approved):

cli/lib/nf3.mjs
```diff
@@ -111,6 +116,29 @@ class Nf3 {
...
     this.zkpKeys = zkpKeys;
     this.currentEnvironment = environment;
+    this.clientAuthenticationKey = clientApiAuthenticationKey;
+
+    this.applyHttpClientAuthentication();
+  }
+
+  applyHttpClientAuthentication() {
+    if (!this.clientAuthenticationKey) {
+      logger.info('No client authentication key is set!');
```
pawelgrzybek, line 126: Don't we want to throw at this point, and use `console.error` at this point?

common-files/utils/httputils.mjs
```diff
@@ -109,9 +111,9 @@ const applyAxiosDefaults = () => {
 const getRequestParam = req => {
   const dataInput = {};
 
-  if (Object.keys(req.query).length !== 0) dataInput.query = req.query;
-  if (Object.keys(req.params).length !== 0) dataInput.params = req.params;
-  if (Object.keys(req.body).length !== 0) dataInput.body = req.body;
+  if (req.query && Object.keys(req.query).length !== 0) dataInput.query = req.query;
```
pawelgrzybek, line 114:
```
if (Object.keys(req?.query ?? []).length !== 0)
```

…both do the same. Just a simplification that easily can be ignored.

common-files/utils/httputils.mjs
```diff
@@ -234,18 +236,70 @@ const responseLogger = (req, res, next) => {
...
+const authenticationHandler = (req, res, next) => {
+  if (
+    req.url.match(HEALTH_CHECK_REGEX) != null ||
+    isEndpointWhitelisted(req) ||
+    req.get('x-api-key') === process.env.AUTHENTICATION_KEY
+  ) {
+    return next();
+  }
+
+  logger.warn('Unauthorized access!');
```
pawelgrzybek, line 260: Shouldn't that be en error?

cli/lib/nf3.mjs
israelboudoux: No, because this feature is not required to be setup. It will depend on the deployer's wish.

common-files/utils/httputils.mjs
israelboudoux: I honestly don't know. I thought that reporting an error would raise a concern that was addressed smoothly by the stack.

common-files/utils/httputils.mjs
israelboudoux: I'll do this in another moment.

PR #1362, opened 2023-01-30
fix: on-chain root calculation

NadavPeled1998:

- Fix calculate root in MerkleTree_Stateless library:
  
    1. Use leafCount instead of lastLeafIndex. This parameter is from Utils.getLeafCount(blockL2.packedInfo), and getLeafCount returns the leaf count and not the last leaf index.
    2. Init node index according to the leaf count and the frontier slot.

- Test MerkleTree_Stateless library. I created tests for calculateRoot and updateFrontier (I didn't find any bugs in this function). I created MerkleTree_StatelessMock so it will be possible to test this library directly.

As far as I know, there are no open issues on this bug.

`npm run unit-test-merkle_tree` or `npm run unit-test`

### DIFF
--- a/nightfall-deployer/contracts/MerkleTree_Stateless.sol
+++ b/nightfall-deployer/contracts/MerkleTree_Stateless.sol
@@ -59,11 +59,11 @@ library MerkleTree_Stateless {
     /**
     @notice Insert multiple leaves into the Merkle Tree, and then update the root, The user must update and persistently stored the new frontier.
     @param _frontier - the current Frontier value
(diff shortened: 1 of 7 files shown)


fullkomnun:
This issue was probably introduced as part of the [following](https://github.com/EYBlockchain/nightfall_3/commit/577cd3005554d7e57b37375102a83b1734fc3897) commit that converted Solidity to inline assembly.

israelboudoux (approved):

PR #1366, opened 2023-02-02
feat: client can inform of l2 tx status for given l2 tx hash

imagobea:

Adds a new endpoint to nightfall-client `transaction/status/:l2TransactionHash`.

The endpoint will first look up the transaction `l2TransactionHash` in the client database - there was already a db method for this.

If the transaction is not found, it will then ask the proposers. For this:
- Have updated `common-files` with an existing function that gets all the registered proposers from the contracts
- Have added a new endpoint to the nightfall-optimist `proposer/mempool/:l2TransactionHash` (and developed + refactored some `mempool` db services`
- Have created an `nf3` method to query the endpoint above from the proposer app (where there's also a new endpoint `proposer/mempool/:l2TransactionHash`)

The client asks all the proposers using `Promise.any()`, which `resolves` as soon as any requests `resolves`. Returning the `transaction`. Otherwise, it raises `404`.

Last but not least, and to honour the name of the new endpoint `status`, the client will return both a string indicating the status `mempool | mined`, plus the block no `blockNumberL2`. Unsuccessful requests can return `404` or `400`.

This dev will be used during part II of the sdk issue [Proof of transaction](https://github.com/NightfallRollup/nightfall-sdk/issues/6)

Take a look at the new tests (yes, the proposer suite contains some refactors, sorry!)

Also, you may `curl` the endpoints. Locally, the apps can be found at the following ports. You can run nightfall, and use the sdk scripts to perform transactions.

- client `8080`
- optimist `8081`
- proposer `8092`

## Any other comments?

**In the overall, the trickiest aspect was to return a consistent response from the 3 APIs that the request may go through**. For the same reason, the client endpoint might not be very "performant" in the event of having a large proposer network.

TODO

- [x] Refactor client endpoint to connect with new proposer logic
- [x] Refactor client tests
- [x] Check pipelines + self-review
- [ ] Add new test suite to GitHub actions


nightfall-client/src/services/transaction.mjs
```diff
@@ -0,0 +1,51 @@
...
+
+const STATUS_MINED = 'mined';
+const STATUS_MEMPOOL = 'mempool';
+
+export async function findTransactionInMempools(l2TransactionHash) {
+  logger.debug('Get all registered proposer URLs from State contract..');
+  const proposerURLs = (await getProposers()).map(p => p.url);
+
+  logger.debug('Query each proposer mempool..');
+  const promises = [];
```
pawelgrzybek, line 17:
```js
const promises = proposerURLs.map((pURL) =>
  axios.get(`${pURL}/proposer/mempool/${l2TransactionHash}`)
);
```

pawelgrzybek: And we cam scrap forEach.

nightfall-client/src/services/transaction.mjs
```diff
@@ -0,0 +1,51 @@
...
+  proposerURLs.forEach(pURL => {
+    promises.push(axios.get(`${pURL}/proposer/mempool/${l2TransactionHash}`));
+  });
+
+  /*
+    The user may have sent the transaction multiple times to different proposers,
+    but this situation should be resolved when including the tx in L2 block
+  */
+  let res;
```
pawelgrzybek, line 26: No need for `res` reassignment.

```js
try {
  const res = await Promise.any(promises);
  return res.data;
} catch (err) {
  logger.trace(err);
  throw new NotFoundError(
    `Could not find L2 transaction hash ${l2TransactionHash}`
  );
}
```

nightfall-client/src/services/transaction.mjs
```diff
@@ -0,0 +1,51 @@
...
+  const { blockNumberL2 } = transaction;
+  logger.debug(`Set status for L2 transaction with L2 block number ${blockNumberL2}..`);
+
+  let status;
+  if (blockNumberL2 >= 0) {
+    status = STATUS_MINED;
+  } else if (blockNumberL2 === -1) {
+    status = STATUS_MEMPOOL;
+  } else {
+    throw new ValidationError(`Incorrect L2 block number ${blockNumberL2}`);
```
pawelgrzybek, line 47: Can we ever end up in that situation? It returns a positive integer when minder, and `-1` when not. I am curious how we can end-up getting anything different here 🤔 ?

nightfall-optimist/src/services/database.mjs
```diff
@@ -366,6 +376,44 @@ export async function getMempoolTxsSortedByFee() {
     .toArray();
 }
 
+/**
+ * Find a transaction in the mempool, given some filters
+ */
+async function getMempoolTransaction(query) {
```
pawelgrzybek, line 382: I am wondering if passing a whole `query` as a parem is not too generic. Maybe we can make more like for-the-purpose abstractions here?

nightfall-client/src/services/transaction.mjs
daveroga: I agree. It's cleaner with map.

nightfall-client/src/services/transaction.mjs
daveroga: The same suggestion.

nightfall-optimist/src/services/database.mjs
daveroga: Here the query means filter I think. You can pass all the filters needed by the mempool transaction you want to query. I see it's reused in other queries for a mempool transaction.

daveroga (commented):
Good work @imagobea :). Only some comments.

nightfall-client/src/services/transaction.mjs
imagobea: Thanks! Very clever, I'm actually annoyed that it didn't occur to me :nerd_face: I really enjoy array methods, must have a sad life...

nightfall-optimist/src/services/database.mjs
imagobea: That's right, have used `query` as a way to pass a bunch of filters for mempool-specific queries.

I chose `query` over `filters` to be consistent with the rest of the file in terms of language.

Agree we can do much better... I was just trying to remove some of the repetition without touching too much code.

imagobea:
Morning @pawelgrzybek @daveroga , I think I have addressed your comments. Pls let me know if I missed anything :pray: thanks for your time!!

daveroga (approved):

pawelgrzybek (approved):

Issue #1367, opened 2023-02-02
[BUG] FinaliseWithdrawal fails if there is more than 1 transaction in the block including the withdrawal


druiz0992:
**Describe the bug**
If a block contains multiple transactions, one of them being a withdrawal, this withdrawal cannot be finalized.

**Expected behavior**
Finalize withdrawal should be successful provided the transactionHash is correct

**To Reproduce**
Just generate several transactions in a block, including a withdrawal. Then, attempt to finalize it.

PR #1368, opened 2023-02-02
withdrawal with multiple transactions in same block that are not in a single client's database can't be finalized

druiz0992:
Allows to finalize a withdrawal included in a block with other transactions

#1367


israelboudoux (approved):

nightfall-client/src/services/finalise-withdrawal.mjs
```diff
@@ -16,7 +16,9 @@ export async function finaliseWithdrawal(transactionHash) {
   const transactions = await Promise.all(
     block.transactionHashes.map(t => getTransactionByTransactionHash(t)),
   );
-  const index = transactions.findIndex(f => f.transactionHash === transactionHash);
+
+  // block can contain multiple transactions
+  const index = transactions.findIndex(f => transactionHash === f?.transactionHash ?? '0x0');
```
imagobea, line 21: I tried to reproduce the bug locally but couldn't: I made a few transfers and a withdrawal > made block > finalised withdrawal successfully.

By looking at the proposed fix, I suspect we are trying to resolve another problem: you cannot access `transactionHash` in `f` if such `f` is null.

The array `transactions` will contain a `null` element if you do not have a particular transaction in the `transactions` collection - to get this far in the service, it means that you have the block in `blocks` collection but this block contains transactions from other clients (not only your transaction).

So the bug is that a withdrawal can't be finalised when a block contains transactions that are not in your database :nerd_face:

test/e2e/tokens/erc20.test.mjs
```diff
@@ -305,6 +306,9 @@ describe('ERC20 tests', () => {
       await makeBlock();
 
       userL2BalanceBefore = await getLayer2Balances(nf3User, erc20Address);
+      // Add two transactions in same block to check that finalize withdrawal works
+      // https://github.com/EYBlockchain/nightfall_3/issues/1367
+      await nf3User.deposit(erc20Address, tokenType, transferValue, tokenId, fee);
```
imagobea, line 311: If the bug is what I framed above, this is not really doing much as transactions are originated from the same client. In which case I will leave the test as it was without further changes.

imagobea (changes requested):
Not asking for particular changes, just for some clarification. Thank you!! :pray:

nightfall-client/src/services/finalise-withdrawal.mjs
druiz0992: thanks @imagobea. You are correct in your analysis. I will modify the test to check for this condition

nightfall-client/src/services/finalise-withdrawal.mjs
```diff
@@ -16,7 +16,9 @@ export async function finaliseWithdrawal(transactionHash) {
   const transactions = await Promise.all(
     block.transactionHashes.map(t => getTransactionByTransactionHash(t)),
   );
-  const index = transactions.findIndex(f => f.transactionHash === transactionHash);
+
+  // block can contain multiple transactions not in our database
+  const index = transactions.findIndex(f => transactionHash === f?.transactionHash ?? '0x0');
```
Westlad, line 21: Is there particular need for the nullish coalescer `?? '0x0'` ?  If `f` is `null` because it wasn't in the database then `f?.transactionHash` will be `undefined` and hence the test function `transactionHash === f?.transactionHash` will return false anyway. It's not harmful in anycase so happy to approve but just curious.

Westlad (approved):

nightfall-client/src/services/finalise-withdrawal.mjs
imagobea: Accessing attributes of `null` causes a `TypeError`.

imagobea (approved):

PR #1370, opened 2023-02-02
Save tx at transaction submit eventHandler for nightfall-client

### LINKED ISSUE #1335: [FEATURE] Client being able to re-use nullifiers that aren't still in a block
Currently, users do not have any mechanism to create a new transaction that contains an already used nullifier that isn't still in a block. One of the main usages of this feature would be allowing clients to re-send a transaction with a higher fee.
In order to implement this, two changes are needed:
- If the user provides the commitments (for either value or fee), we simply need to check in the code that the commitment is not already nullified on chain instead of checking `isNullified`. Note: If no commitments are provided, our default algorithm will never use nullified commitments.

- Database will need to be modified so that it is consistent with what is happening. This means that don't want to store obsolete commitments nor transactions. Also, we need to keep in mind that, even if the client sends a new transaction using the same nullifier, he doesn't know which transaction will be the one that will get into the blockchain. The idea here is that the user will save the transaction when submitted rather than when a block is proposed. And then, when the block is proposed, will search for transactions that was using the same commitment / nullifier and delete from both, commitments and transactions db.

LijuJoseJJ:

Changes in this PR includes
1. implement transactionSubmittedEventHandler for nightfall-client
2.  save transaction in transactionSubmittedEventHandler if any of commitments and nullifiers are store in db for nightfall-client.
3.  update blockProposedEventHandler of nightfall-client to delete duplicate transactions and their commitments from db, if the transaction not in L2 i.e deleted commitments are not nullified on chain yet.

This PR fixes #1335 

## Any other comments?
This PR also implements github workflow with test file`test/client-resync.test.mjs`, where user created two transfers transaction using same set of commitments, while proposer setMakeBlock was off. *Note: second transfer is with higher fee*.
then proposer starts making block pick second tx and make a successful block.
[to test client re-sync feature] we are restarts client and all unsaved transaction are saved at transactionSubmitEventHandler
and at  blockProposedEventHandler second tx updates and first tx deleted with change commitment it created. 

### explanation on  `test/client-resync.test.mjs`
steps are.
1.  do two deposit one by one each deposit test wait for `TransactionSubmitted` event and expect respective transaction hash and saved in client's db.
2. create first transfer transaction, we also make sure it got stored in client database by waiting for TransactionSubmitted and after that querying client transaction collection via api.
3.  create second transfer transaction but this time we do not wait for `TransactionSubmitted` event also do not do `makeBlock', and restart client immediately`await restartClient(nf3User);` after client re-sync complete, check both transfer transactions in client db (this done at test by calling API `GET: /transactions`).
4. after client re-sync and above check, start proposer make block `await makeBlock();`.
5. At last check user balance also check user(client) transaction collections
     ( only three transaction should be in db, and last one should be of higher fee)

### DIFF
--- a/nightfall-client/src/event-handlers/transaction-submitted.mjs
+++ b/nightfall-client/src/event-handlers/transaction-submitted.mjs
@@ -44,7 +44,12 @@ async function transactionSubmittedEventHandler(eventParams) {
   }
 
   if (saveTxInDb) {
-    await saveTransaction({ ...transaction });
+    await saveTransaction({ ...transaction }).catch(err =>
+      logger.error({
+        msg: 'error while saving transaction in transactionSubmittedEventHandler',
+        err,
+      }),
+    );
   }
 
   logger.info({
--- a/nightfall-client/src/services/database.mjs
+++ b/nightfall-client/src/services/database.mjs
@@ -315,7 +315,7 @@ export async function findDuplicateTransactions(commitments, nullifiers, transac
   const query = {
     $or: [{ commitments: { $in: commitments } }, { nullifiers: { $in: nullifiers } }],
     transactionHash: { $nin: transactionHashes },
-    blockNumberL2: { $exists: false },
+    blockNumberL2: { $eq: -1 },
   };
   return db.collection(TRANSACTIONS_COLLECTION).find(query).toArray();
 }


Westlad (changes requested):
Please add a test for your client synchronisation code (like the optimist resync test)

LijuJoseJJ:
> Please add a test for your client synchronisation code (like the optimist resync test)

added

Westlad (changes requested):
Please explain how to test this code (including with a re-sync of the client).

Westlad (approved):

PR #1375, opened 2023-02-06
[Regulator] Add Regulator Service

eugypalu:

- [x] Add REGULATOR_PRIVATE_KEY as parameter for the container to configure a private key for the Regulator Service.
- [x] Add endpoint registerPairSenderReceiver to receive client calls with { PKa, PKb, PKx, pkx * PKb } parameters and result of scalar mult { pkr * PKb }. This registration will store parameters in some collection transfer_secrets.
- [x] Update the client service to be able to decrypt commitments for the regulator using the previous information and the commitments from transactions. Use another commitment collection commitments_regulator in the database for the regulator commitments.
- [ ] Add the corresponding tests

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -107,16 +107,17 @@ module.exports = {
     APPROVERS: process.env.MULTISIG_APPROVERS
       ? process.env.MULTISIG_APPROVERS.split(',')
       : [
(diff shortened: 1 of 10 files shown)


daveroga:
Moved to private repo

PR #1386, opened 2023-02-10
Westlad/nightfall node

Westlad:

This PR adds some extra configuration and instructions for running up a Nightfall node against the deployment on the Mumbai testnet.

It also tidies the `_submitTransaction` method in `Nf3` and adds logging of the proposer and challenger containers to the CI log output.


cli/lib/nf3.mjs
```diff
@@ -415,6 +417,13 @@ class Nf3 {
     return this.web3.eth.sendTransaction(tx);
   }
 
+  async _sendTransactionV2(tx) {
```
pawelgrzybek, line 420: Is there a good reason of making it `async`? None of the operations in inner scope pause the execution using `await`. Is just just for the sake of this being a Promise?

Westlad: Yes, I think you are correct, there's nothing being awaited so you could just return, and it doesn't matter if you are returning a promise or not. I was just copying the signature from `_sendTransaction` to make it a drop in replacement. I was testing if `_sendTransactionV2` is functionally the same as `_sendTransaction`, if it is that makes `sendTransaction` needlessly complex.
Either way I'll tidy up the code and remove the old function.

PR #1387, opened 2023-02-10
docs: deprecating some old and somewhat inaccurate docs

Westlad:

Quite a bit of the documentation in the `doc` folder is out of date.  It's not completely useless but it's also no longer totally accurate. I've moved it to a `deprecated` folder within `doc` to flag this fact.  It can then be updated and promoted back to the `doc` folder, or removed at a later date.


israelboudoux (approved):

israelboudoux:
Hey @Westlad, I understand that change, but to be honest I think some effort should be put into improving the documentation in Nightfall. There are lots of README under the `doc` folder. However, they are totally disconnected which makes things bad when someone wants to find some information.

Westlad:
I agree!  All doc PRs are welcome.

IlyasRidhuan (approved):

PR #1408, opened 2023-03-16
fix: add salt to challenge

Westlad:

This is a bug fix.  The 'historic root block number not correct' challenge is not salted.  This means that if the challenge commitment is front-run, the original challenger would be locked out from sending a new commitment because they could not generate a new commit hash by changing their salt.

No

There is no change to the standard tests.

### DIFF
--- a/nightfall-deployer/contracts/Challenges.sol
+++ b/nightfall-deployer/contracts/Challenges.sol
@@ -357,7 +357,10 @@ contract Challenges is Stateful, Config {
         challengeAccepted(transaction.blockL2);
     }
 
-    function challengeHistoricRootBlockNumber(TransactionInfoBlock calldata transaction) external {
+    function challengeHistoricRootBlockNumber(
+        TransactionInfoBlock calldata transaction,
+        bytes32 salt
+    ) external {
         checkCommit(msg.data);
         state.areBlockAndTransactionReal(
             transaction.blockL2,
--- a/nightfall-optimist/src/services/challenges.mjs
+++ b/nightfall-optimist/src/services/challenges.mjs
@@ -319,12 +319,15 @@ export async function createChallenge(block, transactions, err) {
       }
 
       txDataToSign = await challengeContractInstance.methods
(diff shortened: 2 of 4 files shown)


nightfall-deployer/contracts/Challenges.sol
```diff
@@ -357,7 +357,10 @@ contract Challenges is Stateful, Config {
         challengeAccepted(transaction.blockL2);
     }
 
-    function challengeHistoricRootBlockNumber(TransactionInfoBlock calldata transaction) external {
+    function challengeHistoricRootBlockNumber(
+        TransactionInfoBlock calldata transaction,
+        bytes32 salt
```
israelboudoux, line 362: Hey @Westlad, you added the `salt` parameter, but it is not being used anywhere. Is that correct?

Westlad: Yes, it's correct.  It still appears in the calldata (i.e. `msg.data`), even though it's not used in the function, and that's what we require. The commit hash is the hash of `msg.data.  Having a salt in there lets you change the value of the commit hash by changing the salt. 
If it's not there then the commit hash can only have one value.  Then a front-runner can just take your challenge commitment.  They can't actually challenge themselves but they can prevent you from doing so because the commit hash gets registered under their address.  Having a salt lets you have more attempts.  You may still get front run but now the front-runner has to front-run every challenger multiple times, which is a much harder task.

Westlad: Unfortunately the compiler doesn't realise it's actually being used and so issues a warning.

israelboudoux: I see @Westlad , then having the salt as a parameter makes the salt to be _automatically_ applied?

Westlad: Yes, that's it exactly.

druiz0992 (approved):

israelboudoux (approved):

PR #1410, opened 2023-03-24
implements demo-ui

### LINKED ISSUE #1385: Lightweight demo UI
For enterprise usage, the Client container is deployed and is intended to integrate with local systems, using the nf3 class as a convenient node API (although direct http:// calls can also be used).

However, this functionality is hard to demonstrate.  It would be good to have a lightweight browser-based UI that can be used to call the deposit/transfer/withdraw endpoints in Client.

The Client UI should not store any significant state, and it should interface to the Client and a Proposer only, for simplicity.

LijuJoseJJ:
This PR create a small react app called demo-ui, intension is demo-ui will be used to test contracts deployed in mumbai testnet.
for now: it has been test against local ganache instead of mumbai testnet deployments
fixes #1385 

pending work:-
- this PR remove use of `find-unused-exports` to make demo-ui `npm run lint` work. Solution is to find better module than find-unused-exports to make it work for both node.js and react code.
- code refactor : Although demo-ui app is small and working, but need more react smart code compare what is in this PR.
- better CSS
To run demo-ui:-
  1. start nightfall
  2. cd demo-ui
  3. npm ci
  4. npm start


Westlad (approved):
Approved, for merge.  Details are being added to the Gitbook.

IlyasRidhuan (approved):

PR #1411, opened 2023-03-25
Revert "Configure solcover (Solidity Coverage)"

druiz0992:
Reverts EYBlockchain/nightfall_3#1395


druiz0992:
I reverted this change by mistake

PR #1412, opened 2023-03-27
Refactor offline transaction code

Westlad:

Code for sending a transaction 'directly' to an Optimist (off-chain transaction) is available both in the `Nf3` class and in the Client.  The code in the `Nf3` class only sends the transaction to the Optimist that has its URL set in `Nf3` whereas the Client code attempts to send the transaction to all Proposers (which proxy the request to their Optimist instance). There is no point in having similar code in two places, so it makes sense to refactor the code.  

It is normally optimal to send an off-chain transaction to all Proposers, to minimise censorship opportunities. This should be the default, therefore.
It is not optimal to require a Client to make a decision as to whether to send a computed transaction back to the User, in the case of an on-chain transaction, or to the Proposers in the case of an off-chain transaction. This is inflexible, because the User application may well want to make decisions about which Proposers to send its transaction to.  For example, an enterprise may wish to use a specific Proposer that it controls (even though this may reveal the transaction's origin), or optimise for lowest cost Proposers, or ones located in a specific country, or to have mutual authentication between the User and Proposer. Keeping the transaction routing concern in the User (albeit in the Nf3 class) keeps it in one place, and lets the Client focus on computing transactions; it just sends all transactions back to the User that requested them. Thus, to change how transactions are sent to the Proposers, we only need to customise our User application, which we'll be writing anyway, rather than change the Client code, which is not intended to be customised.

Thus, this refactor moves all off-chain transaction routing code into the User (specifically the `Nf3` class that the User will instantiate).

Normal tests suffice.


cli/lib/nf3.mjs
```diff
@@ -109,6 +109,8 @@ class Nf3 {
 
   clientAuthenticationKey;
 
+  NEXT_N_PROPOSERS = 3;
```
israelboudoux, line 112: This is the max number of Proposers we desire to send an off chain trx, right? If so, could we use a better name for this var?

cli/lib/nf3.mjs
```diff
@@ -1309,19 +1307,64 @@ class Nf3 {
...
+      proposerIdx += 1;
+    }
+
+    logger.debug({ msg: 'Peer List', peerList });
+    return Promise.any(
+      Object.keys(peerList).map(async address => {
+        logger.debug(
+          `offchain transaction - calling ${peerList[address]}/proposer/offchain-transaction`,
+        );
+        return axios.post(
```
israelboudoux, line 1361: I think you can use the function `forwardOffchainTransaction` here.

israelboudoux (approved):
LGTM, just some details to verify. This Optimist/Proposer references cause a bit of confusion, because you mean that the Optimist should be private/firewalled but in the end we're calling the Optimist and not the Proposer, it is the other way around I guess.

cli/lib/nf3.mjs
Westlad: Possibly yes but it's code that I just refactored (moved) from elsewhere, so I'd rather change this as a different PR because I haven't taken time to deeply understand how this code segment works and I don't really want to add that complication here.

cli/lib/nf3.mjs
Westlad: I don't think so because that would assume that the URL which is put on chain is the same as the URL of the Optimist.  This is not the case if you are proxying through the Proposer.

IlyasRidhuan (approved):

PR #1416, opened 2023-04-11
fix: fix loop index in transform circuit

RajeevS17:

The part fixed in the transform circuit checks that the commitment values do not overflow. The fix replaces commitmentsValues[0] by commitmentsValues[i] in line 108. As this is input in a loop, it make sense that the commitment values should be checked for every iteration. 

npm t


Westlad:
Closed as this change is incorporated in #1435

PR #1418, opened 2023-04-13
Westlad/polygon po s

Westlad:

This branch contains extra configuration items that are required for a deployment to polygon PoS but which cannot be added to the main branch (for example large staking requirements). Its main purpose is to run automated code tests.  It will not be merged because all the changes that it is sensible to incorporate into the main branch are already made.


Westlad:
Closed as this was pushed only to run automated tests.

PR #1425, opened 2023-04-26
check balance after transfer rollback

druiz0992:
Expand adversary test coverage to check balances after transfers are rolledback. Expected outcome is that the commitment used for the incorrect transaction is fully available to the user. That is, that the balance before the incorrect transfer is proposed and after the incorrect transfer is rolledback is the same.


Westlad:
Closing because we believe this problem only arises for a user who is (deliberately) generating bad transactions. Therefore, it's not something we will fix because we wish to discourage bad behaviour.

PR #1428, opened 2023-05-17
Liju.jose/refactor cli folder

LijuJoseJJ:
This PR deletes `cli` folder and place its common used files in `common-files` folders and refactor codes accordingly


common-files/constants/sdk.mjs
```diff
@@ -4,13 +4,6 @@ export const TOKEN_TYPE = {
   ERC1155: 'ERC1155',
 };
 
-export const TX_TYPES = {
```
Westlad, line 7: Can you rename this file `nf3.mjs`? Use of `sdk.mjs` is confusing as there is no sdk in this repo.

LijuJoseJJ: renamed

